PR #146, opened 2013-11-26
Add somewhat more robust handling of input email address (not just bare)

w-barath:
Merged shared code address sanitizing code from setFrom() and addAnAddress()


Synchro:
There is now a generic `parseAddresses` function to parse RFC822 addresses which you can then pass into `addRecipient` etc, which does what you're asking in a slightly less direct way.

PR #155, opened 2013-12-20
Add Extra certificates for S/MIME so to include root CA, etc

LeeNX:
Adding extra certificates so that S/MIME items can verify in Mail Clients without having to first add root CA, makes signed mail work better.


Synchro:
This was fixed in #376

PR #197, opened 2014-03-19
SECURITY SSL: add option to allow insecure connection.

tosiara:
By default we should check CN and require a valid server certificate
The parameter "SSLallowInsecure" will allow user to allow insecure connection. By default it should not allow
Second parameter "SSLCaPath" should point to hashed dir that contains CA certs


Synchro:
Closing this in favour of #339. The specific name of this PR is now handled by the provision of the `SMTPOptions` property.

PR #210, opened 2014-04-02
RFC 5321 - Flag to disable Return-Path header

rocketapi:
Per RFC 5321 the return path header should only be included in the message when sent by final delivery client/server, which is very unlikely to be PHPMailer.  Thus a flag was added that will disable the Return-Path header, while leaving the default to send it as always, thus maintaining backwards compatibility.

Some modern SMTP servers may reject messages containing the Return-Path header, such as Haraka.

Resources:
https://tools.ietf.org/html/rfc5321
http://postmastery.net/2013/05/13/about-the-return-path-header/
http://haraka.github.io/

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -101,12 +101,6 @@ class PHPMailer
     public $ReturnPath = '';
     
     /**
-     * If Return-Path header should be disabled.
-     * If strict compliance to RFC 5321 should be met.
-     * @type boolean
-     */
-    public $DisableReturnPath = false;
-    /**
      * The Subject of the message.
      * @type string
      */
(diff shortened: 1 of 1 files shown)


stefanopallicca:
Hi,
some providers (including some I am using for some domains) require to set the Return-path explicitly, or their antispam system for outgoing e-mails reject the message (even this behaviour is out of RFC).
They started to apply this enforcement in the last days, and will eventually propagate this filter all over their infrastructure in the next months.

Therefore, I would be happy to see this function restored, eventually disabled by default, but that may be explicitly enabled. In the documentation it could be marked as out of standard, as a sort of disclaimer. In the meantime, I stick with an old version of the library.

Synchro:
Since no RFC-compliant senders should be setting this, they are blocking a large proportion of clients (e.g. Apple Mail, Outlook) by doing that. The rest of the world is generally moving in a direction of better RFC compliance and less tolerance of contraventions, not trying to make things worse. I'd suggest you open a support ticket with them to that effect. It's not just that it's out of RFC, it's that it doesn't actually make much sense. If you want it back in, you can always subclass and do it yourself without having to drop back to an old version.

stefanopallicca:
You are right, I don't know why they came out with this solution, probably requiring smtp authentication on php scripts, or other solutions would have sorted the same effect while remaining RFC compliant.
I think I'll subclass as you suggested for the moment, it seems the best solution. Thank you for your support.

Synchro:
For the record, I just happened to see this on the postfix mailing list, by the author of postfix:

> By default, Postfix REMOVES Return-Path headers from email messages.
>The default setting is:
>
>   message_drop_headers = bcc, content-length, resent-bcc, return-path

So even if PHPMailer did set it, postfix will strip it anyway, and since postfix is responsible for a pretty big chunk of all internet mail, trying to push it isn't going to work.

PR #274, opened 2014-09-11
Remove "hacked" from "hacked by Lance Rushing"

Zodiac1978:
Remove "hacked" from "hacked by Lance Rushing", because it is a false positive for some malware scanner who check for the string "hacked by".
See also: https://core.trac.wordpress.org/ticket/27946


ahmadawais:
I was just about to do that.

ooxi:
@ahmadawais well, you are just about 2^6 weeks late ;)

ahmadawais:
:palm_tree: :)

PR #312, opened 2014-11-10
big Chinese tex for content will be garbled

mymicroblog:
bug fix for Chinese character when send mail with smtp which will split Chinese and it will make the contents garbled .


mymicroblog:
OK

PR #332, opened 2014-12-19
Initiate a encrypted session before authentication

offboard:
fix error "530 5.7.0 Must issue a STARTTLS command first".


Synchro:
I think [the autotls branch](https://github.com/PHPMailer/PHPMailer/tree/autotls) is a better solution - it enables opportunistic TLS if the server offers it, even if `SMTPSecure` is not set, so you don't need to do this any more.

PR #338, opened 2014-12-26
Update EasyPeasyICS.php

alfupe:
Updated method addEvent() to accept the UID from the parameter.
Updated method render to use $event["uid"] instead of md5(uniqid(mt_rand(), true)) as the UID value.
All these changes are useful if what you want is to control the create or update an existing event from the application logic and not always forcing the creation of a new event.

### DIFF
--- a/extras/EasyPeasyICS.php
+++ b/extras/EasyPeasyICS.php
@@ -33,13 +33,14 @@ class EasyPeasyICS {
 	 * Add event to calendar
 	 * @param string $calendarName
 	 */	
-	public function addEvent($start, $end, $summary="", $description="", $url=""){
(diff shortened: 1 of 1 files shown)


ooxi:
Please refer to the PHP documentation regarding default argument values

> [Note that when using default arguments, any defaults should be on the right side of any non-default arguments; otherwise, things will not work as expected.](http://php.net/manual/en/functions.arguments.php#functions.arguments.default)

Synchro:
See the recent push that fixes this issue - and thanks for giving me an incentive to clean up EasyPeasyICS!

alfupe:
Ok, thanks!

Many thanks ooxi for your comments also, I will take it into account in the future.

PR #342, opened 2015-01-02
Fixed callable call

SDKiller:
Otherwise in case of psr4-autoloading classes with `composer` -  `Call to undefined function` error 
(thow `var_dump(is_callable($advanced))` returns `true` !!!)


Synchro:
Thanks

PR #351, opened 2015-01-31
Expose PHPMailer->setListener(callback) to allow tracking send() progress

ochameau:
This patch introduce a public API on PHPMailer that allows to track progress of PHPMailer.send().
This allows to send feedback to the client about the progress of sending mail to the SMTP server.
In my usecase, I'm sending big files and PHPMailer->send() takes quite some time to execute,
without being able to report the progress of this step...

Here is an example of code, using this new feature:

```
$listener = function ($progress) {
  echo "progress: ".($progress*100)."%\n";
  flush();
};
$mail->send($listener);
```

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -424,6 +424,12 @@ class PHPMailer
     public $action_function = '';
 
     /**
+     * Dictionary of callables. Used by listen() and emit()
+     * in order to provide a basic event mechanism.
+     */
+    protected $listeners = array();
+
+    /**
      * What to put in the X-Mailer header.
      * Options: An empty string for PHPMailer default, whitespace for none, or a string to use
      * @type string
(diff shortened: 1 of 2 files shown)


Synchro:
This is a nice idea, thanks. I'm not convinced about setting it as a param for send() though. There is already a callback mechanism in PHPMailer (see `doCallback()` and `actionFunction`) that could probably be adapted to provide this service (and they may need some attention anyway).

In its current form, you need to make the listener param for `send()` optional (`$listener = null`) otherwise you break everything - see [the travis errors](https://travis-ci.org/PHPMailer/PHPMailer/builds/49040634).

Also it should be labelled as a `callable` in the docblock, and called using `call_user_func`, which is more reliable than using a variable function - see the html2text callback as an example.

ochameau:
Reusing action callback sounds great, but I do not see exactly how to adapt it.
My main concern is that users of action_function are expected to be called only once per mail,
whereas, here, we would call it up to 100 times per mail. I imagine it can break many existing programs.

Would you happen to be at FOSDEM by any chance?

w3d:
> ...and called using `call_user_func`, which is more reliable than using a variable function

What do you mean by "more reliable"? ... the ability to call class methods in earlier versions of PHP? The ability to trap any errors? Although `call_user_func` doesn't support pass-by-reference. (?)

Synchro:
No, I'm not at FOSDEM.

For the callbacks it might be interesting to build a more flexible mechanism around the existing one. The current one is very simple and doesn't do much (I've not heard of anyone using it). Perhaps allow registration for various different events (defaulting to the current behaviour), but add options for additional callbacks. These days I'm wary of adding too many things to PHPMailer since it's already on the heavy side.

Actually the reliabilty thing was not in the `html2text` method, it is in `edebug` - you can see there it has a check for existing output patterns - this is because a string can be considered callable if it contains the name of a built-in PHP function, such as 'echo', and in that case it caused unexpected behaviour because of the name clash. I don't think that applies here.

ochameau:
Sorry for the delay, I just rebased the patch and tried to do something like what you described.
But I'm not sure the integration with doCallback and action_listener is perfect?
I imagine a better integration would be to just get rid of them and replace them completely with emit/listen...

Once we agree on the API, I will start looking at tests!

PR #365, opened 2015-02-16
Change name of setWordWrap() to applyWordwrap()

eexit:
The method `setWordWrap()` name makes no sense since it's not a setter but it actually applies the word wrapping.
This change introduces a BC break so merge for next version might be wise.

Thank you!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
 docs/phpdoc/
 test/message.txt
 test/testbootstrap.php
+test/*key*
+test/cert*
 .idea
 build/
+vendor
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -1614,11 +1614,11 @@ class PHPMailer
     }
 
     /**
-     * Set the body wrapping.
+     * Applies the body wrapping.
      * @access public
      * @return void
      */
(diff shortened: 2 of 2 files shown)


eexit:
Can you guys fix your CI, thanks!

ivantcholakov:
The name is not a big problem for me. But if you insist on this name change, add the wrapping method setWordWrap() for backward compatibility and declare it as deprecated.

Synchro:
I've asked scrutinizer to sort out that inspection.
I was tempted to close this as it's not worth the BC break for what amounts to a cosmetic change, but if you want to do what Ivan suggested, I'll accept that.

Synchro:
I've changed my mind on this - I don't think it's worth doing. Because `setWordWrap()` is called by `createBody()`, there should never be a need to call it yourself (as it just means it would be called twice), so it doesn't matter what it's called. It should really have been a protected method as it's only really relevant for internal use, but it's too late for that.

eexit:
> It should really have been a protected method

Yes, that'd be even better. Issue when using wrong names is that when building interfaces on the top of this will affect them. This method should be a setter, not a command. 

Either renaming it or changing its visibility is a BC.

PR #381, opened 2015-03-19
Add Required File

zulhilmixrahman:
Add "require_once $this->PluginDir . 'class.smtp.php';" in function smtpSend because of warning class SMTP not found


Synchro:
No. Please read the docs. If you use the autoloader as recommended, this does not happen. Applying this fix breaks the abilty to use custom SMTP classes.

PR #382, opened 2015-03-19
`smtp_msg` was changed to `detail` in 5.2.7

glevine:
When reading through the code, I just happened to notice that the contents of `$lasterror` were never going to end up in the message of the error being set.


Synchro:
You're right, and thanks for spotting it, but on wider inspection, the use of the error array is quite inconsistent anyway. Rather than accepting your PR, I'll make it more consistent everywhere with an internal method for setting error values.

glevine:
I can't argue with that.

Issue #119, opened 2013-10-10
Add automatic host finding for direct SMTP sending


Synchro:
If you are using SMTP sending and do not have a local mail server or nearby smarthost, you need to do DNS lookups to get an MX record and use that to send directly to recipient exchangers. This is a common enough scenario that PHPMailer should have built-in support for it, perhaps by setting 'Host' to 'auto', which would cause it to set the host using `getmxrr()` and friends.

This will obviously run into issues with sending from dynamic addresses (which will probably be blacklisted or blocked outbound by ISPs), but that's a separate problem that PHPMailer doesn't need to be concerned about.

andris9:
If this feature ever gets implemented then here's some things to look for when sending direct mail:

  * The hostname used for EHLO/HELO must match the reverse resolved hostname for the server IP address. Otherwise you get spam points.
  * There are really weird/old ciphers out there, also a lot of self signed certificates, so expect STARTTLS to fail if you try to use it – in this case you should try again, this time in plaintext and ignoring STARTTLS.
  * Greylisting is widely used, so expect to get temporary errors and then try again after a while (I guess this would assume you send using a cronjob, otherwise you'd be blocking the PHP script for 10 minutes to wait for the next try)
  * I don't know how widely fake MX records are used but I guess that not too much because of the high risk of loosing legit messages, so I wouldn't care too much about it
  * I wouldn't care too much about dynamic addresses either since this is what haunts sendmail as well. Actually in case of using sendmail the situation is even worse – sendmail queues your messages and does not inform you that recipients server rejects the connection or that your outgoing port 25 is blocked,  so it looks like the message went through even though it didn't

Synchro:
PHPMailer should never attempt to be an MTA, it's far too complex. Sendmail's approach (like all MTAs) is exactly correct - it will send bounces on delivery failure - but this is clearly not something you care about if you're not running a proper MTA in the first place. We can already handle fake MX records because there is support for multiple hosts in the `Host` property. The whole idea of sending a message via SMTP during a web page submission is horribly flawed, but that doesn't stop many using it, and acting all surprised when they run into things like greetdelay breaking their web apps...

andris9:
[...]

PR #386, opened 2015-03-25
my first try on implementing direct send

cbergmann:
As you pointed out in [Your issue 119](https://github.com/PHPMailer/PHPMailer/issues/119) it would be usefull to be able to send mails directly to the recipients mail Server. This allows more direct responses about wrong mail Adresses etc.

This is my first try on this topic.

What Do you think about this?

best regards
Clemens

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -156,7 +156,7 @@ class PHPMailer
 
     /**
      * Which method to use to send mail.
-     * Options: "mail", "sendmail", or "smtp".
+     * Options: "mail", "sendmail", "direct" or "smtp".
      * @type string
      */
     public $Mailer = 'mail';
@@ -685,6 +685,15 @@ class PHPMailer
     }
 
     /**
+     * Send messages using SMTP directly to recipient host.
+     * @return void
+     */
+    public function isDirect()
+    {
+        $this->Mailer = 'direct';
+    }
+
+    /**
      * Send messages using SMTP.
      * @return void
      */
@@ -1066,6 +1075,8 @@ class PHPMailer
                     return $this->sendmailSend($this->MIMEHeader, $this->MIMEBody);
                 case 'smtp':
                     return $this->smtpSend($this->MIMEHeader, $this->MIMEBody);
+                case 'direct':
+                    return $this->directSend($this->MIMEHeader, $this->MIMEBody);
                 case 'mail':
                     return $this->mailSend($this->MIMEHeader, $this->MIMEBody);
                 default:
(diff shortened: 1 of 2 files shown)


Synchro:
Thanks for having a go! I don't think you should implement this as another sending mechanism, because it's just going to be using SMTP anyway. All it needs is auto-setting of the SMTP host from the recipient address. I think your getMXforAddress function should limit itself to DNS lookups - making SMTP connections in here is a bit pointless because it will be doing that when it tries to send anyway, and will make it much slower and complicate error reporting.

Rather than the separate send mechansim I think it would be better to set some kind of 'autohost' flag so that the SMTP send does a lookup for the Host value, ignoring the manually provided one. That should make things quite a bit simpler. As far as TLS goes, nearly all servers will accept incoming mail without encryption, but it would be useful to have a kind of 'autotls', which could be done by making use of the server capabilities array that was added recently.

Alternatively, it could be simplified quite a bit by only providing the getmxforaddress function - that would only work for sending to one domain at a time, but that's all that usually done.

cbergmann:
Hi,

There are three parts in your answer:

First about splitting the smtp connections:
My first implementation was by modifying the smtpSend function. The problem is that the recipients are likely from different domains and therefore we need different SMTP connections for each recipient (or at least for each domain). We could group the recipients by domain to reduce the connections.

Second about how to find the MX Server for each domain:
An interesting part about how to prevent Spam with specially crafted MX entrys can be found in the [SpamAssassin Wiki](https://wiki.apache.org/spamassassin/OtherTricks). This (and the TLS autodetection) is the reason why I did the connections in descending preference order and returned the first matched smtp server. On second thought it would be possible to merge these two steps. 
We could just get MX "candidates" in the getMXforAddress  function and then loop over them trying to connect. When we successfully connect to one we could reuse this connection to send the mail. (possibly including TLS autodetection). This (in combination with the recipient-grouping from above) could reduce the SMTP connections to the minimum of one connection per domain.

Last about TLS autodetection:
I agree that it is a better way to do it like you did. Especially because that would also effect "smtpSend" connections and would therefore increase the overall security.

Synchro:
Grouping multiple addresses in the same domain would be a sensible idea, and caching the successful host value to use for all of them would be good too.

Yes, I think that getmxforaddress should just return a list of candidates - the Host property already has support for trying to deliver to a list of hosts, so you would just stick the results into that for each message.

One limitation is that while it can use keepalive and send to multiple recipients, it's only expecting to use one Host value and one connection for each message sent (i.e. for all recipients), so you could not use it to send this way for multiple domains - you would have to break your to/cc/bcc lists up by domain ahead of time, and keepalive seases to be of any value as different domains will have different Host values, and thus can't use the same connection.

The problem I see with the direction this is all going is that we're starting to turn into an MTA - all this handling of multiple MXs is ok, but we need to remember that probably the vast majority of the millions of PHPMailer users are sending to one recipient during a page load. Thus it's not tolerable to have long delays that the fake MXs would no doubt incur. Most users are using a local mail server or nearby smarthost (like gmail), and anyone sending seriously will be using a proper MTA anyway, at which point we have no need for direct sending anyway. I think this is why there has not been much demand for this feature to date.

PR #393, opened 2015-04-04
Fix concatenation of headers for Dkim

sveneld:
Fix concatenation of headers for Dkim

If you have long headers like this
=?utf-8?B?QW5pbWUgTGluZSBHcm91cCAtINCh0YPQv9C10YDQvNCw0YDQutC10YIg0JA=?=
 =?utf-8?B?0L3QuNC80LU=?= <xxx@xxx.com>

Original code does not correct concatenate header

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -3411,8 +3411,8 @@ class PHPMailer
                 $to_header = $header;
                 $current = 'to_header';
             } else {
-                if ($current && strpos($header, ' =?') === 0) {
-                    $current .= $header;
+                if ($$current && strpos($header, ' =?') === 0) {
+                    $$current .= $header;
                 } else {
                     $current = '';
                 }
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -3411,7 +3411,7 @@ class PHPMailer
                 $to_header = $header;
                 $current = 'to_header';
             } else {
-                if ($$current && strpos($header, ' =?') === 0) {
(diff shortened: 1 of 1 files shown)


Synchro:
I'm not entirely clear what this does - could you post example of what it does before and after your changes please?

sveneld:
if parameter FromName is use UTF-8 symbols and has a long length
for example *"ім'я відправника українською мовою"*
Phpmailer will convert it to such header (2 lines)
```php
string(78) "From: =?utf-8?B?0ZbQvCfRjyDQstGW0LTQv9GA0LDQstC90LjQutCwINGD0LrRgNCw0ZfQvdGB?="
string(58) " =?utf-8?B?0YzQutC+0Y4g0LzQvtCy0L7Rjg==?= "
```
For DKIM sign this to line should be combined in one. 

result before my fix:
```php
string(78) "From: =?utf-8?B?0ZbQvCfRjyDQstGW0LTQv9GA0LDQstC90LjQutCwINGD0LrRgNCw0ZfQvdGB?="
```

result after my fix:
```php
string(136) "From: =?utf-8?B?0ZbQvCfRjyDQstGW0LTQv9GA0LDQstC90LjQutCwINGD0LrRgNCw0ZfQvdGB?= =?utf-8?B?0YzQutC+0Y4g0LzQvtCy0L7Rjg==?= "
```

Synchro:
OK, that's clear, thanks.

PR #398, opened 2015-04-13
Documentation update class.pop3.php

Phhere:
Since it is just a wrapper for authorise it should accept the same parameter types


Synchro:
Thanks

PR #399, opened 2015-04-17
Fix $all_recipients comment

pstast:
$all_recipients does not contain $replyto addresses, see addAnAddress() behavior


Synchro:
Thanks

PR #401, opened 2015-04-20
Saving some memory for echo.

dseguy:


ooxi:
:+1:

Synchro:
This kind of change is really not worth doing. Notice that not a single one is on any kind of critical path, so the time and memory improvements will be immeasurably small and make no quantifiable difference to PHPMailer. The difference between `.` and `,` has dropped dramatically in recent versions of PHP, in some cases down to less than 2%, which is well into statistical noise on something that is only run once and only takes 0.000001 sec anyway.

I think I'd go as far as to say that it's actually doing more harm than good - given that this syntax is used in only a single context in the entire PHP language, it's more likely to confuse and become a source of other problems, certainly more likely than it is to provide any meaningful performance gain. There might be other situations where very large numbers of things are echoed (perhaps in a tempting system), but that does not happen in PHPMailer.

PR #405, opened 2015-04-23
Add custom header getter

eexit:
Hello,

I couldn't believe PHPMailer is able to add and clear header but not able to get them. Let's say I added a wrong header, I'm not sure I want to restart all from scratch.

It would be nice to have this merged into next minor version.

Thank you :)


Synchro:
Merci!

PR #411, opened 2015-05-04
Improved Traditional Chinese PHPMailer language

Cojad:
1. Completed new language key $PHPMAILER_LANG['extension_missing']. 
2. fix translation error of $PHPMAILER_LANG['signing'] with correct meaning in Traditional Chinese 
3. Re-order the translation keys to make it the same with default english locale.


Synchro:
Thanks

PR #415, opened 2015-05-05
Fix typos

maqnouch:


Synchro:
Yep, I'm a Brit, so my style wins :smirk:. 'Licenced' is wrong in either idiom though, so I've changed that. Strictly speaking, 'license' as a noun is American English, but I can live with that.

maqnouch:
:dancer:

Issue #350, opened 2015-01-28
Support google XOAUTH2 authentication


Synchro:
Google have recently started imposing a different authentication mechanism for access to their apps, including gmail. This substitutes normal standards-based SMTP authentication for authorisation via OAuth2, and does not really improve security much at all - it simply means you have to enter your ID and password somewhere else (over the same protocols that they classify as insecure), and add complexity to your project to implement OAuth. This change breaks SMTP and IMAP access to gmail, and you may receive authentication failures (usually "5.7.1 Username and Password not accepted") from many email clients, including PHPMailer, Apple Mail, Outlook, Thunderbird and others.

To do this in PHP you'll need an OAuth2 client class - like [this one](http://www.phpclasses.org/package/7700-PHP-Authorize-and-access-APIs-using-OAuth.html) - to retrieve an authorisation token before making an SMTP connecton as normal. Google's docs on the subject are [here](https://developers.google.com/gmail/xoauth2_protocol).

PHPMailer does not currently support this access mechanism, so whether we like it or not, we should implement it simply because gmail is so popular!

As far as an implementation in PHPMailer goes, it should involve adding the `XOAUTH2` authentication type and provide a mechanism to provide an OAuth token. Creating and obtaining the token should be left to external scripts/libraries such as the one linked to above.

michield:
That would be awesome.

Synchro:
Implemented in the 5.2.11 release.

PR #421, opened 2015-05-17
Support google XOAUTH2 authentication Issues #350 - Added Code

sherryl4george:
Support For Google XOAUTH2 Added - OAUTH2 External Library requires >=PHP5.4 & Guzzle
To begin, obtain OAuth 2.0 client credentials from the Google Developers Console. Then your client application requests an access token from the Google Authorization Server, extracts a token from the response, and sends the token to the Google API that you want to access.
More Info : https://developers.google.com/identity/protocols/OAuth2

The Implementation is as follows: 1. Get a refresh token for the first time when OAUTH 2.0 credentials are requested from Google 2. The refresh token is saved and this refresh token is used to get Access Tokens, which inturn is used for authentication to the GMAIL API.
Why Save Refresh Token when you can get the Access Token itself?
According to Google OAUTH2.0 standards, inorder to obtain Access Token, a End-User cosent is needed. In this case the End-User will be given a Authorization Screen, which is not what we want. Morever we dont want to use the End-User Account info. But we would be using the Developers or Website Owners Email Account to send mails. Using the above method, access token to the End-User Account will be obtained, which is not the case that we want.

As a work around to this a Refresh Token for the Website Owner Account is obtained before hand by setting necessary credentials in Google Developers Console and running the get_oauth_token.php and this Refresh Token is permanently used for getting temporary access tokens.

Please note Refresh Token is send only the first time you request for Access Token and not thereafter.

 >Added a third party library for OAUTH2 access tokens (majority of files)
(https://github.com/thephpleague/oauth2-client)
 >Added a file to easily obtain Refresh Token (get_oauth_token.php). If no refresh token is obtained when running this file, please Revoke Access to API using link: https://accounts.google.com/b/0/IssuedAuthSubTokens.
 >Added example for sending mail using XOAUTH2
 >Documetation is not provided as of now. Will add it ASAP
 >Supports only Google for time being.


Synchro:
This is great, thanks for getting into it, it's been wanted for a while. There are some issues...

* We should not bundle the OAUTH library - it's way bigger than the whole of PHPMailer, and will present version control issues in future. Ideally it would be loaded via composer.
* The OAuth-specific properties have rather generic names even though they are specific to OAuth - better to call them something like `$oauthUserEmail` etc to make it clearer. Or can this be broken out into a separate oauth instance?
* Explicitly loading the class from extras isn't pretty and precludes other implementations. It would be better to inject the OAuth instance - look at how SMTP loading or HTML filtering options work.
* The calls to the oauth class will break in PHP < 5.3 because it's using namespaces; we currently support back to 5.0, and there are many installations that use things that old. This could be solved by making OAuth support require 5.4 (since it's a new feature), but not break in versions older than that. Perhaps implement it as a PHPMailer subclass?
* The SMTP::authenticate method's param list is getting very long. Could we pass in an oauth instance instead? Or subclass SMTP as well?
* The travis build is breaking - might just be a travis glitch by the looks of it.

sherryl4george:
Thanks for the feedback Synchro. I will try making changes according to your feedback and get back to you ASAP.

sherryl4george:
* We should not bundle the OAUTH library - it's way bigger than the whole of PHPMailer, and will present version control issues in future. Ideally it would be loaded via composer. - The OAUTH library has been moved out and entry has been made to composer.json to load OAUTH and Guzzle. But I doubt whether it is commited to Git. Plz comeback if not

* The OAuth-specific properties have rather generic names even though they are specific to OAuth - better to call them something like $oauthUserEmail etc to make it clearer. Or can this be broken out into a separate oauth instance?- names has been changed

* Explicitly loading the class from extras isn't pretty and precludes other implementations. It would be better to inject the OAuth instance - look at how SMTP loading or HTML filtering options work. - A OAuth instance is being created in the file class.oauth.php in the same way as SMTP. Now this is passed to SMTP::authenticate as well

*  The calls to the oauth class will break in PHP < 5.3 because it's using namespaces; we currently support back to 5.0, and there are many installations that use things that old. This could be solved by making OAuth support require 5.4 (since it's a new feature), but not break in versions older than that. Perhaps implement it as a PHPMailer subclass? Subclassed to PHPMailer54 implementing logic to get OAuth instance and continue with the old SMTP class.

*  The SMTP::authenticate method's param list is getting very long. Could we pass in an oauth instance instead? Or subclass SMTP as well? OAuth instance is passed. Thought not to subclass just to add files. Just passed a instance and instance is checked for null before authentication is done

*  The travis build is breaking - might just be a travis glitch by the looks of it. - Please check now.

Synchro:
Thanks for doing that. I've rebased your fork to skip all those external files, cleaned up the code a bit, added some minor documentation mentions and pushed it into the [xoauth branch](https://github.com/PHPMailer/PHPMailer/tree/xoauth). I renamed the oauth subclass to `PHPMailerOAuth` and renamed the file `class.phpmaileroauth.php` to make it clearer what the difference is. Could you please take a look at it and let me know your thoughts.

sherryl4george:
Everything seems to be absolutely fine. Please note the below
* I think there are two files `gmail_xoauth.phps` and `gmail_xoauth.php` in the examples folder. Both are same retain only one
* Add `XOAUTH` as new authtype in `class.smtp.php` Line 356

Synchro:
I've done a load more cleanup, actually got it working, and have written a guide to [how to set up an OAuth client ID](https://github.com/PHPMailer/PHPMailer/wiki/Using-Gmail-with-XOAUTH2) in the wiki. I renamed the `AuthType` value to `XOAUTH2` so that it removes any danger of clashing with older `XOAUTH` implementations. What I could do with now is to have some others give this a test!

fstfrognation:
Forgive me for jumping in here, but I'm willing to test.  I'm probably not the most adept, but I've already run across one issue - I can't seem to find the "vendor" directory in the xaouth branch.  So get_oauth_token.php fails.

Synchro:
Great. @fstfrognation, meet [composer](https://getcomposer.org). You need to run `composer install` and it will create the vendor directory and install the dependencies.

fstfrognation:
Thanks - told you I probably wasn't very adept.  I'll play with this as
soon as I get out from behind my work's firewall.

- Rob

fstfrognation:
Hi,
Quick update (not sure you wanted them or not... not sure what the best way to provide an update)...

All seems to be working very well, I have been able to use it after some initial headaches with composer.  The only issue I had was that the file examples/gmail_oauth.phps probably needs to be updated... need to prepend "oauth" before the OAuth specific variables.

$mail->UserEmail  ... should be $mail->oauthUserEmail
$mail->ClientId  ... should be $mail->oauthClientId
$mail->ClientSecret  ... should be $mail->oauthClientSecret 
$mail->RefreshToken  ... should be $mail->oauthRefreshToken

Its correctly documented in your great set of directions on the wiki, just the examples directory wasn't updated.

- Rob

Synchro:
That's just the kind of thing I was looking for! All those issues are fixed in the xoauth branch now.

sherryl4george:
Hi Synchro,
How can we go about pulling more people to test the OAUTH 2 feature for GMail

Synchro:
Not sure how we can get more help - I've added the 'help needed' label. Could try tweeting about it.

@fstfrognation is it working ok for you now?

fstfrognation:
Sorry - I thought I had commented earlier - but obviously I didn't - sorry about that.  Everything is working perfectly now.  I've been using it for the past 5-6 weeks without any problems.  I haven't found anything else new to report back.

Usage, error messages, and performance all seem great to me.

- Rob

alexplusde:
Just for the record: I tried to follow 
https://github.com/PHPMailer/PHPMailer/wiki/Using-Gmail-with-XOAUTH2
and failed to generate a RefreshToken.

But i followed the instructions given in: 
http://masashi-k.blogspot.de/2013/06/sending-mail-with-gmail-using-xoauth2.html
which seem much more easier as you don't have to mess with an external library.

Synchro:
@alexplusde That makes no sense - that article is doing exactly the same thing using nodemailer in node.js, and so is not usable in PHP. `nodemailer-smtp-transport` has a dependency on the `xoauth2` module from npm, i.e. it's doing exactly the same thing as PHPMailer, and *is* using an external library. If you install PHPMailer with composer, it works exactly the same way.

alexplusde:
I just talked about getting the refresh token part. 
Well, I don't use composer, that's why it was not as easy for me and the instructions for getting the refresh token via https://developers.google.com/oauthplayground may be an alternative for people with less server-side experience (like me).

Anyway, thanks for the effort, the wiki pages and examples really helped a lot.

Synchro:
Ah, ok I see what you mean, point taken. Given that the process for obtaining the refresh token depends on the same libraries as using the token, it's unlikely that auth will work if obtaining the token doesn't. I would expect that obtaining the token via the playground is less secure than doing it yourself as it means the signing process is not happening under your control or on your server.

Synchro:
This has been merged in manually and released in 5.2.11 - thanks for all your work.

ShubhamDalal:
> I've done a load more cleanup, actually got it working, and have written a guide to [how to set up an OAuth client ID](https://github.com/PHPMailer/PHPMailer/wiki/Using-Gmail-with-XOAUTH2) in the wiki. I renamed the `AuthType` value to `XOAUTH2` so that it removes any danger of clashing with older `XOAUTH` implementations. What I could do with now is to have some others give this a test!

Hii @Synchro, 
This wiki guide need some updates.
google developer console has changed a lot.

Synchro:
Ok, so please go ahead and update it; The wiki is open to edits.

ShubhamDalal:
> Ok, so please go ahead and update it; The wiki is open to edits.

Thank you.
before updating it thought of giving it a try and i'm getting this error.
 
```
PHP Fatal error:  Uncaught Error: Class 'League\OAuth2\Client\Provider\Google' not found in /home/user/public_html/api/vendor/phpmailer/phpmailer/get_oauth_token.php:96
Stack trace:
#0 {main}
  thrown in /home/user/public_html/api/vendor/phpmailer/phpmailer/get_oauth_token.php on line 96
```

Synchro:
Well, did you add that dependency as the composer config suggests and the docs say?

ShubhamDalal:
Yes, I did.

ShubhamDalal:
everything worked fine after running `composer require league/oauth2-google`

Synchro:
Right, that's why it's in [the composer suggest list](https://github.com/PHPMailer/PHPMailer/blob/master/composer.json#L41).

Issue #175, opened 2014-01-29
iCalendar invite is not showing RSVP buttons in email clients


ghost:
I'm using PHPMailer to send email. My ics content is:

```
BEGIN:VCALENDAR
METHOD:REQUEST
VERSION:2.0
PRODID:-//FooBar//FooBar Calendar//EN
BEGIN:VEVENT
UID:8f5a0777-bf6d-17d2-f14a-52e7feedf810
ORGANIZER;CN=Foo:foo@bar.com
DTSTART:20140130T191500Z
DTEND:20140130T194500Z
DTSTAMP:20140129T144300Z
SUMMARY:Foo Bar 11
DESCRIPTION:
END:VEVENT
END:VCALENDAR
```

For some reason I don't see the expected RSVP buttons in any email clients. Both Gmail and Mac Mail simply include the ics as a file attachment, I assume as a fallback. So there must be something wrong. I have not tried Outlook, but I don't see any reason why it would be different in a third client. The contents of an email being sent are below. Can you see anything that looks incorrect? Thank you.

```
Return-Path: REDACTED
Received: REDACTED
Date: Wed, 29 Jan 2014 09:43:18 -0500
Return-Path: REDACTED
To: REDACTED
From: REDACTED
Reply-To: REDACTED
Subject: Meeting - Foo Bar 11
Message-ID: <7004961cd9df04181047296f98260834@localhost>
X-Priority: 3
X-Mailer: PHPMailer 5.2.7 (https://github.com/PHPMailer/PHPMailer/)
MIME-Version: 1.0
Content-Type: multipart/alternative;
boundary="b1_7004961cd9df04181047296f98260834"
Content-Transfer-Encoding: base64

--b1_7004961cd9df04181047296f98260834
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: base64

[...]

PR #432, opened 2015-06-08
Added support for sending iCal messages in the "alt_attach" mail type

Lambik:
Copied from the "alt" mail type when creating the body, so now you can send an iCal attachment inline together with regular attachments

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,7 @@
 
 * Don't switch to quoted-printable for long lines if already using base64
 * Fixed Travis-CI config when run on PHP 7
+* Add address parser for RFC822-format addresses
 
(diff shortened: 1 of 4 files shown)


Synchro:
While you're looking at this, make sure you're aware of its limitations, as discussed in #175.

Lambik:
I agree that it's not the best solution to add the block to all switch cases. I've tested in Outlook 2010 and gmail that there was no need to add an .ics file with the attachments (but this was not really the issue I'm fixing, I'm just using the existing logic of iCals in another case)

Synchro:
If gmail is handling it correctly now, that's good news.

PR #447, opened 2015-06-29
do not auto-replace inline images in body that are already replaced

mrbellek:
I ran across a problem where images in the body that are already in 'cid:name' format generate warnings because PHPMailer still attempts to auto-convert them to 'cid:name' format (and fails, because the file does not exist). Is this a useful fix to merge?

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -3108,7 +3108,7 @@ class PHPMailer
                             $message
                         );
                     }
-                } elseif (!preg_match('#^[A-z]+://#', $url) && substr($url, 0, 4) !== 'cid') {
+                } elseif (substr($url, 0, 4) !== 'cid:' && !preg_match('#^[A-z]+://#', $url)) {
                     // Do not change urls for absolute images (thanks to corvuscorax)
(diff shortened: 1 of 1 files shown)


Synchro:
`msgHTML` is really intended for HTML source that *does* need conversion - if you're handling `cid`s yourself, you should probably set the body directly. That said, avoiding errors is a good idea anyway, though `preg_*` functions are overkill for such small comparisons - `substr` would be more appropriate.

mrbellek:
Well, then the check can be replaced with `substr($url, 'cid:') === FALSE` or `strpos($url, 'cid:') === FALSE`. :smiley:

Synchro:
I was thinking `substr($url, 0, 4) !== 'cid:'` for a more accurate match.

mrbellek:
OK, updated the PR.

Synchro:
You missed a colon. It would be slightly faster to put the new check first.

mrbellek:
Fixed. Apologies, I don't contribute to repos often. :)

Synchro:
Thanks

PR #460, opened 2015-07-17
Update gmail.phps

freephile:
The internet is littered with people trying to figure out why PHPMailer doesn't work for them with GMail.  ISPs like Digital Ocean do not support SMTP traffic over IPv6, but Google is IPv6 ready so PHPMailer waits a long time trying that before it tries IPv4.  That wait is either a major bottleneck, or a failure point as execution timeouts come into play.
See https://www.digitalocean.com/community/questions/outgoing-connections-on-port-25-587-143-blocked-over-ipv6
and
https://stackoverflow.com/questions/20068484/phpmailer-using-gmail-smtp-slow-when-sending-emails for example

### DIFF
--- a/examples/gmail.phps
+++ b/examples/gmail.phps
@@ -26,6 +26,9 @@ $mail->Debugoutput = 'html';
 
 //Set the hostname of the mail server
 $mail->Host = 'smtp.gmail.com';
+// use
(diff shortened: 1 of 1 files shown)


Synchro:
Good tip, thanks

Issue #476, opened 2015-08-22
Signed email has signature not attached (but within the mail)


sisu4u:
Hi,

I use the following code to sign and send an email:
```
$mail = new PHPMailer(true); 
$mail->CharSet = 'utf-8';
$mail->AddAddress("me@example.com");
$mail->SetFrom("me@example.com");
$mail->Subject = "TestMail";
$mail->Body = "This is a test mail";
$mail->sign('cert.crt', 'cert.key', 'xxx');
$mail->Send();
```
I generated a certificate with StartSSL. Then I exported the certificate from my chrome (pfx).
I used the following commands to generate the crt and key file:
```
openssl pkcs12 -in crt.pfx -nocerts -out cert.key -nodes
openssl pkcs12 -in crt.pfx -nokeys -out cert.crt
```

The email is sent without any error. However, the signature is not attached as a file as supposed.
This is how the email is shown:
```
MIME-Version: 1.0
Content-Type: multipart/signed; protocol="application/x-pkcs7-signature"; micalg="sha-256"; boundary="----F4131BBA0FF79046CC5FD7B59A3AEABF"

This is an S/MIME signed message

------F4131BBA0FF79046CC5FD7B59A3AEABF
Content-Type: text/plain; charset=utf-8
Content-Transfer-Encoding: 8bit

This is a test mail

------F4131BBA0FF79046CC5FD7B59A3AEABF
Content-Type: application/x-pkcs7-signature; name="smime.p7s"
Content-Transfer-Encoding: base64
Content-Disposition: attachment; filename="smime.p7s"

MIIJFwYJKoZIhvcNAQcCoIIJCDCCCQQCAQExDzANBglghkgBZQMEAgEFADALBgkq
hkiG9w0BBwGgggY0MIIGMDCCBRigAwIBAgIDDx04MA0GCSqGSIb3DQEBCwUAMIGM
MQswCQYDVQQGEwJJTDEWMBQGA1UEChMNU3RhcnRDb20gTHRkLjErMCkGA1UECxMi
U2VjdXJlIERpZ2l0YWwgQ2VydGlmaWNhdGUgU2lnbmluZzE4MDYGA1UEAxMvU3Rh
cnRDb20gQ2xhc3MgMSBQcmltYXJ5IEludGVybWVkaWF0ZSBDbGllbnQgQ0EwHhcN
[...]

PR #477, opened 2015-08-22
Added an example file explaining how to send signed emails

sisu4u:
I added the example as requested in https://github.com/PHPMailer/PHPMailer/issues/476#issuecomment-133739901


Synchro:
Thanks

PR #481, opened 2015-09-02
Do no require a dependency that is not required

frederikbosch:
First of all, do not require dependencies that are not required for PHPMailer to work. It works perfectly without. Secondly, the majority of the users will not use Gmail's XOAUTH2 authentication system and do not need these additional dependencies. Thirdly, guzzle and league packages are used a lot. Therefore you make it hard for people that were already using these packages to upgrade to newer versions of PHPMailer. No offence, good work on the feature, but this should be considered.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,8 +27,7 @@
         "phpunit/phpunit": "4.3.*"
     },
     "suggest": {
-        "league/oauth2-client": "0.10.*",
-        "guzzle/guzzle": "~3.7"
+        "league/oauth2-client": "Needed for Gmail's XOAUTH2 authentication system"
     },
     "autoload": {
         "classmap": [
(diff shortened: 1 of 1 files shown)


Synchro:
This came up yesteday. The guzzle dependency may not be needed at all - oauth2-client now has a dependency on guzzlehttp/guzzle 6, so I think it can be removed from here entirely - do you agree?

frederikbosch:
@Synchro In that case, I would suggest league/oauth2 and remove guzzle completely.

Synchro:
Thanks, I'll push this with a new tag.

frederikbosch:
@Synchro Thanks for the fast responses; and for the tag in advance. Much appreciated.

PR #485, opened 2015-09-03
Update class.smtp.php

marciseli:
Office 365 need that when you send mail from a webhoster.

### DIFF
--- a/class.smtp.php
+++ b/class.smtp.php
@@ -246,6 +246,9 @@ class SMTP
      */
     public function connect($host, $port = null, $timeout = 30, $options = array())
     {
+        if (count($options) == 0) {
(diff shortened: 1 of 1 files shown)


Synchro:
Please [read the docs](https://github.com/PHPMailer/PHPMailer/wiki/Troubleshooting) on this. Making everyone else's email insecure isn't a good solution to fixing one host or your own config.

marciseli:
I read the config but no point helped me. The problem was, that i could send mails when I work on xampp(localhost) but when I load the code on hoststar it didn't work. So if you have a better idea pleas let me know.

Synchro:
The troubleshooting guide tells you exactly how to do the same change you did in this pull request, but without altering the code.

PR #497, opened 2015-09-15
Update French translation, minor changes

fbonzon:


Synchro:
Merci!

glensc:
class.phpmailer.php 100644 → 100755

DISLIKE!

Synchro:
Good catch @glensc, fixed in master

glensc:
:)

fbonzon:
I used the "Source code (zip)" link for PHPMailer 5.2.13 on https://github.com/PHPMailer/PHPMailer/releases as source to prepare this pull request, on Mac OS. After repeating it now, I realize that after unzipping all files have the exec bit on :(

The files from the "Source code (tar.gz)" link do preserve correct exec bits though. A good reason to use only tar.gz links from now on.

Synchro:
Good to know, but if you know you're going to do a PR, surely it's easier to clone the repo instead of downloading the tarball?

glensc:
And even more simplier is just to navigate to file in github and press edit button. you can make such trivial changes like translations straight from github web!

fbonzon:
Thanks for your suggestions.

FYI, I can add about unzipping the GitHub .zip files on Mac OS. I did further tests and can say that whether file permissions are preserved or not actually depends on the software used for decompressing.

With Archive Utility (built-in app from Apple) and 2 other popular free apps StuffIt Expander and The Unarchiver, all files become executable. But with the unzip command in Terminal, correct executable bits are preserved. I learned something today.

PR #500, opened 2015-09-16
More checks for attacments

cesare-brn:
All started with this:
https://bugs.php.net/bug.php?id=70467

This caused some email with an attachment taken from an UNC path to be sent with an empty body and with no error from my script. So i've modified PHPmailer to:
- remove is_readable(), replacing it by directly testing the return value from file_get_contents().
- update some italian translation.
- add a check on the return value from createBody(), to avoid sending empy mail.


Synchro:
That body check could be better - you could check if the body is empty at the end of `createBody()` and use the existing `empty_message` error rather than checking it from the outside, plus also check to see if exceptions are enabled before doing so. There's already a similar check for message signing in there. Also look at `AllowEmpty`. From a quick look, the only other ways that `createBody` can fail is if `encodeString` or `encodeFile` fail.

`file_get_contents` is a reasonable workaround for that annoying bug, but I'm not too keen on it being able to use fopen wrappers at that point - other things are dependent on the thing being attached being a real file and not a remote resource, plus all the complications of needing to allow stream context params etc - PHPMailer shouldn't try to be an HTTP client. I recommend using `addStringAttachment` and dealing with the fetching yourself, and that would allow you to work around the bug you've run into as well, so I don't know how necessary this is.

cesare-brn:
There are different things about your comments i haven't understood quite well.

I thought empty_message was for user error, when no body is provided. What i'm trying to catch are errors that make createBody() return empty, even if user provided a body. I think it could be misleading to throw an exception with an empty_message error when this->Body is set. And the same for AllowEmpty: if a body was provided but createBody() returned empty, that should be an error even it AllowEmpty is true.
By the way, in these days i'll modify the code to use empty_message and to comply with the other requests.

What i understand less is your comment about file_get_contents(): that function was already present, i've only added a check on its return code, to completely avoid is_readable(). I understand your addStringAttachment() workaround and i'll use it if you are not going to accept my proposal. But i don't understand what's wrong with checking file_get_contents() return code. Please could you explain one more time?

Synchro:
I think my issue with that error handling approach is that it wasn't that much more specific - you'd added it because of the is_readable problem, but there are multiple ways it could end up empty.

I'm also slightly hesitant to add more translation strings given the number of languages we have!

`encodeFile` and `addAttachment` make some calls that are dependent on the path being a file system path (is_file, basename) - in fact the `is_readable` check throws an exception already, so to fix your bug, using `file_get_contents` as you suggest and throwing the exception then seems reasonable, so we report the source of the problem (being unable to read a file) and use the `file_access` error message rather than the later, less-accurate consequence of having an empty body. Sound better?

`is_readable` is also used in the `setLanguage` method, so it probably needs fixing there too.

cesare-brn:
Ok with the translation strings: removed.

I've addressed some of your issues and tested the code: i think that now more error conditions are catched. But i admit it is effective only if exceptions are active: if not, the code is substantially the same as before (apart is_readable()) and a failure with file_get_contents() can still cause an empty message to be sent. The problem is that the errors caught by inner functions are lost and at the end createBody() returns empty and nothing checks that. I think that we ought check what createBody() returns, something like what i proposed at the beginning.

For the rest from my tests i've seen that other filesystem functions like is_file() or file_exists() doesn't suffer the problem that is_readable() have. And for the is_readable() inside setLanguage() it's not that easy to avoid it, because it's needed to guard an include, which is not a function, so not testable. But i think it's not a problem, because language files are bundled with phpmailer and are not supposed to be on an UNC path.

Synchro:
I've incorporated these changes into PHPMailer 6.0 manually.

PR #502, opened 2015-09-17
In examples, prefer setFrom() over separate From and FromName

fbonzon:


Synchro:
Thanks, that is a better way of doing From, especially for examples.

PR #503, opened 2015-09-19
Fix comments about $Hostname

fbonzon:
According to changelog, setting the Received header has been removed in Version 1.72 in 2004.

The default HELO string is $Hostname iff $Hostname is not empty.


Synchro:
Thanks - there is a lot of very old stuff in here!

PR #506, opened 2015-09-23
XOAuth Provider - Google,Yahoo,Microsoft

hayageek:
I have added XOAuth Provider support for Google(existing),Yahoo,Microsoft
Check the build status here: https://travis-ci.org/hayageek/PHPMailer/builds/81753588

Locally I have checked the test cases.
$ ../vendor/bin/phpunit --configuration ../travis.phpunit.xml.dist phpmailerXOAuthTest.php 
PHPUnit 4.7.7 by Sebastian Bergmann and contributors.
Warning:	The Xdebug extension is not loaded
		No code coverage will be generated.

...

Time: 67 ms, Memory: 5.75Mb

OK (3 tests, 15 assertions)

### DIFF
--- a/class.phpmaileroauth.php
+++ b/class.phpmaileroauth.php
@@ -54,7 +54,16 @@ class PHPMailerOAuth extends PHPMailer
      * @type OAuth
      * @access protected
      */
+    public $oauthProviderName = '';
+
+    /**
+     * An instance of the OAuth class.
+     * @type OAuth
+     * @access protected
+     */
     protected $oauth = null;
+    
+    
 
     /**
      * Get an OAuth instance to use.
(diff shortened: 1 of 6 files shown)


hayageek:
I did not update composer.json. :(

Synchro:
Nice work, thanks. For future expansion, and ease of adding new providers, I think it would be better to create an abstract base class and then implement each provider as a subclass containing the implementation specific parts, rather than rolling them all into one. It would also mean we don't need the additional `oauthProviderName` property. It would be good if we could retain backward compatibility with the existing google implementation.

A smaller thing is that the 'require's for the class are not necessary - add the paths to the composer.json classmap and the autoloader will find them.

Why did you close this PR?

hayageek:
I did not check in the composer.json.  And i am seing some errors in travis job. But I have tested locally working fine. tested all authentication.
Anyways I requested one more. You can take the changes. make customization. 

https://github.com/PHPMailer/PHPMailer/pull/507

Synchro:
You don't need to do that to make changes to a PR - just push more commits and they will be added automatically. It's considered good practice to squash commits (and force-push) to keep history clean.

PR #507, opened 2015-09-23
XOAuth Provider - Google,Yahoo,Microsoft

hayageek:
I have added XOAuth Provider support for Google(existing),Yahoo,Microsoft

### DIFF
--- a/class.phpmaileroauth.php
+++ b/class.phpmaileroauth.php
@@ -50,15 +50,16 @@ class PHPMailerOAuth extends PHPMailer
     public $oauthClientSecret = '';
 
     /**
-     * An instance of the OAuth class.
-     * @type OAuth
+     * An instance of the OAuth Provider class.
+     * @type PHPMailerOAuthProvider
      * @access protected
      */
     protected $oauth = null;
-
+    
+ 
     /**
      * Get an OAuth instance to use.
-     * @return OAuth
+     * @return PHPMailerOAuthProvider
      */
     public function getOAUTHInstance()
     {
--- a/class.phpmaileroauthmicrosoft.php
+++ b/class.phpmaileroauthmicrosoft.php
@@ -18,17 +18,17 @@
  */
 
 /**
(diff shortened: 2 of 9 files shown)


Synchro:
Don't add those dependencies as `requires` - put them in `suggests` as otherwise it will break BC with anyone running PHP < 5.4.

hayageek:
OK. I will make the backward comparability for additional oauthProviderName property, by defaulting it to Google.
Let me fix all the errors and I will provide the clean code.

Synchro:
No, I'm saying that if you put each provider in its own class, you don't need the provider name at all - it's implicit in the class name.

hayageek:
But how the PHPMailerOAuth knows to pick the right provider class name. 
If you don't mind can you explain bit .

hayageek:
got it 
For example: .
1).PHPMailOauthGoogle  extends PHPMailerOAuth 
$mail = new PHPMailOauthGoogle();
2).PHPMailOauthYahoo  extends PHPMailerOAuth 
$mail = new PHPMailOauthYahoo();

is that OK ?

Synchro:
Yes, that way the individual provider classes will be very simple.

hayageek:
Please review the code.

hayageek:
Fixed all the issues.

hayageek:
Please let me know the changes are ok

Synchro:
I've accepted this into the xoauth branch, I need to change some things before merging into master.

hayageek:
Thanks. Slowly I will add some more providers.
Yandex, Mail.ru

Synchro:
Cool. The main change I want to make is that the `phpmaileroauthprovider` base class should not inherit from the `phpmaileroauth` class. There's no reason for the providers to be PHPMailer child classes - PHPMailer needs to *use* them, but not *be* them. It may be that I need to introduce other child classes to make that work, I've not thought through how best to arrange it yet.

PR #516, opened 2015-09-30
Accept email addresses with an UTF-8 domain

michield:
When using a UTF8 address it is rejected with "You must provide at least one recipient email address".

This patch will accept UTF8 addresses. You can try this with eg "phplist@phplíst.com"

It requires [idn_to_ascii](http://php.net/idn_to_ascii) to work which is PHP5.3 and up

It will still require the MTA to handle UTF8 addresses, which Postfix 3.0 does. But at least PHPMailer won't stop it.

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -840,6 +840,14 @@ class PHPMailer
             }
             return false;
         }
+        if (function_exists('idn_to_ascii') && $this->hasMultiBytes($address)) { ##@TODO: Write our own "idn_to_ascii" function for PHP 5.2 and earlier.
+            // can't use $this->parseAddresses, because it calls validateAddress, which will fail here
+            $elements = explode('@',strrev($address),2);
+            if (!empty($elements[0]) && !empty($elements[1])) {
+                $address = strrev($elements[1]).'@'.idn_to_ascii(strrev($elements[0]));
+            }
+            unset($elements);
+        }
         $address = trim($address);
         $name = trim(preg_replace('/[\r\n]+/', '', $name)); //Strip breaks and trim
         if (!$this->validateAddress($address)) {


Synchro:
I can't accept this into master while we are maintaining 5.0 BC. It would be better to use a local function instead. Surely if you are converting IDNs into punycode before sending, then the MTA doesn't need to be involved?

class.phpmailer.php
```diff
@@ -840,6 +840,13 @@ protected function addAnAddress($kind, $address, $name = '')
             }
             return false;
         }
+        if ($this->hasMultiBytes($address) && function_exists('idn_to_ascii')) {
```
fbonzon: I would add a comment here:

TODO: Write our own "idn_to_ascii" function for PHP 5.2 and earlier.

fbonzon:
The code checks if `function_exists('idn_to_ascii')`, how is this incompatible with PHP 5.2 and before? Those versions just won't have this additional feature, but it will not break them.

class.phpmailer.php
```diff
@@ -840,6 +840,13 @@ protected function addAnAddress($kind, $address, $name = '')
...
+            // can't use $this->parseAddresses, because it calls validateAddress, which will fail here
+            list($U,$D) = explode('@',$address);
```
fbonzon: Like $this->parseAddresses that does `if ($useimap and function_exists('imap_rfc822_parse_adrlist'))`, I would first try to use here the builtin imap_rfc822_parse_adrlist() to extract the domain name, and then only use your own parser with explode().

Also, your parser should first verify that explode('@',$address) has exactly 2 elements. If there are less, it will produce a PHP Notice error for undefined $D. If there are more elements, the e-mail address is not valid…

michield: ok, will update the PR a little

michield: actually, parseAddresses and imap_rfc822_parse_adrlist handle email addresses with name etc. Here we only deal with an email address, so the explode is fine.

michield:
@Synchro Sorry,  I missed your comment. It's fine for BC, it's an addition in functionality, and will continue to work on PHP < 5.3 without problems

class.phpmailer.php
Synchro: That's good. I'd put `function_exists` first as it's probably a cheaper call than `hasMultiBytes`.

Exploding on `@` may run into other issues - it's horrible, but valid, to have an `@` within the local part if it's quoted, like `"a@b"@example.com`. One way around that is to reverse the address before exploding (i.e. split on the rightmost `@`) and limit the explode to 2 parts, then reverse the components.

Synchro:
Well, I'm sorry that I'd missed that you'd already wrapped it in a function check!

michield:
To test it, I had to manually compile postfix 3.0 (how retro :-) ) but it works:

Sep 30 20:47:39 debian8 postfix/local[18927]: AF73761C: to=<phplist@xn--phplst-6va.com>, relay=local, delay=0.09, delays=0.05/0.01/0/0.03, dsn=2.0.0, status=sent (delivered to mailbox)

michield:
ok, did some final optimisations, thanks for the feedback @Synchro and @fbonzon

fbonzon:
Another possible optimization: if the mailbox element (username) has multibytes characters but not the host element (domain), this new code does nothing. Is it cheaper to first split username/domain and then check if domain has multibytes characters?

michield:
@Synchro how do I force push a PR?

Synchro:
[The postfix docs](http://www.postfix.org/SMTPUTF8_README.html) mention that [RFC6530](http://tools.ietf.org/html/rfc6530) says the domain part may use either IDN or UTF-8 regardless of what the local part uses, so I think we should just leave the local part untouched for that distant future when we have more widespread SMTPUTF8 support.

Synchro:
@michield Rebase and squash all the commits you've done in here, then `git push -f` and it will replace your previous commits with one tidy rebased one. [Long version](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request).

michield:
ok, PR tidied up

class.phpmailer.php
```diff
@@ -840,6 +840,14 @@ protected function addAnAddress($kind, $address, $name = '')
             }
             return false;
         }
+        if (function_exists('idn_to_ascii') && $this->hasMultiBytes($address)) { ##@TODO: Write our own "idn_to_ascii" function for PHP 5.2 and earlier.
+            // can't use $this->parseAddresses, because it calls validateAddress, which will fail here
+            $elements = explode('@',strrev($address),2);
```
fbonzon: I rarely use strrpos(), but here IMHO it would make more readable and cheaper code than the triple strrev():

```php
if (false !== $pos = strrpos($address, '@')) {
    ++$pos;
    $address = substr($address, 0, $pos).idn_to_ascii(substr($address, $pos));
}
unset($pos);
```

fbonzon:
I tested the latest snapshot of this PR, and see two problems:

- As @Synchro mentioned, and specified in the `idn_to_ascii()` [documentation](http://www.php.net/idn_to_ascii) too, the domain name must be presented in UTF-8 encoding. If the web pages / PHP scripts are not in UTF-8 (i.e. you did not change the `CharSet` of your PHPMailer object to `'UTF-8'`), additional code is required to convert the domain name from the current CharSet to UTF-8, before passing it to `idn_to_ascii()`. FYI, the default CharSet in PHPMailer is `'iso-8859-1'`.

- `hasMultiBytes()` is not a good criterion to decide if `idn_to_ascii()` should be applied on the domain name or not. E.g. the non-ASCII upper 128 characters in ISO 8859-1 are single byte and should still be converted with `idn_to_ascii()`.

Synchro:
Nice one - it's so easy to forget about some PHP functions - I just ran into a use for `array_fill_keys` recently!

The PHPMailer `CharSet` property should have no real bearing on IDNs, since they always use UTF-8, even if the message does not, and they don't require Q/B encoding like regular headers because punycode is ASCII, however it's always good advice to tread carefully with UTF-8 anyway, and it's a good idea to apply a charset conversion to UTF-8 if `CharSet` is not the same.

Good call on `hasMultiBytes`; would be better to use `has8bitChars`.

fbonzon:
Agree. Thinking further about the two options:
- **A** Following RFC6530 spirit, require the address to be always specified in UTF-8 (whatever is PHPMailer `CharSet`) 
- **B** Apply a conversion to UTF-8 when `CharSet` is not the same

Compare the order:

```php
$mail->CharSet = 'UTF-8';
$mail->addAddress($email, $name);
$mail->Body = $message;
$mail->send();
```
```php
$mail->addAddress($email, $name);
$mail->Body = $message;
$mail->CharSet = 'UTF-8';
$mail->send();
```

Both are valid ways to use PHPMailer and make currently no difference, as all encoding (both `$name` and `$message`) happen inside `send()`.

With option B and current implementation inside the `addAnAddress()` method, we add a dependency in the order in which CharSet and addresses are specified. Shouldn't `idn_to_ascii()` calls rather happen inside `send()`?

In any case, PHPMailer user documentation should be made very clear. E.g.:

With A: IDN domains in e-mail addresses are allowed and MUST be passed in UTF-8 encoding, per RFC6530, even if PHPMailer CharSet property is different.

With B, current implementation in `addAnAddress()`: IDN domains in e-mail addresses are allowed. If different than the default 'iso-8859-1', PHPMailer CharSet property MUST be set before adding any address with an IDN domain, to ensure proper encoding.

My vote: B, but with encoding to punycode happening in `send()`. I don't have strong opinion though. A is good too, with proper documentation.

class.phpmailer.php
```diff
@@ -840,6 +840,14 @@ protected function addAnAddress($kind, $address, $name = '')
...
+            if (!empty($elements[0]) && !empty($elements[1])) {
+                $address = strrev($elements[1]).'@'.idn_to_ascii(strrev($elements[0]));
+            }
+            unset($elements);
+        }
         $address = trim($address);
```
fbonzon: Leave this line before the new code? Makes sense to trim before

Synchro:
I'm definitely for leaving things until `send`, at least partly because that's the only way to support SMTPUTF8 if we ever get to that, where we will need to implement [RFC6857](https://tools.ietf.org/html/rfc6857) too.

fbonzon:
Agree again. Do you have a preference for option A or B? With things moved in `send()` for B. For A, in addAnAddress() or send() doesn't matter.

About using `hasMultiBytes()` or `has8bitChars()`, maybe we shouldn't use either and simply always call `idn_to_ascii()`. Applying this function to a standard domain name in ASCII is no-op and does no harm. The function implementation in C probably already starts with checking if any encoding is required at all. So why re-do the same check ourselves in PHP beforehand?

@michield will you work further on this PR? I will work on it if you don't.

Synchro:
I'd vote for B. I think it has less likelihood of going wrong for users, especially since the default CharSet is not UTF-8.

michield:
@fbonzon feel free to take over. I'm happy to do it, but you seem to have stronger opinions on how.

fbonzon:
@michield OK, I'm taking over.

@Synchro Agreed B is best option. Implementing it requires the two following functional changes. Are you OK with them, and do you consider these backward compatibility issues?

1) Moment when `invalid_address` errors are reported is postponed to `send()`

These 5 public methods currently report an error and return `false` when given an invalid address:

```php
setFrom()
addAddress()
addCC()
addBCC()
addReplyTo()
```

Because we can't validate addresses with IDNs before we know what is PHPMailer `CharSet`, address validation and duplicates detection would happen in `send()` (more precisely, at the start of `preSend()`).

This would cause the 5 above methods to always return `true`. Code that currently checks the return value of these 5 methods would become useless. Note that in all example code, return value is always ignored.

`send()` on the other hand would either (subject to discussion):
- Stop and return `false` when an invalid address was previously passed to any of these 5 methods, and silently ignore duplicates. [My preference]
- Silently ignore invalid addresses and duplicates.

2) Access to addresses before calling `send()`

These 5 public methods currently always return validated and unique addresses, possibly before `send()`:

```php
getToAddresses()
getCcAddresses()
getBccAddresses()
getReplyToAddresses()
getAllRecipientAddresses()
```

Again because addresses would not be validated until `send()`, these 5 methods, when called before `send()`, would return either of (subject to discussion):
- Empty array. [My preference]
- Raw addresses, possibly invalid and with duplicates.

When called after `send()`, there is no change, i.e. only validated and unique addresses are returned.  None of these 5 methods are used in PHPMailer example code.

Synchro:
I think that the address setting functions (which mostly share the use of `addAnAddress`) should apply the current validation *if* the addresses do not contain IDNs, i.e. if we check them with `has8bitChars` first. This will preserve BC as the new behaviour will only apply with IDNs, which have not been supported before.

I agree with failing send if send-time address validation fails. It's an edge case really, and will only apply to IDN addresses.

It might be useful to extend the validation to allow pre-punycoded addresses when setting addresses, which would allow pre-send validation. That might be a useful workaround, perhaps add a public punyencode function (using whatever CharSet is currently set to).

I don't think I've ever seen a real-world use for any of the `get*` functions - PHPMailer instances are rarely kept around for long, you usually know what addresses you've put in, and re-use usually means you're going to call `clearAllAddresses` anyway. I don't have a strong opinion on how these should be handled, so long as it does something sensible, like you suggest.

fbonzon:
OK. Thanks for your suggestions.

About validation. Current rules already allow pre-punycoded addresses. See this [test case](https://github.com/PHPMailer/PHPMailer/blob/870350756b701743421dacc40126aa346251d0b1/test/phpmailerTest.php#L470).

Providing a convenience public punyencode function is good idea. I will do that. This will allow PHPMailer users, if they want, to convert IDNs to punycode themselves prior to adding addresses, and benefit from pre-send validation.

fbonzon:
This PR can be closed. Obsoleted by PR #529.

UA2018:
The Universal Acceptance Steering Group (UASG), supported by the Internet Corporation for Assigned Names and Numbers (ICANN), is a community-led, industry-wide initiative working on creating awareness and identifying and resolving problems associated with the universal acceptance of domain names. The purpose of these efforts is to help ensure a consistent and positive experience for Internet users globally.

You may find the following information useful in developing systems, internet applications that accept IDNs, EAI and New gTLDs.

Universal Acceptance Quick Guide:
https://uasg.tech/wp-content/uploads/documents/UASG005-en-digital.pdf

Good Practice Recommendations:
https://uasg.tech/wp-content/uploads/documents/UASG007-en-digital.pdf

Subscribe to the General Universal Mailing List https://mm.icann.org/mailman/listinfo/ua-discuss

You will find case studies and tools to test EAI on https://uasg.tech/.

PR #520, opened 2015-10-02
Change line endings from CRLF to LF

fbonzon:
All files in PHPMailer have LF line endings, except those two.
Harmonize line endings to LF for consistency.


Synchro:
Thanks.

PR #522, opened 2015-10-07
Do not use 'and' keyword in PHP

frederikbosch:
[Avoid using 'and'](http://stackoverflow.com/questions/2803321/and-vs-as-operator). It makes things too hard to understand, because of the different precedence than one would expect. It does not make a difference now, but it might prevent bugs in the future.

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -680,7 +680,7 @@ class PHPMailer
             return;
         }
         //Avoid clash with built-in function names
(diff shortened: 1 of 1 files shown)


Synchro:
No. This has come up before and it's largely a fallacy. None of the major frameworks or coding standards express a preference, and those that do are in favour of and/or. See #364.

frederikbosch:
@Synchro Hmm, I think you are right. My bad. Thanks for the quick reply.

PR #526, opened 2015-10-13
Improve some comments and minor improvements

fbonzon:
- Define an $error_message and reuse it, instead of recreating the
error message 3 times
- Use in_array() check instead of a regex

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -447,17 +447,6 @@ class PHPMailer
     public $XMailer = '';
 
     /**
-     * Only For XOAUTH - Google
-     * Options: An empty string for PHPMailer default, Enter the email used to get access token
-     * @var string
-     */
-//    public $UserEmail = '';
-//    public $RefreshToken = '';
-//    public $ClientId = '';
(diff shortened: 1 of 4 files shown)


fbonzon:
No functional changes here. I'd like to separate these changes from my coming up PR about IDNs support.

class.phpmailer.php
```diff
@@ -857,7 +849,7 @@ protected function addAnAddress($kind, $address, $name = '')
                 return true;
             }
         } else {
-            if (!array_key_exists(strtolower($address), $this->ReplyTo)) {
+            if (!isset($this->ReplyTo[strtolower($address)])) {
```
fbonzon: For consistency with isset() 6 lines above.

Synchro: While I know it *usually* works, I'm not keen on using `isset` to check if array elements exist since it breaks if you ever have null elements, so I tend to always use `array_key_exists`.

Synchro:
Thanks. Nice to see contributions with attention to detail!

PR #529, opened 2015-10-14
Allow addresses with IDN

fbonzon:
Accepts Internationalized Domain Name everywhere PHPMailler expects email addresses (To, CC, BCC, Reply-To, From, Sender and ConfirmReadingTo).

Requires PHP >= 5.3 with "intl" extension installed and "mbstring" extension enabled. Earlier versions don't see a change, i.e. specifying an address with IDN still fails validation.

Follow-up to PR #516
Ran test/phpmailerTest.php

Other changes:

- From, Sender and ConfirmReadingTo addresses are now validated in send(). Previously, only From and Sender addresses would be validated only if specified via the setFrom() method. ConfirmReadingTo was never validated.

- Half language strings for the 'invalid_address' message used colon at the end and half didn't. Harmonized messages to always include colon, and not add a second one with PHP code.


test/phpmailerTest.php
```diff
@@ -107,7 +107,6 @@ public function setUp()
             $this->Mail->Mailer = 'smtp';
         } else {
             $this->Mail->Mailer = 'mail';
-            $this->Mail->Sender = 'unit_test@phpmailer.example.com';
```
fbonzon: Duplicate to 6 lines above

Synchro:
This is great stuff. Could you rebase it on HEAD and force push to tidy up the multiple commits please?

fbonzon:
Done.

For own reference:
```
git reset --hard 6b12c212c40d31378e6250f26145c3c2c0445262
git merge --squash HEAD@{1}
git commit
git push --force
```

Synchro:
Beautiful, thanks!

michield:
Nice!

PR #530, opened 2015-10-14
Fix comments, suggested by scrutinizer

fbonzon:


Synchro:
Thanks, I'd not looked at scrutinizer for a while - it's often been broken.

Issue #525, opened 2015-10-13
GMail SMTP connect() failed


mitsus:
Hi all,
i'm try to send eamil using gmail with my raspberry Pi (it it into my Home Lan).
I try to use exactly this example https://github.com/PHPMailer/PHPMailer/blob/master/examples/gmail.phps, but i've these errors:

       # php -q test_gmail.php 
       Connection: opening to smtp.gmail.com:587, timeout=300, options=array (  &#039;ssl&#039; =&gt;   array (    &#039;verify_peer&#039; =&gt; false,    &#039;verify_peer_name&#039; =&gt; false,    &#039;allow_self_signed&#039; =&gt; true,  ),)<br>
       Connection: opened<br>
       SERVER -&gt; CLIENT: 220 smtp.gmail.com ESMTP uq5sm4060328wjc.3 - gsmtp<br>
       CLIENT -&gt; SERVER: EHLO mauNone<br>
       SERVER -&gt; CLIENT: 250-smtp.gmail.com at your service, [91.213.153.133]250-SIZE 35882577250-8BITMIME250-STARTTLS250-ENHANCEDSTATUSCODES250-PIPELINING250-     CHUNKING250 SMTPUTF8<br>
       CLIENT -&gt; SERVER: STARTTLS<br>
       SERVER -&gt; CLIENT: 220 2.0.0 Ready to start TLS<br>
       CLIENT -&gt; SERVER: EHLO mauNone<br>
       SERVER -&gt; CLIENT: 250-smtp.gmail.com at your service, [91.213.153.133]250-SIZE 35882577250-8BITMIME250-AUTH LOGIN PLAIN XOAUTH2 PLAIN-CLIENTTOKEN OAUTHBEARER  XOAUTH250-ENHANCEDSTATUSCODES250-PIPELINING250-CHUNKING250 SMTPUTF8<br>
       CLIENT -&gt; SERVER: AUTH LOGIN<br>
       SERVER -&gt; CLIENT: 334 VXNlcm5hbWU6<br>
       CLIENT -&gt; SERVER: bWl0c3VzZGV2QGdtYWlsLmNvbQ==<br>
       SERVER -&gt; CLIENT: 334 UGFzc3dvcmQ6<br>
       CLIENT -&gt; SERVER: OW1hZDltYXg5<br>
[...]

Issue #149, opened 2013-12-11
SMTP ERROR: Failed to connect to server: Network is unreachable (101) SMTP connect() failed


kilinkis:
hello guys. I need help please
this is my code:

``` php
require 'PHPMailerAutoload.php';
$mail = new PHPMailer;
$mail->isSMTP();
$mail->Host       = "smtp.gmail.com";
$mail->SMTPDebug  = 2;
$mail->SMTPAuth   = true;           
$mail->SMTPSecure = "ssl";
$mail->Port       = 465;  
$mail->Username = 'some@gmail.com';
$mail->Password = 'somepass';
$mail->addAddress('another@gmail.com', 'Josh Adams');
$mail->Subject = 'PHPMailer GMail SMTP test';
$body    = 'This is the HTML message body <b>in bold!</b>';
$mail->MsgHTML($body);
if (!$mail->send()) {
    echo "Mailer Error: " . $mail->ErrorInfo;
} else {
    echo "Message sent!";
}
```

and I get this error:
2013-12-11 15:15:02 SMTP ERROR: Failed to connect to server: Network is unreachable (101) SMTP connect() failed. Mailer Error: SMTP connect() failed.

any help please?

PR #531, opened 2015-10-16
Solving problems sending mail using SMTP

JoseCage:
In order to solve Bug #525 and #149 using phpmailer lib, changed the line 16 in the file located in PHPMailer/examples/gmail.phps, corresponding to the SMTP server configuration pasting $mail->Ismail(); instead of $mail->isSMTP(); and it worked perfectly in my testing, both on localhost as in remote server.

 failures already discussed in issues #525 and #149.

  With this hope I can help other programmers have passed the problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,8 @@ $mail = new PHPMailer;
 
 //$mail->SMTPDebug = 3;                               // Enable verbose debug output
 
+// $mail->isMail();											  // If isSMTP(); return error, use isMail();
+
 $mail->isSMTP();                                      // Set mailer to use SMTP
 $mail->Host = 'smtp1.example.com;smtp2.example.com';  // Specify main and backup SMTP servers
 $mail->SMTPAuth = true;                               // Enable SMTP authentication
--- a/examples/gmail.phps
+++ b/examples/gmail.phps
@@ -13,7 +13,12 @@ require '../PHPMailerAutoload.php';
 $mail = new PHPMailer;
 
 //Tell PHPMailer to use SMTP
-$mail->isSMTP();
+// $mail->isSMTP();
+
+$mail->isMail();
+
+// If isSMTP(); return error
+// set isMail(); and change the port address to 465
 
 //Enable SMTP debugging
 // 0 = off (for production use)
@@ -31,7 +36,9 @@ $mail->Host = 'smtp.gmail.com';
 // if your network does not support SMTP over IPv6
 
 //Set the SMTP port number - 587 for authenticated TLS, a.k.a. RFC4409 SMTP submission
-$mail->Port = 587;
+//$mail->Port = 587;
+
+$mail->Port = 465;
 
 //Set the encryption system to use - ssl (deprecated) or tls
 $mail->SMTPSecure = 'tls';


Synchro:
I'm sorry but this is an utterly clueless change. You're suggesting "fixing" SMTP by not using SMTP. It's not an appropriate solution to either of the bugs linked, and why would you alter a "how to use SMTP" example so that it does not use SMTP?

JoseCage:
Hello @Synchro, I understand you and you are quite right. I have reviewed my code and tested it again and confirmed that she was wearing the wrong way. Since using Ismail(); literally not the SMTP credentials necessary. With this I learned once more from my mistakes. So I hope suggestions. Thank you very much. :+1:

Synchro:
It's not really that - if you want to use SMTP and it's broken, fix whatever is wrong with you mail server or network - it's not a PHPMailer problem. Switching to `isMail` usually isn't an appropriate solution - it's like suggesting that the way to fix your broken-down car is to walk instead.

JoseCage:
Got it. Thank you once again for the explanation and excuse for failure.

  In case you need to delete this pull request in order to avoid possible mistakes (:bug:) by other developers, accepted without problems.

FalsAlarm:
I'm trying to use phpmailer with gmail for smtp mail settings. however doesn't matter what settings i try, still get GMail SMTP connect() failed

FalsAlarm:
Does anyone know how to fix that error?

FalsAlarm:
below is my code for the phpmailer and gmail smtp settings.

[code]
	public function SendEmailToUser($email, $subject, $message, $plainTextMessage)
	{
		//require_once './includes/mailer/class.phpmailer.php';
		
		/**
		 * This example shows settings to use when sending via Google's Gmail servers.
		 */
		//SMTP needs accurate times, and the PHP time zone MUST be set
		//This should be done in your php.ini, but this is how to do it if you don't have access to that
		date_default_timezone_set('Etc/UTC');
		//require '../PHPMailerAutoload.php';
		require_once("./includes/mailer/PHPMailerAutoload.php");
		//Create a new PHPMailer instance
		$mail = new PHPMailer;
		//Tell PHPMailer to use SMTP
		$mail->isSMTP();
		//Enable SMTP debugging
		// 0 = off (for production use)
		// 1 = client messages
		// 2 = client and server messages
		$mail->SMTPDebug = 2;
		//Ask for HTML-friendly debug output
		$mail->Debugoutput = 'html';
		//Set the hostname of the mail server
		$mail->Host = 'smtp.gmail.com';
		// use
		// $mail->Host = gethostbyname('smtp.gmail.com');
		// if your network does not support SMTP over IPv6
		//Set the SMTP port number - 587 for authenticated TLS, a.k.a. RFC4409 SMTP submission
		$mail->Port = 587;
		$mail->isHTML = false;
		//Set the encryption system to use - ssl (deprecated) or tls
		//$mail->SMTPSecure = 'tls';
		$mail->SMTPSecure = 'tls';
		//Whether to use SMTP authentication
		$mail->SMTPAuth = true;
		//Username to use for SMTP authentication - use full email address for gmail
		$mail->Username = "******@gmail.com";
		//Password to use for SMTP authentication
		$mail->Password = "********";
		//Set who the message is to be sent from
		$mail->setFrom('canadeals@gmail.com', 'Ricky Mossip');
		//Set an alternative reply-to address
		$mail->addReplyTo('canadeals@sasktel.net', 'Ricky Mossip');
		//Set who the message is to be sent to
		$mail->addAddress($email, 'First Nation Circle Member');
		//Set the subject line
		$mail->Subject = $subject;
		//Read an HTML message body from an external file, convert referenced images to embedded,
		//convert HTML into a basic plain-text alternative body
		$mail->msgHTML($message);
		//Replace the plain text body with one created manually
		$mail->AltBody = $message;
		//Attach an image file
		//$mail->addAttachment('images/phpmailer_mini.png');
		//send the message, check for errors
		if (!$mail->send()) {
			echo "<p>Mailer Error: " . $mail->ErrorInfo . "</p>";
		} else {
			echo "<h2>Message sent!</h2>";
		}
	}
[/code]

FalsAlarm:
by the way, the default code in gmail.phps file also produces this same error.

Synchro:
Don't hijack old, closed issues for unrelated requests. Read the docs that the error message points you at.

PR #532, opened 2015-10-18
Brackets in nameless recipient - $addBrackets flag

madolo:
Even if bracketless recipients are RFC 5322 compliant, them activate
SpamAssassin "*NO_BRKTS*" rules.

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -445,6 +445,13 @@ class PHPMailer
      * @var string
      */
     public $XMailer = '';
+	
+	/**
+     * Whether add angular brackets to nameless recipients
+     * to avoid SpamAssassin's "*NO_BRKTS*" rules
+     * @var boolean
+     */
+    public $addBrackets = false;
 
     /**
      * An instance of the SMTP sender class.
@@ -1730,6 +1737,7 @@ class PHPMailer
     public function addrFormat($addr)
     {
         if (empty($addr[1])) { // No name provided
+			if ($this->addBrackets) $addr[0] = '<'.$addr[0].'>';
             return $this->secureHeader($addr[0]);
         } else {
             return $this->encodeHeader($this->secureHeader($addr[1]), 'phrase') . ' <' . $this->secureHeader(


Synchro:
It's spamassassin that's wrong here. File a bug with them.

madolo:
Unfortunately SpamAssassin won't change its rules (there are some thread about that). To be more "friendly" we need add recipients names everytime.
Anyway, for my reference, which is the reason to not make this change?

Synchro:
PHPMailer's behaviour is entirely compliant with the applicable email RFCs. Spamassassin's is not. The workaround is trivial - set a recipient name yourself, even if it's the same as the email address, for example:

    $mail->addAddress('user@example.com', 'user@example.com');

This will result in a `To` header like this:

    To: user@example.com <user@example.com>

The `addOrEnqueueAnAddress` method trims the name value, so it won't let you use a space, but you can always override that, or put what's in your PR into a subclass - it's what libraries are for.

The only problem I have with doing this is that it's much more likely to trigger anti-phishing defences, which tend not to like seeing addresses or URLs in non-address or URL contexts. Fundamentally it's spamasassin's behaviour that is at fault - when it classifies RFC-compliant behaviour as spam, it's pressuring users to break RFCs, which is stupidly irresponsible, and frankly I'm unimpressed by their abusive bug reports.

madolo:
Got it. Probably I will override addrFormat() as it's the best path that I've found to change only the header filed without make problems to the recipient list at sending time.

Just to share my idea: the RFC5322 define the recipient as "\[...] (2) an addr-spec address enclosed in angle brackets ("<" and ">"). \[...]" - see https://tools.ietf.org/html/rfc5322#section-3.4
The bracketless notation is also accepted as simple form, so it's right that SA it's too strict, but it's also true that in this case they don't go against RFC.

Thanks again for the reviews!

Synchro:
No problem, and I reserve the right to change my mind on rejecting your PR once I've got over SA's abusive tactics... :wink:

PR #545, opened 2015-10-28
Updated Russian localization

Forst:
* Translated "Extension missing: "
* Replaced minus with an em-dash
* Removed duplicate "Extension missing: " stub in Galician
* Added "Extension missing: " stub to Hebrew


Synchro:
Thanks

PR #549, opened 2015-11-01
Upgrading comments and simplifying if expression

phelipealves:


Synchro:
Like many other projects, we've had a policy of preferring `if` over ternary operators, and in this case I don't think changing to the ternary operator helps.

The phpdoc type change on that oauth param is correct, but it will all be changing - see the xoauth branch.

PR #556, opened 2015-11-10
Fix unit tests to load the correct autoloader

tm1000:


Synchro:
Not loading vendor/autoloader.php breaks the loading of composer packages, the lines after it do the same as the PHPMailer autoloader on recent-ish PHP versions, which is all that's needed for Travis-CI, which is the ony reason this file exists. Also, the old autoloading system will be going away in 5.4 - see the 5.4 branch.

PR #557, opened 2015-11-11
Use a closure now that PHP 5.4 is min requirement

fbonzon:
Resurrects 0f24d1e67974926fd5fa316c3b13f89a73a3d8c1.


Synchro:
Thanks

PR #565, opened 2015-11-23
Fix PHP <5.4 compatibility

bd808:
Replace short array syntax (`[...]`) with `array(...)` to restore
compatibility with PHP versions before 5.4.

### DIFF
--- a/class.phpmaileroauthgoogle.php
+++ b/class.phpmaileroauthgoogle.php
@@ -50,10 +50,10 @@ class PHPMailerOAuthGoogle
     }
 
     private function getProvider() {
-        return new League\OAuth2\Client\Provider\Google([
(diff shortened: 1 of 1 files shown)


fbonzon:
This is on purpose. PHP 5.4 is minimum requirement to use the XOAUTH2 extension in PHPMailer. Compatibility down to PHP 5.0 is for the remaining of PHPMailer only.

From PHPMailer homepage in GitHub: "If you want to use Google's XOAUTH2 authentication mechanism, you need to be running at least PHP 5.4, and load the dependencies listed in composer.json."

Synchro:
Thanks

Synchro:
Eep! Sorry, I shouldn't have accepted this PR because of what @fbonzon said. I've reverted it. Not quite got my brain in gear after the weekend...

PR #569, opened 2015-11-24
Fix base64 encoding in alt body with long lines

bartbutler:
Extend body base64 fix from https://github.com/PHPMailer/PHPMailer/commit/9269a656ca2dca1ab2653c4abe7c0f796f31af33 to alt body.


Synchro:
Good call, thanks

bartbutler:
Thanks for all the good work!

Issue #568, opened 2015-11-24
get_oauth_token.php script only returning basic gmail access auth token


djc-developer:
Hi, I can't tell if this is a known problem, as the comment on the script *get_oauth_token.php* says not to mess around with the accessType.

**Problem**: In the troubleshooting page https://github.com/PHPMailer/PHPMailer/wiki/Using-Gmail-with-XOAUTH2#fetch-the-token

It says *get_oauth_token.php* should return an authorisation token for "viewing and managing gmail" with consent. But instead *get_oauth_token.php* keeps giving me "Basic account Info" authorisation token only. So no gmail modifcation authorisation, no authenticate token for gmail account smtp access, etc.

*The refresh token used for XOAUTH2 during SMTP login is probably the reason for so many authentication PASSWORD/USERNAME SMTP failures, when really it's just a bad auth token* 

*Included is a dump of my debug
I'm guessing the SMTP authenticate is failing because the proper authorisation token is not working. If I'm wrong, I'd like any kind of feedback. I may be doomed to just go the basic route.*

**PS, I removed the AUTH key because of the base64 encoding of my gmail account**

*SERVER -> CLIENT: 220 smtp.gmail.com ESMTP rc5sm11791624pbc.95 - gsmtp
CLIENT -> SERVER: EHLO localhost
SERVER -> CLIENT: 250-smtp.gmail.com at your service, [58.6.216.202]250-SIZE 35882577250-8BITMIME250-STARTTLS250-ENHANCEDSTATUSCODES250-PIPELINING250-CHUNKING250 SMTPUTF8
CLIENT -> SERVER: STARTTLS
SERVER -> CLIENT: 220 2.0.0 Ready to start TLS
CLIENT -> SERVER: EHLO localhost
SERVER -> CLIENT: 250-smtp.gmail.com at your service, [58.6.216.202]250-SIZE 35882577250-8BITMIME250-AUTH LOGIN PLAIN XOAUTH2 PLAIN-CLIENTTOKEN OAUTHBEARER XOAUTH250-ENHANCEDSTATUSCODES250-PIPELINING250-CHUNKING250 SMTPUTF8
CLIENT -> SERVER: AUTH XOAUTH2 [AUTH REMOVED BY DJC-Developer]
[...]

PR #579, opened 2015-12-07
Resolved Scope issues which resulted in GMail Auth failure

sherryl4george:
This is to solve issue #568 . 

As changes have been made to the OAuth Library for Google, the scope was set to default all times. This has been overridden now by writing a custom class into the _get_oauth_token.php_ file. Also changes have been made to _composer.json_.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,8 +27,7 @@
         "phpunit/phpunit": "4.7.*"
     },
     "suggest": {
-        "league/oauth2-client": "Needed for XOAUTH2 authentication",
-        "league/oauth2-google": "Needed for Gmail XOAUTH2"
+        "league/oauth2-client": "Needed for XOAUTH2 authentication"
     },
     "autoload": {
         "classmap": [
--- a/get_oauth_token.php
+++ b/get_oauth_token.php
@@ -12,8 +12,20 @@
  * This script requires PHP 5.4 or later
  * PHP Version 5.4
  */
+/**
+ * Added a new class for getting the Refresh Token with right scopes
+ * as the OAuth-Client for Google from GitHub didnot provide setting the 
+ * scope out of the box
+ */
+
+namespace League\OAuth2\Client\Provider;
 
 require 'vendor/autoload.php';
+
+use League\OAuth2\Client\Provider\Exception\IdentityProviderException;
+use League\OAuth2\Client\Token\AccessToken;
+use League\OAuth2\Client\Tool\BearerAuthorizationTrait;
+use Psr\Http\Message\ResponseInterface;
 
 session_start();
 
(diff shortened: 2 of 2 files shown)


Synchro:
Thanks very much for doing this. I find oauth horribly confusing and would never have figured this out!

Since you obviously have a clue about this, could you please take a look at the 5.4 and xoauth branches? The 5.4 branch is a big cleanup (it does the same thing as your PR in composer.json) and renames everything in a PSR-4 style, and the xoauth branch adds an xoauth abstract base class and supports more auth providers. The get_auth_token script will need altering to make that work too. Ideally I'd like to roll all that together for the 5.4 release, but I could use some help.

sherryl4george:
I am more than glad to help you. 

Can you please guide me on what I should be doing. If we are extending OAuth for more providers, which all are we targeting at? And other things should I do further

Synchro:
Someone else contributed some implementations for Yahoo and Microsoft (Outlook etc), but it's the general pattern that I want to get nailed down. I was thinking about merging them into 5.2, but that's really not sensible, so they are going into 5.4.

I have done a rough merge of all recent changes into the [5.4](https://github.com/PHPMailer/PHPMailer/tree/5.4) and [xoauth](https://github.com/PHPMailer/PHPMailer/tree/xoauth) branches. The xoauth branch is 5.4 + the new providers, and will eventually be merged back into 5.4, so I'd recommend looking at both of those. You'll see I've done a basic base class for the xoauth providers, but it will need corresponding changes in the token script, and possibly some other utility scripts to help use them.

Because my grasp of oauth is limited, I've been flying blind as far as the implementations go, and I have no idea how you go about testing them.

sherryl4george:
Let us take this forward 
I will provide the required coding latest by this weekend. Please let me know if there are more services to be added

Synchro:
Great, thanks. Don't worry about adding more services - those are enough for now, and more can be added later. I just want to be sure we have a solid base to build from, and the three we have will be plenty for that.

Synchro:
Did you get anywhere with this?

sherryl4george:
I was working on this. Below are my findings:
1. Inorder to completely eliminate the use of ClientProviders like Google,
Yahoo for OAuth, we have to create classes which does the
 functionality for us. These can be placed inside a seperate folder in
OAuthProvider.
2.

Regards,
Sherryl George
+91 7036858184

sherryl4george:
I was working on this using the branch 5.4. Below are my findings:
1. Inorder to completely eliminate the use of _ClientProviders_ like Google, Yahoo for OAuth from composer, we have to create classes which does the functionality of those classes for us. These can be placed inside a seperate folder in _OAuthProvider_. The phpleauge/oauth-client will still be used as the base.
2. While creating the OAuth instance, how can one distinguish between GMail, Yahoo, Microsoft or anyother.
* There should be a new variable( say a integer for each service ) added to PHPMailerOAuth class which indicate the service provider being used. This has to be set by the user
* Pass the OAuth instance of the service provider itself. It adds the overhead to the end user to create the instance. Not very advisable.
*  According to the SMTP passed, the provider has to be figured out using some logic in the _getOAUTHInstance()_ method of _PHPMailerOAuth_ class. Incase of Microsoft, will we end up having more than one SMTP addresses.

In all above cases when a new provider is added, code changes are required into the _PHPMailerOAuth_ class.
Please share your views & suggest on this, so that we can go ahead.

Synchro:
Eek! This is why I pointed you at the [xoauth branch](https://github.com/PHPMailer/PHPMailer/tree/xoauth) where I've already done quite a lot of what you describe!

I don't want to use a fixed enumeration for the providers as it makes maintaining them difficult, so I would prefer to do what you suggest in your second option, so users would need to do something like:

```php
$mail->setOAuthProvider(new OAuthProvider\Google($mail));
```

This is more of a dependency injection approach and is much easier to maintain.

Issue #583, opened 2015-12-11
Add a method to allow changing of RCPT command, leaving headers intact


kruisdraad:
This ticket is a preface for a pull request being submitted

Add a method to allow changing of RCPT command, leaving headers intact. This method (envelop_to) would be used if you want to 'redirect' an email to another address without changing it. So the original TO and other headers will remain intact.

Example:

You have a raw e-mail intended for someone@example.com, and you want to redirect it do third@example.com, leaving the original TO intact to mail clients would keep their display of the original mail. The code would need to do:

```
2015-12-11 11:01:52 SERVER -> CLIENT: 220 ar3.ip6.nu ESMTP Postfix (Ubuntu)
2015-12-11 11:01:52 CLIENT -> SERVER: EHLO ar3.ip6.nu
2015-12-11 11:01:52 SERVER -> CLIENT: 250-ar3.ip6.nu
                                      250-PIPELINING
                                      250-SIZE 10240000
                                      250-VRFY
                                      250-ETRN
                                      250-STARTTLS
                                      250-ENHANCEDSTATUSCODES
                                      250-8BITMIME
                                      250 DSN
2015-12-11 11:01:52 CLIENT -> SERVER: STARTTLS
2015-12-11 11:01:52 SERVER -> CLIENT: 220 2.0.0 Ready to start TLS
2015-12-11 11:01:52 CLIENT -> SERVER: EHLO ar3.ip6.nu
2015-12-11 11:01:52 SERVER -> CLIENT: 250-ar3.ip6.nu
                                      250-PIPELINING
                                      250-SIZE 10240000
                                      250-VRFY
                                      250-ETRN
                                      250-ENHANCEDSTATUSCODES
                                      250-8BITMIME
                                      250 DSN
2015-12-11 11:01:53 CLIENT -> SERVER: MAIL FROM:<sender@example.com>
2015-12-11 11:01:53 SERVER -> CLIENT: 250 2.1.0 Ok
2015-12-11 11:01:53 CLIENT -> SERVER: RCPT TO:<third@example.com>
2015-12-11 11:01:53 SERVER -> CLIENT: 250 2.1.5 Ok
2015-12-11 11:01:53 CLIENT -> SERVER: DATA
2015-12-11 11:01:53 SERVER -> CLIENT: 354 End data with <CR><LF>.<CR><LF>
2015-12-11 11:01:53 CLIENT -> SERVER: Date: Fri, 11 Dec 2015 12:01:52 +0100
[...]

PR #584, opened 2015-12-11
allow to override the envelope_to (rcpt), but leave the to headers in…

kruisdraad:
…tact for display purposes. This is needed if you want to 'redirect' an email to another address without changing it.

linked to issue #583


Synchro:
@fbonzon What do you think of this?

class.phpmailer.php
```diff
@@ -1474,8 +1496,8 @@ protected function smtpSend($header, $body)
         }
 
         // Attempt to send to all recipients
-        foreach (array($this->to, $this->cc, $this->bcc) as $togroup) {
-            foreach ($togroup as $to) {
+        if (!empty($this->envelope_to)) {
+            foreach ($this->envelope_to as $to) {
```
fbonzon: If I read this correctly: as soon as I specify at least 1 "envelope_to" recipient, then the mail is delivered only these "envelope_to" recipients, and none of the "to", "cc", and "bcc" recipients? This should be made clear in the code comments here, and elsewhere where "envelope_to" recipients appear.

In the same PR, a new or modified example in the "examples" directory would be welcome too.

class.phpmailer.php
```diff
@@ -461,6 +461,13 @@ class PHPMailer
     protected $to = array();
 
     /**
+     * The array of 'to' names and addresses.
```
fbonzon: s/to/envelope_to/

class.phpmailer.php
```diff
@@ -1485,6 +1507,20 @@ protected function smtpSend($header, $body)
                 }
                 $this->doCallback($isSent, array($to[0]), array(), array(), $this->Subject, $body, $this->From);
             }
+
```
fbonzon: remove blank line

class.phpmailer.php
```diff
@@ -908,10 +926,14 @@ protected function addAnAddress($kind, $address, $name = '')
             }
             return false;
         }
+
```
fbonzon: remove blank line

Synchro:
Thanks - I was really meaning about the overall concept. My first thought as the way to do this was to use PHPMailer to set up, use `getSentMIMEMessage()` to get the message, then implement an SMTP transport to send it, since everything envelope-related happens down at the SMTP level. This way is easier to use though, so I think I'm fairly happy to let it in.

class.phpmailer.php
```diff
@@ -908,10 +926,14 @@ protected function addAnAddress($kind, $address, $name = '')
...
         if ($kind != 'Reply-To') {
             if (!array_key_exists(strtolower($address), $this->all_recipients)) {
                 array_push($this->$kind, array($address, $name));
-                $this->all_recipients[strtolower($address)] = true;
```
fbonzon: By not adding "envelope_to" recipients in the $this->all_recipients array, you allow duplicates in the "envelope_to" array. You need to add a mechanism to disallow duplicates in the "envelope_to" array.

class.phpmailer.php
kruisdraad: Thank you for your replies.  That is correct and intended as per RFC.  Currently i am sick and will add an example when i am back again. My example would be an implementation snippet.

// Make mailer object using SMTP method
$mail = new PHPMailer;
$mail->isSMTP();
$mail->SMTPDebug = 2;

// Bypass actuall receipiant(s)
$mail->addEnvelopeTo('bart@vrancken.eu', 'Bart');

// Enforce these headers with their original values
foreach($headers['from'] as $from) {
    $mail->setFrom($from['address'], $from['name']);
}
foreach($headers['to'] as $to) {
    $mail->addAddress($to['address'], $to['name']);
}
$mail->Subject = !empty($headers['subject']) ? $headers['subject'] : '';
$mail->Body = !empty($body) ? $body : '';
$mail->MessageID = !empty($headers['message-id']) ? $headers['message-id'] : '';
$mail->MessageDate = !empty($headers['date']) ? $headers['date'] : date('D, j M Y H:i:s O');
$mail->ContentType = !empty($headers['content-type']) ? str_replace(';', '', $headers['content-type']) : '';
$mail->XMailer = 'AbuseIO Failed Mail Sender';

// Add headers accourding to RFC 2822 section 3.6.6.
$mail->addCustomHeader('Resent-From', $resendFrom);
$mail->addCustomHeader('Resent-To', $resendTo);
$mail->addCustomHeader('Resent-Date', $mail->buildMessageDate());
$mail->addCustomHeader('Resent-Message-Id', $mail->buildMessageID());
$mail->addCustomHeader('Resent-User-Agent', 'AbuseIO Failed Mail Sender');


if (!$mail->send()) {
    echo "Mailer Error: " . $mail->ErrorInfo;
} else {
    echo "Message sent!";
}

fbonzon: > foreach($headers['from'] as $from) {
> $mail->setFrom($from['address'], $from['name']);
> }

There can be only one sender. Fix your code to have a scaler in $headers['from'], or arbitrary select the first item in array $headers['from'].

fbonzon:
@Synchro See my comments about the overall concept in the issue. I was writing them simultaneously to your reply :)

fbonzon:
Comments for $RecipientsQueue line 503 would need update too. "envelope_to" addresses with IDN would be enqueued there.

Luisinho77:
Greetings, 

I just wanted to say  HI and  tell you thet you're amazing! Please  take a look  at some stuff I've just found  <http://together.mylanpro.com/e4zuzh>

Best regards, ljlorente

fjaeger:
What's the reason for this merge request to be closed eventually? I am looking for exactly this kind of mechanism of specifying "envelope_to" and is also conforms to the RFC. If required, I would like to help ironing out remaining issues.

Synchro:
@fjaeger I think it complicates things quite a lot for a tiny edge case that's better dealt with at the SMTP level. See my closing comment on #607, which lets you achieve the same thing very simply.

fjaeger:
Is there an option to use a SMTP subclass with PHPMailer or would I also use a PHPMailer subclass?

Synchro:
You need to override PHPMailer too (BC issues). There's [an article in the wiki](https://github.com/PHPMailer/PHPMailer/wiki/Overriding-the-SMTP-class) that shows how.

Synchro:
I've just pushed a change to 6.0 that means you don't need to subclass PHPMailer any more, you can just inject an instance directly.

fjaeger:
Awesome, thanks!

PR #591, opened 2015-12-21
XOAUTH2 for Google Yahoo Microsoft

sherryl4george:
Changes to incorporate XOAUTH2 for Google, Microsoft and Yahoo has been finsihed
1. New Providers have been created to remove the dependencies on external libraries. The providers are in OAuthProvider/Provider
2. Changes have been made to _get_oauth_token.php_ to account for the newly added providers
3. Slight changes to _gmail_example.phps_
<br><br>
* Yahoo Mail Service OAUTH2 is not working as of now. I am not sure whether this is specific to me, but I was not able to set scopes for Web Mail Service when creating the app in Yahoo!. If someone has this working please do help in testing it as I have not tested Yahoo Mail sending. _[https://developer.yahoo.com/mail/]_
* Request Marcus to review the code and rename the namespaces and files wherever necessary as I was bit unsure about naming, especially inside the OAuthProvider folder
<br>
* Please find the attached file for more details

[oauth.txt](https://github.com/PHPMailer/PHPMailer/files/68884/oauth.txt)

### DIFF
--- a/test/phpmailerTest.php
+++ b/test/phpmailerTest.php
@@ -607,7 +607,7 @@ class PHPMailerTest extends PHPUnit_Framework_TestCase
             'first.last@[IPv6:a1:a2:a3:a4:b1:b2:b3:]',
             'first.last@[IPv6::a2:a3:a4:b1:b2:b3:b4]',
             'first.last@[IPv6:a1:a2:a3:a4::b1:b2:b3:b4]',
-            "(\r\n RCPT TO:websec02@d.mbsd.jp\r\n DATA \\\nSubject: spam10\\\n\r\n Hello,\r\n this is a spam mail.\\\n.\r\n QUIT\r\n ) a@gmail.com" //This is valid RCC5322, but we don't want to allow it
+            "(\r\n RCPT TO:user@example.com\r\n DATA \\\nSubject: spam10\\\n\r\n Hello,\r\n this is a spam mail.\\\n.\r\n QUIT\r\n ) a@example.net" //This is valid RCC5322, but we don't want to allow it
         );
         // IDNs in Unicode and ASCII forms.
         $unicodeaddresses = array(


composer.json
```diff
@@ -27,9 +27,7 @@
         "phpunit/phpunit": "4.*"
     },
     "suggest": {
-        "league/oauth2-client": "Needed for XOAUTH2 authentication",
-        "hayageek/oauth2-yahoo": "Needed to use XOAUTH2 with Yahoo",
-        "stevenmaguire/oauth2-microsoft": "Needed to use XOAUTH2 with Microsoft services"
+        "league/oauth2-client": "Needed for XOAUTH2 authentication"
```
Synchro: This is a dependency of all the specific implementations, so specifying it by itself isn't useful.

Synchro: Ah, ok, I didn't see that you had reimplemented these so that we do only need that dep - sorry!

Synchro:
We already have the `OAuthProvider` namespace, so I don't think we need to go further info `OAuthProvider\Provider` **and** put `Provider` after the name of every class. `PHPMailer\PHMailer\OAuthProvider\Provider\YahooProvider` is unnecessarily long; I think all we need is `PHPMailer\PHPMailer\OAuthProvider\Yahoo`.

Synchro:
All those shell scripts need to be executable.

sherryl4george:
So, now how should we go further? What changes need to be reverted?

Synchro:
Can these two sets of files be combined? Would it make sense to make the `Base` class extend `AbstractProvider` so that all that functionality is available at that level? It really looks like they should use the base class as there is a lot of duplication.

I would have a go at it, but I'm not clear why they are separate, so could you either do the combining or explain why we should not?

BTW I merged in some changes from trunk and the 5.4 branch you may want to pull too.

sherryl4george:
Can you please check and let me know whether further changes are needed.

* One drawback I can see is If we implement for 100 providers then 100 new
classes extending _
Base_ has to be created. Would that be a good idea to go with. Or adding
dependencies in composer would be good

On Tuesday, December 22, 2015, Marcus Bointon <notifications@github.com>
wrote:

> Can these two sets of files be combined? Would it make sense to make the
> Base class extend AbstractProvider so that all that functionality is
> available at that level? It really looks like they should use the base
> class as there is a lot of duplication.
>
> I would have a go at it, but I'm not clear why they are separate, so could
> you either do the combining or explain why we should not?
>
> BTW I merged in some changes from trunk and the 5.4 branch you may want to
> pull too.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/PHPMailer/PHPMailer/pull/591#issuecomment-166589076>.
>


-- 
Regards,
Sherryl George
+91 7036858184

PR #592, opened 2015-12-22
XOAUTH2 Google Microsoft - 2

sherryl4george:
* _Base_ inherits _AbstracProvider_ now
* Yahoo files not included as YMail API is not working as of now

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "phpunit/phpunit": "4.*"
     },
     "suggest": {
-        "league/oauth2-client": "Needed for XOAUTH2 authentication"
+        "league/oauth2-client": "Needed for XOAUTH2 authentication"        
     },
     "autoload": {
         "psr-4": {
--- a/examples/gmail_xoauth.phps
+++ b/examples/gmail_xoauth.phps
@@ -8,7 +8,6 @@ namespace PHPMailer\PHPMailer;
 //This should be done in your php.ini, but this is how to do it if you don't have access to that
 date_default_timezone_set('Etc/UTC');
 
-require '../vendor/autoload.php';
 
 //Load dependencies from composer
 //If this causes an error, run 'composer install'
@@ -24,7 +23,7 @@ $mail->isSMTP();
 // 0 = off (for production use)
 // 1 = client messages
 // 2 = client and server messages
-$mail->SMTPDebug = 0;
+$mail->SMTPDebug = 3;
 
 //Ask for HTML-friendly debug output
 $mail->Debugoutput = 'html';
@@ -48,19 +47,26 @@ $mail->AuthType = 'XOAUTH2';
 $mail->oauthUserEmail = "someone@gmail.com";
 
 //Obtained From Google Developer Console
-$mail->oauthClientId = "RANDOMCHARS-----duv1n2.apps.googleusercontent.com";
+$mail->oauthClientId = "0************74";
 
 //Obtained From Google Developer Console
-$mail->oauthClientSecret = "RANDOMCHARS-----lGyjPcRtvP";
+$mail->oauthClientSecret = "KS***************aU";
 
 //Obtained By running get_oauth_token.php after setting up APP in Google Developer Console.
 //Set Redirect URI in Developer Console as [https/http]://<yourdomain>/<folder>/get_oauth_token.php
 // eg: http://localhost/phpmail/get_oauth_token.php
-$mail->oauthRefreshToken = "RANDOMCHARS-----DWxgOvPT003r-yFUV49TQYag7_Aod7y0";
+$mail->oauthRefreshToken = "MCeW******************mxyQ$$";
+
+$option = [ 'mailObj' => $mail ];
+
+//This has to be set 
+$mail->setOAuth(new OAuthProvider\Google($option));
+//$mail->setOAuth(new OAuthProvider\Yahoo($option));
+//$mail->setOAuth(new OAuthProvider\Microsoft($option));
 
 //Set who the message is to be sent from
 //For gmail, this generally needs to be the same as the user you logged in as
-$mail->setFrom('from@example.com', 'First Last');
+$mail->setFrom('someone@gmail.com', 'First Last');
 
 //Set who the message is to be sent to
 $mail->addAddress('whoto@example.com', 'John Doe');
--- a/examples/gmail_xoauth.phps
+++ b/examples/gmail_xoauth.phps
@@ -23,7 +23,7 @@ $mail->isSMTP();
 // 0 = off (for production use)
 // 1 = client messages
 // 2 = client and server messages
-$mail->SMTPDebug = 3;
+$mail->SMTPDebug = 0;
 
 //Ask for HTML-friendly debug output
 $mail->Debugoutput = 'html';
--- a/get_oauth_token.php
+++ b/get_oauth_token.php
@@ -1,167 +1,123 @@
 <?php
 /**
- * Get an OAuth2 token from Google.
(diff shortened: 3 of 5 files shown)


sherryl4george:
Can you please check and let me know whether further changes are needed.

* One drawback I can see is If we implement for 100 providers then 100 new
classes extending _Base_ has to be created. Would that be a good idea to go with. Or adding
dependencies in composer would be good. Eventually adding more providers would make PHPMailer heavy.

Another thought can be to maintain the _OAuthProvider_ classes in a separate repo or similar kind of thing that can be maintained by us

Please suggest.

sherryl4george:
@synchro Can you suggest on what changes has to be made to this?

Synchro:
Can you take a look at the [xoauth branch](https://github.com/PHPMailer/PHPMailer/tree/xoauth)? I've pushed all my changes to it, but I'm still having trouble getting anything to work, the league oauth classes made some BC-breaking changes that I don't understand, and the oauth class structure is still not pretty.

In your fork, you've got a different implementation of a google class inside get_auth_token - there's no way we should be doing that; it should be part of our base classes we've already got.

As you say, we don't really want to include too many implementations, but I think a few key ones is ok, and the abstract base means it's easier to provide compatible implementations that can be injected easily, possibly from other composer packages.

Synchro:
Why did you close this? I still need your help on this, and it's not usable at the moment, plus it's the main thing blocking the next PHPMailer release.

sherryl4george:
I am really sorry for closing the request. I was under the assumption that everything was working fine.
I was going through the code XOAUTH branch. I could find a little bit of issues. I would like to get clarifications on the following:
1. Are we going to use external dependency like league/oauth2-google as the OAUTH Client or do we need a class implementation of our own
2. All data like ClientId, ClientSecret etc for successful OAUTH is contained in objects of classes like Google. How are we passing it to the PHPMailer class, since we are planning to go ahead with only one PHPMailer main class. In this commit I have passed it as a object, with all the data in it to the method setOAuth of PHPMailer class.

Synchro:
I think oauth is too complex to roll our own, and I don't want to maintain one, so I'm happy to have an external dependency for now, and the League packages seem to be well regarded.

Rather than putting all the oauth properties in the PHPMailer class, I was proposing to put them all in the oauth abstract base class, since they are not needed at any other time. As far as usage goes, I think we should inject the oauth object, something like:

    $mail->AuthType = 'xoauth2';
    $mail->setOAuth(new OAuthProvider\Google);

That way all the oauth properties can stay inside the oauth class and don't pollute the global space, and we're not quite so tied to the underlying implementation. We can either provide a config array to the constructor, or allow setting properties inside the oauth instance via accessors.

This is approximately what I have been doing in the xoauth branch, but it's not really been tested, largely because I can't get the oauth token scripts to work.

I also propose to put all the important things that are needed by the `get_oauth_token` script in the same classes so that one class deals with both getting and using a token. I've not tried to do that yet.

Synchro:
Just to clarify - I'm not using the League classes directly from the PHPMailer class, only within the abstract base class, so we do have a kind-of wrapper to abstract the implementation a bit.

sherryl4george:
If we use League classes like (`LeagueGoogle`, `SteveMicrosoft` etc.), then most part of the `Base` class will not be used, as the League classes are self-contained and the code in `Base` will be redundant. We can actually convert `Base` into a `trait` since the OAuth variables and methods taken out of PHPMailer is contained in `Base`. There is a method argument for `setOAuth` which is of type of `Base` class. This would have to be changed to some other type or without any type in case we are going this way.

Synchro:
I think we will still need to use some kind of wrapper, even if it's very thin - it's better to code to an interface than an implementation even if we have no particular plan to use another implementation. Otherwise it gives us little opportunity to work around peculiarities of particular services while keeping a consistent interface, or if perhaps there are issues with a particular League class implementation that we need to override. I also don't see how we could integrate getting tokens without a wrapper since that does not seem to be in the League classes at all.

sherryl4george:
So what do you suggest is to keep the `Base` class intact? In that case there will be more method additions to the `Google` class which is needed to provide Authorization URL and other methods, which would more likely look like the code in this pull. I dont think a code cleanup like the on in XOAUTH branch has a chance to work. Also getting tokens wont be a problem as far as I see. If we code `Base` extending `AbstractProvider` then there will be no use of League classes as we can override all necessary methods of `AbstractProvider` and make it work.

Synchro:
I do prefer a shallower `$x-> x` to  `$x->x->x` where possible, and doing what you suggest would help that, i.e. make the base class a subclass of the league base class rather than a wrapper, reducing boilerplate in the base class. Would that not mean that we run into inheritance problems because League's own classes extend its own base class and not ours?

The code at the moment is someone half-baked - `OAuthProvider\Base` *is* `AbstractProvider`. What I do like about the current structure is that the specific implementations are very small, even when the providers are completely different.

I do have a bit of a motivation problem with oauth in general. There are so many things wrong with it I hardly know where to start, but it's a necesary evil that needs to be done for PHPMailer!

Incidentally someone pointed out that League's classes now require PHP 5.5. as a minimum, so the same will apply to PHPMailer.

sherryl4george:
So just for a clarification: How will the class be now `Base extends AbstractProvider` and `Google extends Base` or directly implement `Google` from LeagueClient? The former is already implemented in this Pull, but as you can see needs a lot of redundant code. The latter is a simpler approach. Please clarify.
Yes it is true that League needs PHP5.5 as a minimum, which is sad to hear. PHPMailer would also go to 5.5 which is not a good thing.

Synchro:
It would definitely be better to be able to use the league provider classes directly, then we don't need to bundle and maintain our own implementations at all.

Is there some way we can implement this with a single class in PHPMailer? I'm seeing it heading in this direction:

```php
$provider = new League\OAuth2\Client\Provider\Google([
    'clientId'     => '{google-app-id}',
    'clientSecret' => '{google-app-secret}',
    'redirectUri'  => 'https://example.com/callback-url',
    'hostedDomain' => 'example.com',
]);
$mail->setOAuth(new PHPMailer\OAuth($provider));
```
Is that too ugly?

sherryl4george:
I don't find a the above piece of code to be ugly. But would it be difficult for the user to provide the long League class names. Different providers have different naming. Someone not aware of the names can easily break it.
We can go with one class each for each provider. To avoid the declaration OAuth variables and `getOAuth64` method in each class and to keep it simple we can use a `trait`.

Synchro:
I don't think long class names alone is a good enough reason to have to provide wrappers for everything - `PHPMailer\PHPMailer\OAuthProvider\Google` isn't really much shorter, and if someone wants to use a provider we don't have, they would have to write a wrapper to use it first, which is a far more complicated thing than typing a single long name. I'd say long names are a mostly unavoidable consequence of having namespaces, and it's pretty normal in other things - look at Zend Framework or Laravel. PHPMailer itself is already `PHPMailer\PHPMailer\PHPMailer`! You can always hide/qualify long names with `use` statements anyway.

sherryl4george:
Then it has pretty much boiled down to using the league classes directly to create the provider and use another generic class to pass all details and set the `OAuth64` string. This class should accept `Email Id` and `Refresh Token` also as parameters along with the provider object. There would be some overhead at the user end of declaring proper scopes(different for each provider) when obtaining refresh tokens. Shall we finalize this and go ahead?

Synchro:
I think that sounds good - I don't see extra params would be much of an issue; we can fix a fair amount with documentation.

sherryl4george:
Can we close this pull as the code in this is no more needed for us.

Synchro:
OK, closing

PR #597, opened 2015-12-27
Favicon

Senasum:
Created a favicon to use with example and other purposes


Synchro:
Sorry, I know that you're just trying to get credits for a course, but there's no point in adding unnecessary weight to the project like this.

Senasum:
It's okay, thank you for your feed back. I understood your point.

PR #607, opened 2016-01-15
Ability to not send to specific recipients using SMTP mailer

bartbutler:
This allows one to optionally designate certain recipients as "do-not-send-to-this-address" addresses while keeping the addresses in the mail headers. It only works with the SMTP mailer as far as I can tell--sendmail and mail do not allow this.

The use case is specialized but powerful--one can strip out 'internal' recipients if sending mail to recipients which are part of an internal system before passing to the MTA while keeping them in the headers. One can also send different versions of the message to different recipients, which is useful when it comes to supporting different or mixed encryption methods.

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -2539,7 +2539,8 @@ class PHPMailer
                 4 => $type,
                 5 => false, // isStringAttachment
                 6 => $disposition,
-                7 => 0
+                7 => '',
+                8 => ''
             );
 
         } catch (phpmailerException $exc) {
@@ -2575,6 +2576,7 @@ class PHPMailer
         // Return text of body
         $mime = array();
         $cidUniq = array();
+        $locationUniq = array();
         $incl = array();
 
         // Add all attachments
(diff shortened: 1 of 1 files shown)


Synchro:
This looks very similar to #584, but implemnted in a different way.

bartbutler:
Yeah, it is similar. I didn't see the earlier one. Either would work for my purposes. I'm not sure whether a separate envelope_to list or having a flag in the recipient adding functions is 'cleaner' from an interface perspective. I also noticed that changing the iteration from $this->to, $this->cc, and $this->bcc to all_recipients might suppress multiple messages to recipients listed in both To and CC, for example. That's easily fixable though--I'll do that now. Let me know what you decide to do.

bartbutler:
I'm trying to find the relevant RFC for what should happen if a recipient is in To and CC simultaneously, two messages received or one? Gmail seems to do one, in which case iterating over all_recipients is fine.

bartbutler:
It might be worth noting that this version, using all_recipients, is a much more minimal change to the codebase, which probably speaks in it's favor.

fbonzon:
Actually, your PR would not change the behaviour about a recipient  in To and CC simultaneously, because [this line](https://github.com/PHPMailer/PHPMailer/blob/47f66d3aacd65a14023afc2bd7b96398852de4e4/class.phpmailer.php#L912):

```php
if (!array_key_exists(strtolower($address), $this->all_recipients)) {
```

Relevant is only the first call to `addAddress()` or `addCC()`. The second call with a duplicate address returns `false` and is ignored.

bartbutler:
Yeah, noticed that. So it's all set.

bartbutler:
Ahh, I forgot this PR was still open. Yeah, we are using this branch in production, so trying to address this: https://hackaday.com/2016/12/25/santa-knows-if-your-contact-form-uses-phpmailer-5-2-18/

Sorry for the noise. At some point I'll rebase onto the head.

kidplus2002:
I have been working on CRM systems for the purpose of integration.
With the increase of Hacking into Mailer systems and Thusly investigation on possible patch deployments as quickly as possible,  We are thankful.

Synchro:
I'm thinking that this approach is all a bit overcomplicated. As this is a corner case that only really applies at the SMTP level, I'd say the easiest way to do this is to subclass the SMTP class and override the `recipient()` method to just ignore certain addresses. It's then maintainable in your app's context, can survive library updates, and does not require inclusion in core. Along the lines of:

```php
class MySMTP extends PHPMailer\PHPMailer\SMTP
{
   public function recipient($address)
    {
        if (in_array($address, ['user@example.com', 'user2@example.net'])) {
            return true;
        }
        return parent::recipient($address);
    }
}
```

The Content-Location commit (9874728) is unrelated and should be moved into a separate PR.

@kidplus2002 I have no idea what you're talking about!

PR #618, opened 2016-01-31
Increase uniqueness of uniqueid

bertpoort:
The uniqid() function does not always return truly unique id's, especially on Windows (try on Windows: `for ($i = 0; $i < 10; $i++) echo md5(uniqid(time()))."\n";`). Using uniqid(..., TRUE) for more entropy would improve the uniqueness but it is relatively slow. time().mt_rand() is faster and sufficiently unique

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -2088,7 +2088,7 @@ class PHPMailer
     {
         $body = '';
         //Create unique IDs and preset boundaries
-        $this->uniqueid = md5(uniqid(time()));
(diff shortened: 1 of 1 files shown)


ooxi:
I agree that `uniqid` without `more_entropy = TRUE` is insufficient on windows (as [noted in the documentation](http://php.net/manual/en/function.uniqid.php#refsect1-function.uniqid-notes)) but wouldn't it be enough to use a static counter instead? Boundaries only have to be unique in the same mime message.

Synchro:
As @ooxi said, while this is clearly "more unique", it's not in a place where that actually matters.

PR #626, opened 2016-02-09
Updating README.md fixing some paths on the links

rfp:


Synchro:
Thanks, I swear I've fixed these about 3 times now! They keep coming back during branch merges that are proving very painful!

Issue #612, opened 2016-01-28
Create a diagnostic utility


Synchro:
So many questions on SO and issues raised in here are addressed by [the troubleshooting guide](https://github.com/PHPMailer/PHPMailer/wiki/Troubleshooting), but many seem unable to heed its advice. I propose creating a utility function that tries to diagnose a PHPMailer config issue by running through all the tests suggested in the guide in an automated fashion, producing diagnostic output that will direct the user to an appropriate fix.

ooxi:
Should the utility run automatically or by the user? More often than not the error message already includes steps to fix the failure (e.g. issue #525) but not read by the user. Thus I fear that an utility that should be run by the user is no more useful than linking to a troubleshooting guide :(

Synchro:
No, not automatically as it's too intrusive, but many have no idea how to even get to a command line on their server, let alone install/run telnet or do DNS lookups, check SSL certs. It may still lead to duplicate questions on SO for those incapable of reading an error message, but it would make them much simpler to deal with.

I think the most effective way would be to implement it as a method that you would call just before or after `send()` - perhaps adjust the examples to call it if a send fails, perhaps:

```php
if ($mail->send()) {
    echo 'yay!';
} else {
    echo $mail->diagnose();
}
```
We could also make the error message pointing at the guide a bit harder to ignore!

```
There was a problem sending your message.
 ____  _____    _    ____    _____ _   _ ___ ____  _
|  _ \| ____|  / \  |  _ \  |_   _| | | |_ _/ ___|| |
| |_) |  _|   / _ \ | | | |   | | | |_| || |\___ \| |
|  _ <| |___ / ___ \| |_| |   | | |  _  || | ___) |_|
|_| \_\_____/_/   \_\____/    |_| |_| |_|___|____/(_)
Before asking a question on SO or opening an issue on github, read this page:
https://github.com/PHPMailer/PHPMailer/wiki/Troubleshooting
```

ooxi:
That's a great idea :+1:

elminson:
Who is working on this ?

elminson:
I create the implementation of this but i don't now how to create a pull request. Who can help me?
e.g.   __  __                                               _                   _                                        _           
 |  \/  |   ___   ___   ___    __ _    __ _    ___    | |__     ___     __| |  _   _      ___   _ __ ___    _ __   | |_   _   _ 
[...]

PR #639, opened 2016-02-24
 Prevent invalid combinations of encryption algo and port

ozzi-:
Added a check that prevents invalid combinations of encryption algorithm and port by throwing a critical exception.


Synchro:
This isn't a reasonable thing to do as while it's unlikely, none of these combinations are set in stone and all are legitimate - it's not our job to prevent anyone using SSL on 587 and TLS on 465 if they want to. I'd suggest generating a warning if debug mode is enabled instead, and it would be a good thing to include in #612.

PR #640, opened 2016-02-24
Diagnose Function

ozzi-:
The diagnose function tries to help the user with some plausibility / network checks and points to the trouble shooting guide.

See commit https://github.com/ozzi-/PHPMailer/commit/27f10e119412fccd2cff248af9a5dc2aa25b760f

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -309,7 +309,7 @@ class PHPMailer
      * Default of 5 minutes (300sec) is from RFC2821 section 4.5.3.2
      * @var integer
      */
-    public $Timeout = 300;
+    public $Timeout = 30;
 
     /**
      * SMTP class debug output mode.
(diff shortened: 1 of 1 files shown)


SDKiller:
IMO, diagnostic features should be put in separate class/file, not to overbloat `class.phpmailer.php` - it is too large itself.

Synchro:
I'm inclined to agree with @SDKiller on this. We don't need a massive expansion of private methods either - it can be a simple single function. It would be nice to keep it separate as PHPMailer is too big already, however part of the advantage of the diagnose function as proposed is that it automatically gets access to everything in the correct context, and making it external would necessarily make it much bigger and more complex.

Generally, since this is a library that anyone should be free to subclass and do anything with (I can't really think of anything that should not be accessible to child classes), we don't make anything private, only protected.

Also, please rebase before making PRs.

PR #665, opened 2016-03-23
Bug fix for inclusion test bootstrap file on windows server

cyberkaa:
the test bootstrap file could not be resolved on a windows IIS server.


Synchro:
Rather than calling `realpath`, it might be simpler to remove the leading `./` from the paths.

cyberkaa:
works for me.

Synchro:
This has been fixed by a mixture of skipping the `../` prefix and using realpath. There were several other places that used such paths, so hopefully those are fixed too.

PR #669, opened 2016-03-25
A tip about DKIM_selector.

ivantcholakov:
https://www.emailarchitect.net/domainkeys/doc/html/selector.htm

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -389,6 +389,7 @@ class PHPMailer
 
     /**
      * DKIM selector.
+     * Indicates your DKIM public key location.
      * @var string
      */
     public $DKIM_selector = '';


Synchro:
This isn't true. A DKIM selector *is* a selector, it's not a location. It gets combined with the domain and the `_domainkey` prefix, and *those three things together* form a location, but the selector itself does not do that, so I think this comment is misleading and not helpful. If you're going to say something there, more useful would be a link to authoritative docs, such as http://www.dkim.org/specs/rfc4871-dkimbase.html#def-selectors

ivantcholakov:
http://www.dkim.org/specs/rfc4871-dkimbase.html#def-selectors

I've read this and my understanding remains the same. The public key has to be taken from somewhere and this option **indicates** how. Anyway.

Synchro:
Sure it has to come from somewhere, but the selector is not the location so calling it one is just wrong. I have no objection to adding explanations in phpdocs, but they do need to be both helpful and correct.

The link I gave says very similar things to the link you gave (as it should), but the difference is that it's the authoritative source on DKIM. The emailarchitect page has too much coverage of obsolete DomainKeys rather than DKIM, so it just adds to the confusion.

PR #675, opened 2016-04-04
Xoauth

sherryl4george:
Below are the changes:
1. `get_xoauth_token.php` - Changes  are made to use the League classes directly inorder to obtain refresh tokens
2. `src/OAuthProvider/Base.php` - Only essential code is being kept and the rest is deleted making it a cleaner implementation. Changes to constructor is made. Please do a rename of this file if necessary. Other files in this folder can be omitted
3. `examples/gmail_xoauth.phps` - Has been changed as we discussed it above.
`//Change the Class Name depending on the Provider Used (Line 65)
if (!isset($provider)) {
    $provider = new Google([
        'clientId' => $clientId,
        'clientSecret' => $clientSecret
    ]);
}` - This piece of code is used to fetch the provider. To make a cleaner implementation this can be moved to `Base` class. But in that case there should be some identifier for each provider (Google, MS, Yahoo) etc. so that the provider can be created inside the `Base` class based on the identifier. One way can be to assign a numeric value for each provider and pass it to `Base` class. Do we have a cleaner way to do this?

### DIFF
--- a/examples/gmail_xoauth.phps
+++ b/examples/gmail_xoauth.phps
@@ -81,10 +81,10 @@ $mail->setOAuth(new OAuthProvider\Base(
 
 //Set who the message is to be sent from
 //For gmail, this generally needs to be the same as the user you logged in as
-$mail->setFrom('sherryl4george@gmail.com', 'First Last');
+$mail->setFrom('someone@gmail.com', 'First Last');
 
 //Set who the message is to be sent to
-$mail->addAddress('gmathew016@gmail.com', 'John Doe');
+$mail->addAddress('someone@gmail.com', 'John Doe');
 
 //Set the subject line
 $mail->Subject = 'PHPMailer GMail SMTP test';
--- a/examples/gmail_xoauth.phps
+++ b/examples/gmail_xoauth.phps
@@ -1,8 +1,16 @@
 <?php
+
 /**
  * This example shows settings to use when sending via Google's Gmail servers.
  */
+
 namespace PHPMailer\PHPMailer;
+
+// Give alias to the League Provider Classes that may be used
+use League\OAuth2\Client\Provider\Google as Google;
+use Stevenmaguire\OAuth2\Client\Provider\Microsoft as Microsoft;
+use Hayageek\OAuth2\Client\Provider\Yahoo as Yahoo;
+
 
 //SMTP needs accurate times, and the PHP time zone MUST be set
 //This should be done in your php.ini, but this is how to do it if you don't have access to that
@@ -15,7 +23,7 @@ require '../vendor/autoload.php';
 require '../vendor/autoload.php';
 
 //Create a new PHPMailer instance
-$mail = new PHPMailerOAuth;
+$mail = new PHPMailer;
 
 //Tell PHPMailer to use SMTP
 $mail->isSMTP();
@@ -44,26 +52,39 @@ $mail->SMTPAuth = true;
 //Set AuthType
 $mail->AuthType = 'XOAUTH2';
 
-//User Email to use for SMTP authentication - Use the same Email used in Google Developer Console
-$mail->oauthUserEmail = "someone@gmail.com";
-
-//Obtained From Google Developer Console
-$mail->oauthClientId = "RANDOMCHARS-----duv1n2.apps.googleusercontent.com";
-
-//Obtained From Google Developer Console
-$mail->oauthClientSecret = "RANDOMCHARS-----lGyjPcRtvP";
+// Fill all necessary Authentication details here
+$email = 'someone@gmail.com';
+$clientId = 'RANDOMCHARS-----duv1n2.apps.googleusercontent.com';
+$clientSecret = 'RANDOMCHARS-----lGyjPcRtvP';
 
 //Obtained By running get_oauth_token.php after setting up APP in Google Developer Console.
 //Set Redirect URI in Developer Console as [https/http]://<yourdomain>/<folder>/get_oauth_token.php
 // eg: http://localhost/phpmail/get_oauth_token.php
-$mail->oauthRefreshToken = "RANDOMCHARS-----DWxgOvPT003r-yFUV49TQYag7_Aod7y0";
+$refreshToken = 'RANDOMCHARS-----DWxgOvPT003r-yFUV49TQYag7_Aod7y0';
+
+//Change the Class Name depending on the Provider Used
+if (!isset($provider)) {
+    $provider = new Google([
+        'clientId' => $clientId,
+        'clientSecret' => $clientSecret
+    ]);
+}
+
+$mail->setOAuth(new OAuthProvider\Base(
+        [
+    'provdier' => $provider,
+    'clientId' => $clientId,
+    'clientSecret' => $clientSecret,
+    'refreshToken' => $refreshToken,
+    'userName' => $email]
+));
 
 //Set who the message is to be sent from
 //For gmail, this generally needs to be the same as the user you logged in as
-$mail->setFrom('from@example.com', 'First Last');
+$mail->setFrom('sherryl4george@gmail.com', 'First Last');
 
 //Set who the message is to be sent to
-$mail->addAddress('whoto@example.com', 'John Doe');
+$mail->addAddress('gmathew016@gmail.com', 'John Doe');
 
 //Set the subject line
 $mail->Subject = 'PHPMailer GMail SMTP test';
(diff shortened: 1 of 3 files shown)


examples/gmail_xoauth.phps
```diff
@@ -44,26 +52,39 @@ $mail->SMTPAuth = true;
...
+if (!isset($provider)) {
+    $provider = new Google([
+        'clientId' => $clientId,
+        'clientSecret' => $clientSecret
+    ]);
+}
+
+$mail->setOAuth(new OAuthProvider\Base(
+        [
+    'provdier' => $provider,
```
Synchro: typo

Synchro:
Looking good, thanks.

Now we are not bundling OAuth provider classes, we can scrap the OAuthProvider folder. The OAuth class is no longer abstract or a base class, so we can call it PHPMailer\OAuth and move it up a level so it's simply:

```php
namespace PHPMailer\PHPMailer;
  		  
use League\OAuth2\Client\Grant\RefreshToken;		 
class OAuth {
...
```

I don't think using indexed IDs for providers is a good idea - by passing in the instance we have a clean injection of the external provider class, so all the League supported and third party providers should work without any modifications needed.

sherryl4george:
Please let me know if any further help required regarding OAuth

Synchro:
Thanks. I'll make the changes I mentioned and work through some examples to make sure it's happy.

Synchro:
I'm running into a silly PHP problem in the get token script - it seems impossible to create a class from a string name if the class name is aliased from another namespace, which is exactly what we're doing! [I posted it on SO](http://stackoverflow.com/questions/36432073/create-new-class-instance-from-string-name-in-an-aliased-namespace). Got any ideas?

Synchro:
It seems it is impossible to use dynamic class names if they are in aliased namespaces, so I just rewrote to use literals instead. The get_auth_token script should work now (you'll still have to jiggle composer.json packages), please give it a try.

Do you know if the Yahoo and MS providers need values in the options array like Google does?

sherryl4george:
Can we try using class_alias. I tried using this. But it gave me a error
that the class was found even when I was using the qualified class name.
May be I was doing something wrong.

On Tuesday, April 5, 2016, Marcus Bointon <notifications@github.com
<javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:

> I'm running into a silly PHP problem in the get token script - it seems
> impossible to create a class from a string name if the class name is
> aliased from another namespace, which is exactly what we're doing! I
> posted it on SO
> <http://stackoverflow.com/questions/36432073/create-new-class-instance-from-string-name-in-an-aliased-namespace>.
> Got any ideas?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/PHPMailer/PHPMailer/pull/675#issuecomment-205886632>
>


-- 
Regards,
Sherryl George
+91 7036858184

sherryl4george:
Yes. MS needs the values to be populated. I wasn't able to get to any
documentation regarding Yahoo OAuth2 for their mail sending. So I don't
think we can make Yahoo work for the time being. I can provide you with the
exact values that need to be passed for MS latest by tomorrow.

On Thursday, April 7, 2016, Marcus Bointon <notifications@github.com> wrote:

> It seems it is impossible to use dynamic class names if they are in
> aliased namespaces, so I just rewrote to use literals instead. The
> get_auth_token script should work now (you'll still have to jiggle
> composer.json packages), please give it a try.
>
> Do you know if the Yahoo and MS providers need values in the options array
> like Google does?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/PHPMailer/PHPMailer/pull/675#issuecomment-206505659>
>


-- 
Regards,
Sherryl George
+91 7036858184

sherryl4george:
I found the scopes that need to be included in options array for MS
`'wl.imap', 'wl.offline_access'`. I have tested this long back but I
suppose this to be working now also. Apologies that I can't test because I
am out currently.

On Thursday, April 7, 2016, Sherryl George <sherryl4george@gmail.com> wrote:

> Yes. MS needs the values to be populated. I wasn't able to get to any
> documentation regarding Yahoo OAuth2 for their mail sending. So I don't
> think we can make Yahoo work for the time being. I can provide you with the
> exact values that need to be passed for MS latest by tomorrow.
>
> On Thursday, April 7, 2016, Marcus Bointon <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:
>
>> It seems it is impossible to use dynamic class names if they are in
>> aliased namespaces, so I just rewrote to use literals instead. The
>> get_auth_token script should work now (you'll still have to jiggle
>> composer.json packages), please give it a try.
>>
>> Do you know if the Yahoo and MS providers need values in the options
>> array like Google does?
>>
>> —
>> You are receiving this because you authored the thread.
>> Reply to this email directly or view it on GitHub
>> <https://github.com/PHPMailer/PHPMailer/pull/675#issuecomment-206505659>
>>
>
>
> --
> Regards,
> Sherryl George
> +91 7036858184
>
>
>

-- 
Regards,
Sherryl George
+91 7036858184

Synchro:
There doesn't seem to be a way of getting a Google refresh token using a redirect URI only accessible to the developer (like localhost or a .local). Makes it pretty hard to test. Do you know any way around that?

sherryl4george:
You can provide the Redirect URI as the `http://localhost/folder/get_oauth_token.php` when you register the App. I have been testing all the time with a localhost only.

Synchro:
It seems inconsistent - I tried to add that URI to my credentials and it refused saying it must contain a valid TLD, but I tried it again and it worked!

sherryl4george:
I have never faced such a issue. I have never ever taken it to the live server actually. But I am not sure if they have something country specific or so.

Synchro:
I'm running into another problem with this. It's failing to verify the google cert after requesting permission. I have installed the latest [cacert.pem file from curl](http://curl.haxx.se/ca/cacert.pem) and set my php.ini to point at it as both `openssl.cafile` and `curl.cainfo` properties, but I'm still getting "cURL error 60: SSL certificate problem: unable to get local issuer certificate":

screen shot 2016-04-18 at 11 02 01

If I manually patch guzzle to point directly at the CA file it still fails; if I disable verification it succeeds, so it's definitely a verification problem.

Looking at the certificate chain for `accounts.google.com`, it uses the `Google Internet Authority G2` intermediate certificate (which is not in the CA file), and then the `GeoTrust Global CA` cert (which is in the CA file). I downloaded the google G2 cert, converted it to PEM and added it to the CA file, but that didn't help.

Curiously, the string `verify` does not appear in the league classes at all, suggesting that it relies on the default guzzle setting of `true`, and is then supposed to get the CA file from the PHP cafile or capath properties.

Does the callback URI need to be secure as well?

sherryl4george:
I am not very sure on how to mitigate the error. I am currently out of
town. Hope to help you as soon as I am back.

On Monday, April 18, 2016, Marcus Bointon <notifications@github.com> wrote:

> I'm running into another problem with this. It's failing to verify the
> google cert after requesting permission. I have installed the latest cacert.pem
> file from curl <http://curl.haxx.se/ca/cacert.pem> and set my php.ini to
> point at it as both openssl.cafile and curl.cainfo properties, but I'm
> still getting "cURL error 60: SSL certificate problem: unable to get local
> issuer certificate":
>
> [image: screen shot 2016-04-18 at 11 02 01]
> <https://cloud.githubusercontent.com/assets/81561/14598951/0b92bcc2-0555-11e6-9e9c-242befddf4d1.png>
>
> If I manually patch guzzle to point directly at the CA file it still
> fails; if I disable verification it succeeds, so it's definitely a
> verification problem.
>
> Looking at the certificate chain for accounts.google.com, it uses the Google
> Internet Authority G2 intermediate certificate (which is not in the CA
> file), and then the GeoTrust Global CA cert (which is in the CA file). I
> downloaded the google G2 cert, converted it to PEM and added it to the CA
> file, but that didn't help.
>
> Curiously, the string verify does not appear in the league classes at
> all, suggesting that it relies on the default guzzle setting of true, and
> is then supposed to get the CA file from the PHP cafile or capath
> properties.
>
> Does the callback URI need to be secure as well?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/PHPMailer/PHPMailer/pull/675#issuecomment-211293634>
>


-- 
Regards,
Sherryl George
+91 7036858184

Synchro:
Aha - I think I found the cause - my curl install (that PHP was using) was using libressl instead of openssl and it seems it doesn't like that.

Synchro:
BTW I've merged the xoauth branch into the new 5.5 branch

PR #682, opened 2016-04-11
Use $exceptions default value if not set on construct.

eusonlito:


Synchro:
Why do this? What do you think it helps?

eusonlito:
If you have a default value for exceptions into variables definition list, why to force to use one value from construct?

Now you have two `false` definition to $exceptions, one into construct and one into variable definition. If you change one, you need to remember to change the other or values will be inconsistent.

Synchro:
Fair enough, DRY I guess. Thanks.

PR #683, opened 2016-04-16
Add unit test that shows how address parser returns its result

PunchyRascal:
Also I have 2 questions, if anyone could help:
1) What setup is needed after cloning the repo to have all unit tests pass?
2) Why are all the assert names ("Failed to recognise address list etc.") more like descriptions for when the test fails? Is it not obvious that a test failed, when it does? Why put it in the description? The description could be a good documentation.

### DIFF
--- a/test/phpmailerTest.php
+++ b/test/phpmailerTest.php
@@ -1417,6 +1417,19 @@ EOT;
             ),
             'Failed to recognise address list (IMAP parser)'
         );
+        $this->assertEquals(
+            array(
+                array("name" => 'Joe User', 'address' => 'joe@example.com'),
+                array("name" => 'Jill User', 'address' => 'jill@example.net'),
+                array("name" => '', 'address' => 'frank@example.com'),
+            ),
+            $this->Mail->parseAddresses(
+                'Joe User <joe@example.com>,'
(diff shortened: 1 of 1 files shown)


Synchro:
Assertions don't have names. The description is only shown when an assertion fails, and you don't need any extra info when a test passes. There may be multiple assertions in a test, so there isn't anywhere else to put this message. What were you thinking of?

PunchyRascal:
I consider tests the ultimate documentation (apart from being tests of course) and for that purpose the assertion comment/name is a useful lead as to what the test is doing, how the actual code behaves. 

An example: instead of "Failed to recognise address list", it might be better to use "Recognizes list of addresses". Because like you said, you only see this message, apart from reading the test itself, only when it fails and in that case, there seems to be no need to say that in the message.

Synchro:
Well often the message contains info about why the assertion failed, and I'd find that confusing and irrelevant at any other time. During failures is pretty much the only time I see the messages, and it needs to describe what went wrong and why, not what should have gone right (which is what the assertion type itself already conveys). Tests as documentation is very much a secondary function - it needs to be effective, descriptive and concise as a test first. This is also in line with how other packages use assertion messages, including PHPUnit itself.

PR #693, opened 2016-04-25
Added 'extension_missing' translation for ka locale

akalongman:
Added 'extension_missing' translation for ka locale


Synchro:
Thanks

PR #700, opened 2016-05-02
Set BCC in header in case SMTP is used.

marekfri:
BCC was not set in createHeader() function of PHPMailer class if SMTP was used.

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -1957,7 +1957,7 @@ class PHPMailer
 
         // sendmail and mail() extract Bcc from the header before sending
         if ((
-                $this->Mailer == 'sendmail' or $this->Mailer == 'qmail' or $this->Mailer == 'mail'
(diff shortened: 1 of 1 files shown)


Synchro:
Nope. I don't think you understand how BCC works. When sending via SMTP, BCC addresses do not appear in the message headers, only in the SMTP envelope.

risototh:
Maybe we do not understand, how the Bcc is working with the SMTP, **but in the practice, all clients shows the Bcc, when it's set.** An the recipient is happy, when it sees, why he got that particular email.
This was our solution (modified PHPMailer class), but now i'm forced to do a child class that will fix this, as we want to use the base via composer.

Synchro:
If you want to see the recipient addresses in the message, use CC, not BCC. That's the difference between them. It is intended to work exactly as you describe; when you are sent a message via BCC, the headers will not include your own address, other than in `Received` header lines added by your own mail server.

PR #702, opened 2016-05-02
Configuration option to only allow specific domains

TheJosh:
Very useful when in a test environment to protect against accidental mailing to unintended recipients.

We desired this feature for our CMS, as an added protection in case someone imports a list which they shouldn't have.

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -342,6 +342,15 @@ class PHPMailer
     public $Debugoutput = 'echo';
 
     /**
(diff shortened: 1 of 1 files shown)


Synchro:
I really don't think this is worth having as a core feature - it's exactly the sort of thing to put into a subclass. What's more, checking it at send time seems wrong - I would override `validateAddress` so that these bad domains are not allowed in in the first place. Why wouldn't you filter these address before they even get into your system?

PR #764, opened 2016-07-06
Fix ISO-639-1 code for Danish

arnested:
The ISO-639-1 code for Danish is `da` (`dk` is the ISO 3166-1 alpha-2 code for Denmark).


ooxi:
While I agree with the rename I'm troubled whether we should add a special handling for `dk` in [PHPMailer::setLanguage](https://github.com/PHPMailer/PHPMailer/blob/master/class.phpmailer.php#L1682) to retain backward compatibility.

On the one hand, renaming `dk` to `da` can cause as much as error messages in the wrong language, on the other hand somebody might depend on danish error messages :/

Synchro:
I agree about patching setLanguage too, but we could just leave it as is and fix it in 6.0.

arnested:
OK, I add backward compatibility handling in `setLanguage`.

Synchro:
Thanks

PR #774, opened 2016-07-19
Add a save() method

flahol:
Sometimes it is useful to save a mail to a file.
A save method could be added :

```
public function save($dir, $filename='')
{
  $filename = $filename ? $dir . $filename . '.eml' : $dir . date('Y-m-d_H:i') . '_' . str_replace('/', '-', $this->Subject) . '.eml';
  try {
    $file = fopen($filename, 'w+');
    fwrite($file, $this->GetSentMIMEMessage());
    fclose($file);
    return true;
  } catch (Exception $e) {
    $this->SetError($e->getMessage());
    return false;
  }
}
```


Synchro:
No. PHPMailer's job is email, not writing files. You wrote a lot of code to do the same thing as `file_put_contents()`. You also made a total mess of this PR?

PR #779, opened 2016-07-22
Update German and Spanish localization

pedritoelcabra:
Added the 'extension' translation for German and Spanish


Synchro:
Thanks!

Issue #765, opened 2016-07-07
Default encoding charset for PHPMailer 6


aduh95:
Greetings,

The default encoding charset in PHPMailer is ISO-8859-1 (also called Latin-1), and maybe we should reconsider this choice, regarding the fact that UTF-8 is now the web standard. As PHPMailer 6 is still in development, this change should be made before its official released.

As UTF-8 and Latin-1 use the same character table for the ASCII characters, it would be harmless for full English projects to use UTF-8 instead, and an  improvement for projects using other languages (I'm currently working on a project using French language and UTF-8, that's why I care ;) ).

If this change is accepted, we have to add something like this in the `UPGRADING.md` file:

``` php
PHPMailer now uses UFT-8 as default encoding charset.
If you want to keep using Latin-1, you should add something like this to your code:
<?php
    $phpmailer->CharSet = 'iso-8859-1';
?>
```

Synchro:
While I think I agree, it's not necessarily an obvious choice. For example, you're using French, for which UTF-8 is entirely unnecessary and is less efficient than ISO-8859-1. In English especially even 8-bit encodings are usually unnecessary, and PHPMailer contains code to automatically fall back to US-ASCII and a 7bit transfer encoding, even if you originally specified UTF-8.

Synchro:
It's actually an RFC contravention to specify an encoding that exceeds what's needed for the message!

aduh95:
I get that UTF-8 is not the absolute best choice, but I think it's a better choice as a default encoding: ISO-8859-1 is limited to 189 characters, and many applications offer more possibilities to their users (from emoji to different alphabets). Especially if PHPMailer optimises automatically for full ASCII emails as you said, I don't see why ISO-8859-1 should remain as default encoding charset.

Synchro:
There's another issue - the `mbstring` extension (which PHPMailer requires for UTF-8 support) is not installed by default. It's also not installed by default in any of the default PHP packages on any Ubuntu LTS release of PHP, including PHP 7 on Xenial, nor on Centos and other Red Hat variants, so it must be explicitly installed. That it often is installed isn't really the point - it's not the job of a library to dictate such requirements. We already have some code that checks for the presence of mbstring, but since we don't rely on it by default it needs checking more thoroughly before we can change the default.

In other words, there's a lot more to consider than simply changing the default value. Patches welcome.

PR #795, opened 2016-08-08
Switch to UTF-8 default encoding

aduh95:
This PR is about issue [#765](https://github.com/PHPMailer/PHPMailer/issues/765).

 * Add [`symfony/polyfill-mbstring`](https://github.com/symfony/polyfill-mbstring) dependency to provide a partial, native PHP implementation for the Mbstring extension.
 * Set `UTF-8` as default charset instead of `iso-8859-1`.

Please tell me if I should add or revert something, I'll be glad to help!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,9 @@
     "suggest": {
         "league/oauth2-google": "Needed for Google XOAUTH2 authentication",
         "hayageek/oauth2-yahoo": "Needed for Yahoo XOAUTH2 authentication",
-        "stevenmaguire/oauth2-microsoft": "Needed for Microsoft XOAUTH2 authentication"
+        "stevenmaguire/oauth2-microsoft": "Needed for Microsoft XOAUTH2 authentication",
+        "ext-mbstring": "Needed to send email in multibyte encoding charset",
+        "symfony/polyfill-mbstring": "To support UTF-8 if the Mbstring PHP extension is not enabled (^1.2)"
     },
     "autoload": {
         "psr-4": {


Synchro:
'partial' is a bit of an understatement! It provides support for only UTF-8 and ASCII - if you use ISO-8859-x, or any windows codepage, Russian, Chinese etc it will break, which is considerably worse than doing without it at all! Although that polyfill avoids the need for the mbstring extension, it doesn't provide a pure PHP implementation at all - it substitutes a dependency on the iconv extension instead, which is very old and very slow, and is moving the problem rather than fixing it, though that might be permissible since we're already talking about a less-than-optimal installation anyway.

You will also have noticed that PHPMailer has no dependencies for its primary functions; A large proportion of the user base installs by hand (despite me telling them not to!), so this is important. I'm also really not keen on hard-linking to symfony components - at best this option should go into 'suggests' as an optional install (since most people won't need it) or even just as documentation.

All this also needs test cases (which that polyfill lacks also), and that applies to #765 as well.

aduh95:
Actually, Symphony's polyfills have PHPUnit test implemented, you can find the Mbstring one [here](https://github.com/symfony/polyfill/blob/master/tests/Mbstring/MbstringTest.php).

I'd suggest that we highly encourage users to activate the `Mbstring` extension, especially if they're not  using Composer or using a different encoding. For those who don't want to or cannot, the polyfill implementation is good enough as far as I'm concerned, since `iconv` is enabled by default in all supported versions of PHP.

About linking to Symphony components, as the library is under [MIT license](https://github.com/symfony/polyfill-mbstring/blob/master/LICENSE), we *could* "copy/paste" the code into the PHPMailer repo, but we don't want to do that, do we? ☺ You have certainly already noticed that this mbstring library is performance-wise harmless when the real PHP extension is activated.

About test cases, I'm kind of a newbie on that, what do you want me to do? To replace in the phpmailerTest.php ISO-8859-1 by UTF-8 and vice versa? I feel like it must something a bit more complicated than that...

Synchro:
We do 'encourage' enabling mbstring by the somewhat passive-aggressive approach of not supporting UTF-8 if they don't! I think putting it into suggest and improving docs on it would be enough - I don't want to start pulling packages for everyone just for a few edge cases - we already get complaints about having too many files, though I doubt anyone using symfony is bothered by that!

The things I'd want to see covered by tests would be what happens when substitution with the polyfill is done. That does mean that the Travis-CI test env would need to be reconfigured to disable the mbstring extension.

SDKiller:
IMO, it's easier for users, espesially on shared hosting tariffs, to contact support to enable mbstring, than to use composer to install additional dependencies.
More to say - nowadays, anno 2016, it is difficult to imagine hosting with mbstring disabled.
Suggested polyfill is an overbloat.

aduh95:
> nowadays, anno 2016, it is difficult to imagine hosting with mbstring disabled.

As @Synchro [said](https://github.com/PHPMailer/PHPMailer/issues/765#issuecomment-231421629):

 > That it often is installed isn't really the point - it's not the job of a library to dictate such requirements.

And if many users won't use composer, I disagree that it's a difficult way to install a library or : there are plenty of tutorials all over the web; even more it is a good practice and PHPMailer does encourage it. Whatever the hosting policies you have to deal with, it doesn't matter if you have downloaded the PHP files manually or through Composer, it's PHP files anyway.

If we don't want PHPMailer to have any dependency, maybe `ISO-8859-1` should remain PHPMailer's default encoding charset?

Synchro:
If you're planning on implementing any kind of web app that needs UTF-8, mbstring is really a requirement - If it was missing you'd run into problems very quickly anyway, so again, I don't think it's really up to PHPMailer to dictate terms. I don't see any harm in adding the compat package to composer's suggest list as a last-ditch option, and mentioning it in docs. I think it's extremely unlikely anyone would implement an app without UTF8, but still require it in email.

I know some libraries are quite opinionated - for example trying to force PHP 7 upgrades - but I think the job of a library is to do just what it needs to do, with as little in the way of requirements as is reasonable. Trying to force updates from a library is backseat driving.

So, on balance, maybe we should stick with ISO-8859-1 as the default - if you want UTF-8 it's only a 1-line change that's well documented already. If you want to have UTF-8 by default, just set it in a subclass and you won't have to worry about it again - that's what I've done for the last few years.

Synchro:
Could you rebase this please?

aduh95:
Could I? The [Git documentation](https://git-scm.com/book/en/v2/Git-Branching-Rebasing#The-Perils-of-Rebasing) says:

 > Do not rebase commits that exist outside your repository.
 >
 > If you follow that guideline, you’ll be fine. If you don’t, people will hate you, and you’ll be scorned by friends and family.

When I try, it seems to affect only my fork repo, and not this pull request. If you want to avoid to merge, you may close my pull request and make a commit on your own. In the end, the only file that changed is `composer.json` in my commit [c5ab0d47f74dc8fdbe51c129ab76c245c3c66920](https://github.com/aduh95/PHPMailer/blob/c5ab0d47f74dc8fdbe51c129ab76c245c3c66920/composer.json).

Synchro:
Don't worry about that warning - it worrried me when I started out! It's only really a problem for repos that have large numbers of users, where forcing pushes can cause issues, but it's really not a problem when there are no sub-forks of your own repo. What you need to do is rebase and squash all your commits into a single commit then push (with `-f` to force it) the result into your repo, and it will show up here. It's a common scenario and helps keep the commit history clean, especially where you have later commits that undo earlier ones within the same PR.

composer.json
```diff
@@ -29,7 +29,9 @@
     "suggest": {
         "league/oauth2-google": "Needed for Google XOAUTH2 authentication",
         "hayageek/oauth2-yahoo": "Needed for Yahoo XOAUTH2 authentication",
-        "stevenmaguire/oauth2-microsoft": "Needed for Microsoft XOAUTH2 authentication"
+        "stevenmaguire/oauth2-microsoft": "Needed for Microsoft XOAUTH2 authentication",
+        "ext-mbstring": "Needed to send email in multibyte encoding charset",
+        "symfony/polyfill-mbstring": "To support UFT-8 if the Mbstring PHP extension is not enabled (^1.2)"
```
svenluijten: `UFT-8` should be `UTF-8`, right? :smiley:

aduh95: Indeed! Thanks for pointing out ;)

aduh95:
BTW @Synchro, should I update the lock file? I'm not sure it's up to me.

Synchro:
No, just do your own changes, the lock file will get updated elsewhere.

Synchro:
For the record, no we are not switching to UTF-8 as a default but sticking with ISO-8859-1 as the default Charset.

PR #796, opened 2016-08-08
Feature/refactor

lexville:


Synchro:
This whole PR is a mess. What are you doing this for? Turning a single instance of inline code into a function that's never re-used is wasteful and pointless refactoring, and declaring all those methods as public is a really bad idea - they should be private or protected as they should not be making such radical and useless changes to the public API.

The changes in the language files are just wrong and fail to understand how the language system works - look at the language tests.

You've also commented out nearly all of the test suite!

If you want to start somewhere, I suggest you look at the PHPMailer 6.0 branch.

class.phpmailer.php
```diff
@@ -3670,6 +3670,17 @@ public function DKIM_QP($txt)
         return $line;
     }
 
+    public fucntion privateKey($privKeyStr)
```
Eternal-Infinity: Misspelled "function" as "fucntion"

Issue #804, opened 2016-08-17
Validity of unquoted space in mail address


bor0:
We have:

```
boro@bor0:~/dev$ php test.php
/Users/boro/dev/test.php:4:
bool(true)
boro@bor0:~/dev$ cat test.php
<?php
require('./vendor/phpmailer/phpmailer/PHPMailerAutoload.php');
var_dump(PHPMailer::validateAddress("test @test.com"));
```

According to the [tests written](https://github.com/PHPMailer/PHPMailer/blob/master/test/phpmailerTest.php#L445), it seems like this was done on purpose.

Using OS X Mail, it complains it's not a valid e-mail address.
Also, per the [RFC](https://tools.ietf.org/html/rfc5322#section-3.4.1):

> Comments and folding white space SHOULD NOT be used around the "@" in the addr-spec.

In my understanding, `SHOULD NOT` is used because it is allowed in some cases (e.g. when surrounded by quotes), which is not the case for `test @test.com`.

Additionally, there are inconsistencies when using different patterns:

```
<?php
require('./vendor/phpmailer/phpmailer/PHPMailerAutoload.php');

$patterns = array(
    'pcre8',
    'pcre',
    'html5',
    'noregex',
    'php',
);

foreach ($patterns as $pattern) {
    echo "$pattern => " . (PHPMailer::validateAddress("test @test.com", $pattern) ? 'ok' : 'nok') . "\n";
}
```

Returns:

```
pcre8 => ok
pcre => nok
html5 => nok
noregex => ok
php => nok
```

So in my opinion, pcre8 should also return `nok` for that e-mail address (it's okay that noregex returns `ok` since that one is a really simple one).

Synchro:
I agree - if you'd like to tweak the regex to not allow this specific case, be my guest, but not like your #805 PR just did!

bor0:
As far as I investigated, the issue seems to be around this on the [page](http://web.archive.org/web/20151017073126/http://squiloople.com/2009/12/20/email-address-validation):
```
// Comments and folding white spaces

'/^((?>(?>(?>((?>(?>(?>\x0D\x0A)?[\t ])+|(?>[\t ]*\x0D\x0A)?[\t ]+)?)(\((?>(?2)(?>[\x01-\x08\x0B\x0C\x0E-\'*-\[\]-\x7F]|\\\[\x00-\x7F]|(?3)))*(?2)\)))+(?2))|(?2))?)$/iD'
```

[...]

PR #805, opened 2016-08-17
Update the regex to match PHP's internal one

bor0:
Fixes https://github.com/PHPMailer/PHPMailer/issues/804. Note that this change is a breaking change.

In case this PR gets approved, need to update tests accordingly.


bor0:
See [this link](https://github.com/php/php-src/blob/PHP-5.6.25/ext/filter/logical_filters.c#L529) for PHP's internal regex.

Synchro:
In that same source code you'll see that it mentions Michael Rushton (of Squiloople) - he also wrote the expression used in PHPMailer (see the comments in the code), which is used specifically because the stock PHP one is much less permissive of valid addresses - as you will see from all the Travis test failures you generated.

Aside from that, this PR is pointless because PHPMailer already has support for PHP's built-in filter, and you can tell it to use it by simply changing the default validator in your code:

    PHPMailer::$validator = 'php';

And we don't do breaking changes in minor versions.

bor0:
@Synchro agreed with your points, but https://github.com/PHPMailer/PHPMailer/issues/804 is still valid imom

PR #814, opened 2016-08-26
6.0(Refactor and Improve documentation)

andela-iadeniyi:
This `PR` address the inconsistency in the code documentation and also a few code duplication. For example, making all method have a defined format for the `docblock`(`@link`>`@access`, `@param`>`see`>`@use`>`throw`>`@return`>`static`).

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -740,19 +740,24 @@ class PHPMailer
         $this->Mailer = 'mail';
     }
 
+    private function transportAgent($type, $path)
+    {
+        $ini_sendmail_path = ini_get('sendmail_path');
+
+        if (!stristr($ini_sendmail_path, $type)) {
+            $this->Sendmail = $path;
+        } else {
+            $this->Sendmail = $ini_sendmail_path;
+        }
+    }
+
     /**
      * Send messages using $Sendmail.
      * @return void
      */
     public function isSendmail()
     {
-        $ini_sendmail_path = ini_get('sendmail_path');
-
-        if (!stristr($ini_sendmail_path, 'sendmail')) {
-            $this->Sendmail = '/usr/sbin/sendmail';
-        } else {
-            $this->Sendmail = $ini_sendmail_path;
-        }
+        $this->transportAgent('sendmail', '/usr/sbin/sendmail');
         $this->Mailer = 'sendmail';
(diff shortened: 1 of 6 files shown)


src/PHPMailer.php
```diff
@@ -818,6 +817,21 @@ public function addReplyTo($address, $name = '')
...
 
+    private function enqueue($address, $data)
+    {
+        if (!array_key_exists($address, $data)) {
+            $data[$address] = $params;
+            return true;
+        }
+    }
+
+    private function throwException($error_message)
```
Synchro: This is a bad idea. Exceptions should be thrown from where they occur, otherwise you can't tell from the error alone where the error occurred and you have to generate a stack trace to find out.

Synchro:
Sorry there's just too much wrong with this to accept it all.

* Turning every `if` into a method that's only ever called once doesn't help readability.
* Wrapping exception throwing in a method will just make my job harder as I'll get endless people reporting there's a problem in the exception throwing function.
* There's no point in adding `@access public` and `@return void` since those are the defaults, and the order isn't specified in PSR-2. `@access` is mainly pointless anyway since phpdoc can get that info from the code anyway - it is a vestige of old PHP versions which did not have access control at all, so it was maintained manually.
* We don't use private methods in a library like this - our policy is everything should be overridable.

There are some legitimate fixes in here, but they are all mixed in with the ones that are counterproductive. If you want to do anything structural, I recommend you target the unreleased 6.0 branch where such changes might be more acceptable.

andela-iadeniyi:
@Synchro: Okay, thanks for the update. If I understand what you're suggesting:
- Exception should not be abstracted from where it is thrown
- Avoid the use of `private methods` because it goes against PHPMailer's policy
- `@access public` and `@return void` should not be defined because they are default.
- Turning `if` into a method doesn't help readability.

To fix these issues:

- Throw exception where its suppose to be thrown
- Use `protected methods` instead of `private methods`
- Remove `@access public` and `@return void`, though the order isn't specified in PSR-2 but PHPMailer can have a default order to enhance consistency.
- Avoid abstracting `if` statement
- Target unreleased 6.0 branch, but only one branch is available for apart from 6.0 which is `xoauth` which is dated later than `6.0`

PR #829, opened 2016-09-16
Rename SECURITY to SECURITY.md

glensc:
this way it is rendered properly (github, vim, etc that use file extension to detect file content)


Synchro:
Thanks

PR #842, opened 2016-10-02
[EasyPeasyICS.php] Use random_bytes(), if possible

oittaa:


SDKiller:
IMO its overkill for this purpose, depends on platform, availability of `/dev/urandom` and throw unnesessary exception if it fail.

Synchro:
I agree that the default system isn't secure, but it's not that critical either. The default works ok, and it's only used if you don't supply your own value, which is what you would do if you have higher requirements. I don't want to change things that are likely to create support issues right now! I would accept dropping the use of md5 altogether, as that's causing more harm than good.

EasyPeasyICS is being dropped from PHPMailer in 6.0 anyway - it will be managed as a separate project, and this change will certainly be welcome there.

oittaa:
>  depends on platform, availability of /dev/urandom and throw unnesessary exception if it fail.

Please read the documentation before commenting something like that about security related functions.

```
 The sources of randomness used for this function are as follows:

    On Windows, » CryptGenRandom() will always be used.
    On Linux, the » getrandom(2) syscall will be used if available.
    On other platforms, /dev/urandom will be used.
    If none of the aforementioned sources are available, then an Exception will be thrown.

```

> EasyPeasyICS is being dropped from PHPMailer in 6.0 anyway

OK, then this patch isn't  necessary.

PR #843, opened 2016-10-03
Add .gitattributes

vinkla:
This will allow smaller download size in production.

https://www.reddit.com/r/PHP/comments/2jzp6k/i_dont_need_your_tests_in_my_production

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,13 @@
+* text=auto
+
+/docs export-ignore
+/examples export-ignore
+/test export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.scrutinizer.yml export-ignore
+/.travis.yml export-ignore
+/travis.phpunit.xml.dist export-ignore
+/SECURITY.md export-ignore
+/changelog.md export-ignore
+/README.md export-ignore


Synchro:
Good idea, thanks

.gitattributes
```diff
@@ -0,0 +1,13 @@
...
+/examples export-ignore
+/test export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.scrutinizer.yml export-ignore
+/.travis.yml export-ignore
+/travis.phpunit.xml.dist export-ignore
+/SECURITY.md export-ignore
+/changelog.md export-ignore
+/README.md export-ignore
```
Synchro: In pentest audits, presence of files like that is considered a risk due to information leakage. No need for them in production.

vinkla: Yeah, I agree with @Synchro. These files just take up space in production. The only file important to keep is LICENSE.

vinkla:
Thanks for merging!

Synchro:
There's an unexpected problem with this - it also affects the zip file download - see #856

vinkla:
@Synchro yeah, this is how `export-ignore` works. Though, I don't think it is much of a problem. We want to be able to exclude files in production.

vinkla:
The PHP League has this as one of their 10 commandments https://thephpleague.com/

Synchro:
Yes, I can understand that for production use with composer - but who would be using zip downloads in production?

vinkla:
Then you need to revert this pull request. Though, then we can't exclude those files in production.

glensc:
i'm using github export tarball to make distro rpm package (https://github.com/pld-linux/php-phpmailer); and it's affected as well. now the rpm package is without changelog and documentation; 

some solution i see is to upload full tarball (tar.xz perhaps even) to github release so i could use that as source for distro packaging.

PR #855, opened 2016-10-16
Fixing typos in German translation ('Deppenleerzeichen')

e7o-de:
Fixed some smaller but aweful errors in German translation related to correct usage of a dash.


Synchro:
Due to conflicts with #866 I had to merge this manually in af59766. Can you please check that it's OK?

PR #857, opened 2016-10-17
Fix: function call is not case-sensitive

corentinheadoo:
- function call is not case-sensitive
- PHPDoc doesn't contain @throws tag
- Parameter 'false' type is not compatible with declaration


Synchro:
Thanks - though note that the everything in the extras folder is being dropped in 6.0.

PR #862, opened 2016-10-20
Closes #861

### LINKED ISSUE #861: PHP 7.1 compatibility - Escaped command exceeds the allowed length
In PHP 7.1 we get

> mail(): Escaped command exceeds the allowed length of 2097152 bytes

This happens at 
https://github.com/PHPMailer/PHPMailer/blob/master/class.phpmailer.php#L692

You get this error if
`$params === NULL `

PR coming ;)

neo22s:


Synchro:
Thanks. I cleaned up the code a little - the outcome of those first two conditions was the same, so I combined them.

neo22s:
You are right!! Didnt notice. Thanksbgot fast reply ;)

PR #863, opened 2016-10-20
Added support for php 7.1 in .travis.yml

twish:
Now possible to test against php 7.1.


Synchro:
Thanks, good to know that works!

PR #867, opened 2016-10-26
Update README.md

vedipen:
Closed PHP tag in the simple example provided, for its fast and easy use.


Synchro:
Nope. It's like that (as are all the code files in PHPMailer) because it's required by [PSR-2 section 2.2](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md), for very good reasons.

PR #868, opened 2016-10-28
switch uniqid for random_bytes

udf2457:
As per the PHP docs, using uniqid is really not sensible when you need to generate unique values.

As per the PHP docs, switch to random_bytes() instead.

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -2138,7 +2138,7 @@ class PHPMailer
     {
         $body = '';
         //Create unique IDs and preset boundaries
-        $this->uniqueid = md5(uniqid(time()));
+        $this->uniqueid = md5(random_bytes(16));
         $this->boundary[1] = 'b1_' . $this->uniqueid;
         $this->boundary[2] = 'b2_' . $this->uniqueid;
         $this->boundary[3] = 'b3_' . $this->uniqueid;


SDKiller:
@udf2457 
I wonder, have you seen "as per PHP docs" in which php version it was introduced?

Synchro:
Yes, the current mechanism in't cryptographically secure, but on the other hand, it really doesn't need to be. As @SDKiller says, it also breaks on a bunch of supported versions, hence the build failures. If you're going to switch to `random_bytes`, you should also stop using `md5`, but that's academic in this case. Perhaps consider a different PR for the 6.0 branch, which can at least guarantee PHP 5.5.

udf2457:
(a) Its nothing to do with cryptography
(b) What on earth are you doing *still* supporting obsoltete, unsupported and insecure versions of PHP.

Synchro:
a) You didn't understand what I said
b) Because we have ~100 million users that need it? You're looking at the legacy branch which supports back to PHP 5.0; this lib has been around since 2001, and it's not a library's job to dictate PHP versions. Like any library, it should deliver what it does at as low a version of PHP as strictly necessary; Anything else is back-seat driving. Opinions are for frameworks, not components.

PR #877, opened 2016-11-09
ability to create custom message IDs and boundaries

1stborn:
Due to a problem with Yandex FBL, single possible way to determine sender is a boundary, but there is no way to set it from outside, cuz even i'll do this ( by extending), it will be rewritten by createBody anyway. So I want to propose simple way to customize message boundary and message-id as bonus

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -2128,6 +2128,14 @@ class PHPMailer
     }
 
     /**
+     * Create unique ID
+     * @return string
+     */
+    protected function generateId() {
+        return md5(uniqid(time()));
+    }
+
+    /**
      * Assemble the message body.
      * Returns an empty string on failure.
      * @access public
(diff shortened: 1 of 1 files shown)


Synchro:
I'll accept this anyway, as it does make it easier to override. What is it that the FBL does? Can't you use VERP? If it breaks return paths, it can't be much use for email!

1stborn:
FeedBack Lists, Yandex hide all possible way to detect which our user is complain for spam from us and automatically unsubscribe it from mailing lists - they call it "security". But they left some headers of out message untouched `content-type` one of them

Synchro:
Ah, ok, tricky! Yahoo and Hotmail leave everything open.

Synchro:
How can they automatically unsubscribe an address you've sent to without you knowing about it?

1stborn:
I ask myself the same question :)

Synchro:
Thanks

PR #879, opened 2016-11-10
Allow DKIM private key to be provided as a string

lamansky:
For my project I needed to be able set the DKIM private key dynamically from the database, so setting a file path wasn't feasible. This allows the DKIM private key to be set as either a file path or a string.


Synchro:
Thanks

PR #883, opened 2016-11-18
add getLastMessageId()

JesusTheHun:
Hello,

In some ecosystems, tracking message's id can be important. So I add this method to be able to fetch it from last reply.
Unfortunately the message id is lost in the return string, which changes from software to software, to a empirical approach is used. It just works.

### DIFF
--- a/class.smtp.php
+++ b/class.smtp.php
@@ -1211,4 +1211,29 @@ class SMTP
             self::DEBUG_CONNECTION
         );
     }
+
+	public function getLastMessageId() {
+		$reply = $this->getLastReply();
+
+		switch (true) {
+
+			// Exim
+			case preg_match("/[0-9]{3} OK id=(.*)/", $reply, $id):
+				return $id[1];
+			break;
+
+			// Sendmail
+			case preg_match("/[0-9]{3} 2.0.0 (.*) Message/", $reply, $id):
+				return $id[1];
+			break;
+
+			// Postfix
+			case preg_match("/[0-9]{3} 2.0.0 Ok: queued as (.*)/", $reply, $id):
+				return $id[1];
+			break;
+			default:
+				return false;
+			break;
+		}
+    }
 }


Synchro:
This is wrong, or at least confusing. A message ID is determined by the client, as described in [RFC5322 section 3.6.4](https://tools.ietf.org/html/rfc5322#section-3.6.4), and PHPMailer gives you access to that in `PHPMailer::getLastMessageID()`. Servers may generate their own internal transaction IDs, but that's nothing to do with the message ID. The codes you are parsing out are not message IDs but transaction IDs - for example a single message that experiences deferrals before successful delivery may have several such transaction codes for a single message ID. Similarly, a message that passes through several mail servers (as most do - usually at least local + remote) will have a chain of such identifiers, but only a single message ID, which is generally the one you should use for tracking, because the transaction ID you get at point of submission will not propagate to the point of delivery, and so isn't very useful. Fundamentally, a message ID is a property of a message (i.e. in RFC822-land), not an SMTP transaction (RFC821).

Maybe rename it as getTransactionID or something? I'm curious as to why you need this.

JesusTheHun:
I may have missnamed things, the point is to track the smtp id

JesusTheHun:
Shall I add a proxy method to this method in PHPMailer ? Or better keep it in the smtp class to avoid confusion ?

Synchro:
Definitely leave it in the SMTP class. I'm thinking that it would be best to grab the transaction id when a DATA command is successful (i.e. when a message delivery succeeds), rather than leaving it up to the user to call it at the right time, and make `getTransactionID()` simply return the value. That approach should make it more reliable - or do you want to get the ID earlier so you can log unsuccessful delivery transaction IDs?

JesusTheHun:
Well, as for `getLastReply()` method, it will simply returns `false` if you call it before a call. I don't see why we should restrict its access only upon success. Plus as you mentioned, if may be useful in case of error too.

Synchro:
Fair enough. Do you happen to have some transcripts that demonstrate these? It would be useful to make it work for things like gmail, which seems to be the most common use case for PHPMailer.

JesusTheHun:
After some mess up (I deleted my fork, oups), I re-opened a PR.
What do you mean by "transcript" ? to demonstrate what ? ^^

Synchro:
BTW - you say its for legal purposes, but it provides no such assurance. All it shows is that you submitted to your local mail server successfully - you can't tell if the destination server is down, blocking you or deferring you. Submitting directly to destination servers is generally a bad idea - it's unreliable because PHPMailer is not an MTA, and if the target server's response doesn't match any of the patterns, it returns the same value as if it had failed to deliver altogether.

I mean transcripts of SMTP conversations that show the transaction IDs, like what you get with SMTPDebug = 2.

JesusTheHun:
We do know it's not perfect. But at least we can prove we tried, even if something that happened made it fail.
Well I'm a fan of exceptions, but since PHPMailer does not use them, I'm afraid to throw some.

JesusTheHun:
And for transcripts :

Exim: 250 OK id=1REMFv-0001wd-BG
Sendmail : 250 2.0.0 p9DERpqf001411 Message accepted for delivery
Postfix : 250 2.0.0 Ok: queued as 3VmHNZ28NLz2Nr7

Synchro:
BTW PHPMailer can throw exceptions if you ask it to - just pass `true` to the constructor, look in the examples. Error reporting is slightly finer-grained with exceptions.

JesusTheHun:
The SMTP has no constructor :/ I'll send null & false. Better than two false...

PR #884, opened 2016-11-18
Add getLastTransactionID()

JesusTheHun:
Retrieve the last transaction id from the smtp server response

### DIFF
--- a/class.smtp.php
+++ b/class.smtp.php
@@ -150,6 +150,17 @@ class SMTP
      */
     public $Timelimit = 300;
 
+	/**
+	 * @var array patterns to extract smtp transaction id from smtp reply
+	 * Only first capture group will be use, use non-capturing group to deal with it
+	 * Extend this class to override this property to fulfil your needs.
+	 */
+	protected $smtp_transaction_id_patterns = array(
+		'exim' => '/[0-9]{3} OK id=(.*)/',
+		'sendmail' => '/[0-9]{3} 2.0.0 (.*) Message/',
+		'postfix' => '/[0-9]{3} 2.0.0 Ok: queued as (.*)/'
+	);
+
     /**
      * The socket for the server connection.
      * @var resource
@@ -1210,5 +1221,29 @@ class SMTP
             $notice . ' Error number ' . $errno . '. "Error notice: ' . $errmsg,
             self::DEBUG_CONNECTION
         );
+    }
+
+	/**
+	 * Will return the ID of the last smtp transaction based on a list of patterns provided
+	 * in SMTP::$smtp_transaction_id_patterns.
+	 * If no reply has been received yet, it will return null.
+	 * If no pattern has been matched, it will return false.
+	 * @return bool|null|string
+	 */
+	public function getLastTransactionID()
+	{
+		$reply = $this->getLastReply();
+
+		if (empty($reply)) {
+			return null;
+		}
+
+		foreach($this->smtp_transaction_id_patterns as $smtp_transaction_id_pattern) {
+			if(preg_match($smtp_transaction_id_pattern, $reply, $matches)) {
+				return $matches[1];
+			}
+		}
+
+		return false;
     }
 }


Synchro:
I can't think of a good reason to make the patterns public - I think they would be better as `protected static`, or even `const` - it's unlikely you would ever want two instances with different patterns, and I'm keen to keep the number of public properties down - there are already too many. This would mean that you need to subclass to add patterns, but I don't see that as a big problem. What do you think?

JesusTheHun:
public : if the user wants to add its own pattern, he can.
const : unless you turn PHPMailer to be compatible php7 only, it's compromised.

I chose public because you used to set a lot of properties as public, avoiding setter and getter, so I tried to copy the project's code style. I can add `prependPattern` and `appendPattern`, but it's easier to manipulate if you have direct access : maybe one server has a very close pattern to another, so one must be placed before the other. For exemple :

Foo's message : 250 OK TR 123456ABCDEF ( with `TR ([A-Z0-9]+)` )
Bar's message  250 OK TR ID123456ABCDEF (with `TR ID([A-Z0-9]+)` ) 

Foo's pattern is provided by PHPMailer and will catch any alphanum character after TR
Bar's pattern will be catch by Foo's pattern, unless bar's pattern is added before foo's pattern or Foo's pattern is modified.

SDKiller:
Maybe it should be not property at all - but implemented as protected method to retrieve patterns list.

User may subclass to override method - for example to retrieve patterns from database, etc.

Synchro:
I agree with @SDKiller here - I don't really want to add to the already excessive list of public properties; Overriding is a more appropriate approach to this kind of thing. I'm not quite sure why everyone thinks every option should be in the base class - it's what overriding is for, and it's not exactly hard. It's what I do myself when I use PHPMailer. In this case it should really be a protected static method, but we can't do that sensibly without late static binding which didn't appear until PHP 5.3, so next best thing is  a protected method. I'd prefer to do this properly in 6.0, and not even try to do this in 5.2 since it's a new feature. @SDKiller - I could use some help with 6.0 if you have time - it's really pretty complete, really just needs a last sanity check before release.

SDKiller:
> I could use some help with 6.0 if you have time - it's really pretty complete, really just needs a last sanity check before release.

Yes, this is just what I was going to do this week.
I need to upgrade several projects and testing 6.0 on them was planned already.

JesusTheHun:
The part where I think we diverge is I think it could be very nice if PHPMailer provides a list of known patterns. The `SMTP` class has 8 public properties, 2 related to debug. However I do understand your point. Moreover if the list of known pattern grows, significantly, we cannot store dozen of patterns in the class. So, what do you think of a protected property with getter & setter, and a static JSON resource ?

@Synchro : what is the minimum PHP version of the 6.0 ?

Synchro:
PHPMailer 6.0 will require PHP 5.5. If you check out [the 6.0 branch](https://github.com/PHPMailer/PHPMailer/tree/6.0), take a look at the readme and the examples to see how things have changed.

Using a method override makes it easy to add or remove patterns, and you don't need setters & getters, and you can have as many as you like. Moreover, this is not a thing that anyone will need to do very often - you are the first person to ask for it in 15 years - and so an override is the appropriate way to go about it.

JesusTheHun:
I respectfully disagree. You're the maintainer, I comply to your expectation for the project. I'm updating the PR.

Synchro:
How does using an override prevent you from doing any of the things you asked?

JesusTheHun:
It does not. I simply am against overriding third-party classes on one hand, and on the other hand I think a class should include the knowledge required to make a feature work (a list of patterns here). On my perspective, extending third-party classes introduces legacy complexity in your project.

Synchro:
But it will contain a list of patterns - the ones you've already proposed?

I'd contend that writing wrappers for third-party classes is a vital part of abstracting your own interfaces away from third-party implementations, exactly in the OO design philosophy of "code to an interface, not an implementation". In my own code I use a two-layer wrapper that allows me to override bits of PHPMailer, but also to swap out the whole thing for completely different implementations, such as Zend_Mail. I do the same thing with many other classes doing entirely different things.

JesusTheHun:
You want it to include the list I proposed ? I understood you wanted an empty property.

I do agree with that vision, in theory. From my experience it's idealistic, and introduce complexity without real necessity. Look at ORMs "I can switch database engine", it's wrong and even if true, nobody actually does it. Moreover, unless you start a project, you have to deal with all the legacy stack and code and practices.
But I concede for a mailer, it can be done without much complexity.

Synchro:
No, of course populate it, it needs to be useful!

I'd guess that anyone using PHPMailer frequently will probably be wrapping it anyway, just to do simple things like preset the Host, Username etc properties, or to map them into their own interfaces. All the frameworks I know of using PHPMailer already do this - Drupal, Yii, CodeIgniter, Kohana, Wordpress etc. This is just business as usual. If they don't explicitly subclass, they'll usually wrap it in some mail function that effectively does the same thing.

I agree with you on ORM though!

The place where I've been saved by wrappers several times is in user agent string analyser classes - several popular ones have collapsed and disappeared, but because I wrapped them, I've not had to make substantial app changes, just update my wrappers.

Synchro:
Could you squash and force-push your commits please?

JesusTheHun:
Seems ok to me.

PR #927, opened 2016-12-28
Use long array syntax as PHP >= 5.0.0 is supported

reedy:
https://github.com/PHPMailer/PHPMailer/blob/master/composer.json#L23 says you require PHP >= 5.0.0, but you've some files with short array syntax, which is new in PHP 5.4, so breaks support for PHP < 5.4...

Causes problems for inclusions when people are linting with PHP 5.3

### DIFF
--- a/extras/htmlfilter.php
+++ b/extras/htmlfilter.php
@@ -772,7 +772,7 @@ function tln_fixstyle($body, $pos, $trans_image_path, $block_external_images)
     tln_defang($contentTemp);
     tln_unspace($contentTemp);
 
-    $match   = Array('/\/\*.*\*\//',
+    $match   = array('/\/\*.*\*\//',
                     '/expression/i',
                     '/behaviou*r/i',
                     '/binding/i',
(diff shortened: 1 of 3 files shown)


Synchro:
Because all this is a BC break, it is all being done much more comprehensively in [the 6.0 branch](https://github.com/PHPMailer/PHPMailer/tree/6.0).

extras/htmlfilter.php
```diff
@@ -772,15 +772,15 @@ function tln_fixstyle($body, $pos, $trans_image_path, $block_external_images)
...
+    $match   = array('/\/\*.*\*\//',
                     '/expression/i',
                     '/behaviou*r/i',
                     '/binding/i',
                     '/include-source/i',
                     '/javascript/i',
                     '/script/i',
                     '/position/i');
-    $replace = Array('','idiocy', 'idiocy', 'idiocy', 'idiocy', 'idiocy', 'idiocy', '');
+    $replace = array('','idiocy', 'idiocy', 'idiocy', 'idiocy', 'idiocy', 'idiocy', '');
```
glensc: this probably worth merging in, as separate MR and with different purpose (`Array` is like from JavaScript, `array` is the php capitalization)

Synchro:
I've patched the array names - that file is chock-full of PSR violations, and it's being removed for 6, so it's not worth expending any more effort on!

glensc:
agreed: it's copy in from other project and in modern world that should be separate composer package.

PR #929, opened 2016-12-28
Fix for CVE-2016-10045 and CVE-2016-10033

### LINKED ISSUE #924: CVE-2016-10045: Critical RCE vulnerability introduced in 5.2.18
A critical RCE vulnerability has been introduced by the patch for CVE-2016-10033.  Please contact me at paul@namepros.com for more details.  PGP: https://keybase.io/zenexer

Zenexer:
Fixes #924.  Needs to be tested to ensure no significant functionality is broken.  Will avoid sending -f argument to sendmail when it's deemed unsafe to do so.  Note that while this is a big step forward, more work needs to be done to increase security on Windows.  `sendmail` and `mail` transport options should not be used on Windows for the time being.

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -1444,7 +1444,7 @@ class PHPMailer
         //This sets the SMTP envelope sender which gets turned into a return-path header by the receiver
         if (!empty($this->Sender) and $this->validateAddress($this->Sender)) {
             // CVE-2016-10033, CVE-2016-10045: Don't pass -f if characters will be escaped.
-            if (escapeshellcmd($this->Sender) === $this->Sender) {
+            if (escapeshellcmd($this->Sender) === $this->Sender && in_array(escapeshellarg($this->Sender), array("'$this->Sender'", "\"$this->Sender\""))) {
                 $params = sprintf('-f%s', escapeshellarg($this->Sender));
             }
         }
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -1364,7 +1364,8 @@ class PHPMailer
      */
     protected function sendmailSend($header, $body)
     {
-        if (!empty($this->Sender)) {
+        // CVE-2016-10033, CVE-2016-10045: Don't pass -f if characters will be escaped by escapeshellcmd when popen is called due to safe mode.
+        if (!empty($this->Sender) || ini_get('safe_mode')) {
             if ($this->Mailer == 'qmail') {
                 $sendmail = sprintf('%s -f%s', escapeshellcmd($this->Sendmail), escapeshellarg($this->Sender));
             } else {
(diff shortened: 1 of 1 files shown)


Zenexer:
Note: if safe mode is enabled, -f will never be used for the `sendmail` transport, but may be used for the `mail` transport.  The checks in `mail` can be extracted to their own function and used in `sendmail` if that behavior needs to be preserved, but safe mode is inherently insecure and incompatible anyway.

To-do: Use fork + execve instead of the shell where available (e.g., pcntl_fork + pcntl_exec).

Synchro:
Safe mode was deprecated in PHP 5.3 and removed in 5.4 anyway. All mention of it will be going away in future versions.

Zenexer:
@Synchro I figured, which is why I didn't go out of my way to support it.  The CVEs are patched, but some minor functionality will be lost in safe mode.

Synchro:
Thanks for this - I think this will do as a general approach for now. If you've got any further comments let me know ASAP and I'll tag a new release so we can get this out.

class.phpmailer.php
```diff
@@ -1364,7 +1364,8 @@ public function postSend()
      */
     protected function sendmailSend($header, $body)
     {
-        if (!empty($this->Sender)) {
+        // CVE-2016-10033, CVE-2016-10045: Don't pass -f if characters will be escaped by escapeshellcmd when popen is called due to safe mode.
+        if (!empty($this->Sender) || ini_get('safe_mode')) {
```
torunar: Is it me, or this condition will result in always passing the `Sender` (even when empty) when the `safe_mode` is enabled?

class.phpmailer.php
```diff
@@ -1442,7 +1443,10 @@ protected function mailSend($header, $body)
         $params = null;
         //This sets the SMTP envelope sender which gets turned into a return-path header by the receiver
         if (!empty($this->Sender) and $this->validateAddress($this->Sender)) {
-            $params = sprintf('-f%s', escapeshellarg($this->Sender));
+            // CVE-2016-10033, CVE-2016-10045: Don't pass -f if characters will be escaped.
+            if (escapeshellcmd($this->Sender) === $this->Sender && in_array(escapeshellarg($this->Sender), array("'$this->Sender'", "\"$this->Sender\""))) {
```
torunar: Why don't put this check into the `setFrom()` method?

Synchro: Because `Sender` could be reset later.

Zenexer:
I wrote this while tired; I'm going to need to go through it more thoroughly in the morning. There's one particular case I'm not confident about.

class.phpmailer.php
Zenexer: Yes, you're right. I originally rearranged the if statement a bit and didn't properly refactor.

Synchro: I'll fix that.

Synchro: I think it should be `&& !ini_get('safe_mode')`. I checked that if `safe_mode` is not defined, e.g. because it's in PHP > 5.4, it behaves correctly.

Zenexer: Yes, it should be.  I have a pull request on the way; I'm just checking something else while I'm at it.

Synchro: Ok, I'll wait for that, thanks.

PR #930, opened 2016-12-28
CVE-2016-10033, CVE-2016-10045: More patches, stricter anti-escaping

Zenexer:
Needs manual testing, particularly on Windows.

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -1364,20 +1364,24 @@ class PHPMailer
      */
     protected function sendmailSend($header, $body)
     {
-        // CVE-2016-10033, CVE-2016-10045: Don't pass -f if characters will be escaped by escapeshellcmd when popen is called due to safe mode.
-        if (!empty($this->Sender) || ini_get('safe_mode')) {
+        // CVE-2016-10033, CVE-2016-10045: Don't pass -f if characters will be escaped.
+        if (!empty($this->Sender) and self::isShellSafe($this->Sender)) {
             if ($this->Mailer == 'qmail') {
-                $sendmail = sprintf('%s -f%s', escapeshellcmd($this->Sendmail), escapeshellarg($this->Sender));
+                $sendmailFmt = '%s -f%s';
             } else {
-                $sendmail = sprintf('%s -oi -f%s -t', escapeshellcmd($this->Sendmail), escapeshellarg($this->Sender));
+                $sendmailFmt = '%s -oi -f%s -t';
             }
         } else {
             if ($this->Mailer == 'qmail') {
-                $sendmail = sprintf('%s', escapeshellcmd($this->Sendmail));
+                $sendmailFmt = '%s';
             } else {
-                $sendmail = sprintf('%s -oi -t', escapeshellcmd($this->Sendmail));
+                $sendmailFmt = '%s -oi -t';
             }
         }
+
+        // TODO: If possible, this should be changed to escapeshellarg.  Needs thorough testing.
+        $sendmail = sprintf($sendmailFmt, escapeshellcmd($this->Sendmail), $this->Sender);
+
         if ($this->SingleTo) {
             foreach ($this->SingleToArray as $toAddr) {
                 if (!@$mail = popen($sendmail, 'w')) {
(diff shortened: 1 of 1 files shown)


class.phpmailer.php
```diff
@@ -1364,20 +1364,24 @@ public function postSend()
...
-                $sendmail = sprintf('%s', escapeshellcmd($this->Sendmail));
+                $sendmailFmt = '%s';
             } else {
-                $sendmail = sprintf('%s -oi -t', escapeshellcmd($this->Sendmail));
+                $sendmailFmt = '%s -oi -t';
             }
         }
+
+        // TODO: If possible, this should be changed to escapeshellarg.  Needs thorough testing.
+        $sendmail = sprintf($sendmailFmt, escapeshellcmd($this->Sendmail), $this->Sender);
```
Zenexer: Ideally this should be escapeshellarg rather than escapeshellcmd, but that would break any applications that rely on the existing behavior to pass their own parameters to sendmail via `sendmail_path`.

class.phpmailer.php
```diff
@@ -1444,8 +1480,8 @@ protected function mailSend($header, $body)
         //This sets the SMTP envelope sender which gets turned into a return-path header by the receiver
         if (!empty($this->Sender) and $this->validateAddress($this->Sender)) {
             // CVE-2016-10033, CVE-2016-10045: Don't pass -f if characters will be escaped.
-            if (escapeshellcmd($this->Sender) === $this->Sender && in_array(escapeshellarg($this->Sender), array("'$this->Sender'", "\"$this->Sender\""))) {
-                $params = sprintf('-f%s', escapeshellarg($this->Sender));
```
Zenexer: Forgot to remove escapeshellarg here, although it shouldn't have had much effect.

class.phpmailer.php
```diff
@@ -1364,20 +1364,24 @@ public function postSend()
      */
     protected function sendmailSend($header, $body)
     {
-        // CVE-2016-10033, CVE-2016-10045: Don't pass -f if characters will be escaped by escapeshellcmd when popen is called due to safe mode.
-        if (!empty($this->Sender) || ini_get('safe_mode')) {
+        // CVE-2016-10033, CVE-2016-10045: Don't pass -f if characters will be escaped.
```
Zenexer: I've completely removed the safe_mode test because it's still insecure even without safe mode.  escapeshellcmd and escapeshellarg miss far too many cases, especially on Windows.

class.phpmailer.php
Synchro: Isn't `escapeshellcmd` appropriate here since it is a command, not an arg?

Zenexer: Given its actual implementation, I'm not really sure what the PHP developers were thinking when they created it.  I doubt there's ever a legitimate reason to use it.  Each argument--including the command/binary itself--should be escaped as a separate argument.  At least, that would be the best solution if escapeshellarg were actually complete.

As it stands, escapeshellcmd's effect in this statement is unintuitive and probably not what anyone would expect.  For example, it will allow the path to contain spaces, but only if the path is quoted (e.g., `set_ini('sendmail_path', '"/some dir/sendmail"')`)  It's also going to have weird effects on certain characters in the pathname, including the backslashes commonly found on Windows.

class.phpmailer.php
Synchro: Do you have further commits to clean up these small things before I merge?

class.phpmailer.php
Zenexer: Here's an adaptation of PHP's escapeshellcmd and escapeshellarg code that will help you test different scenarios: https://gist.github.com/Zenexer/c123604d57914970ac297413751c3f21  Note that argument parsing in Windows is nothing like it is in bash.  For example, commas, equals signs, and semicolons are argument separators, not just whitespace, and escaping highly dependent on context.

class.phpmailer.php
Zenexer: Probably not.  I was trying to fix the CVEs without giving away too much about the underlying issues, hence the incompleteness, but I've pretty much given up on that with this latest pull request.  However, it's 7 AM here and I haven't slept, so it's certainly possible I'm missing something.

Synchro: I think the cat was out of the bag some years ago: https://bugs.php.net/search.php?cmd=display&search_for=escapeshellarg, but it doesn't look like much has been done.

Zenexer: That's a bit concerning.  I'll see if I can come up with some sort of hackish long-term solution when I wake up.  This should do for now, though.

class.phpmailer.php
Synchro: I don't know what to do with that, but it looks like it would form the basis of a good PR for PHP itself.

Synchro:
Merged, via a slightly roundabout route, thanks

PR #937, opened 2016-12-28
composer: indicate ctype ext being used

glensc:
833c35fe3 added ctype extension requirement

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,7 @@
         }
     ],
     "require": {
+        "ext-ctype": "*",
         "php": ">=5.0.0"
     },
     "require-dev": {


glensc:
also, [ctype_alnum](http://php.net/ctype_alnum) is locale aware, are you sure that is intended?

glensc:
sadly i don't see testcases being submitted. it would be easy to rewrite it with regexp knowing it behaves the same as before with given testcases. it's hard figuring out what the method is supposed to do now :)

glensc:
it's supposed to handle some kind of escaping for windows that php native escapeshellarg/escapeshellcmd don't do? i've came across this library at some point in the past: https://github.com/johnstevenson/winbox-args

Synchro:
This function tests to see if the passed string is likely to need any escaping at all, firstly by using the built-in functions, but if they pass, by allowing only a very restricted set of characters, which amount to not much more than `[a-zA-Z0-9@\._-]` - note that this does not include `=+'"\` because they are used in paths, separators and escaping in various different shells. The reasoning for this is that [the PHP shell escape functions are not reliable](https://gist.github.com/Zenexer/40d02da5e07f151adeaeeaa11af9ab36), so we need to simply avoid the need to do any escaping at all, if possible.

Synchro:
And tests would indeed be welcome too.

glensc:
#938 - no actual tests filled

PR #938, opened 2016-12-28
add shellsafe method tests

glensc:
made `isShellSafe` not-static so it could be accessed from unit tests.

made it this way to be compatible with php 5.0, as otherwise could had used:
- `static::` (requires php 5.3)
- ReflectionMethod::setAccessible (requires php 5.3.2)


glensc:
i'm thinking maybe just test each byte and put it either to bad or good list? that doesn't test combination of bytes, but should help to cover so the underlying method could be refactored (https://github.com/PHPMailer/PHPMailer/pull/937#issuecomment-269540929)

glensc:
can be merged likely if tests pass as every single byte is tested, sequence tests can be added later

Synchro:
Isn't it still the case that you can call static methods dynamically without errors? The other way around doesn't work, obviously.

glensc:
yes, that would work, but it's ugly/inconsistent, and may broke in the future php engine? is there reason why it needs to be static in main class? if you insist removing that change, i'll do it (local test still passes)

Synchro:
Well it's static because that's the right 'shape' for it, IMHO - it does not depend on any dynamic properties or $this, and it's common to all instances. The distinction is pretty academic so I'm not too bothered. It can go back to being static in 6.0 when we don't have language issues in the way of the tests. I'm about to head for bed, so I'll merge this in the morning. Thanks for this BTW.

glensc:
ok, i'll leave as is, static methods/properties are hard to test, phpunit doesn't support TestDoubles for static methods for example.

sleep time here as well :)

glensc:
@Synchro slept well? :)

Zenexer:
This will be insecure in many Eastern countries because it's not locale-aware.  You have no idea what locale the shell is using, and if it's not ASCII-compatible, your "@" character could end up meaning something completely different, in theory.  Of course, there's always the possibility that the locale of the shell and the locale of PHP are incompatible, but that seems less likely than an alternative 7-bit charset.  Testing for specific bytes actually ends up being dangerous.  (For example, Shift JIS encodes backspace and tilde differently than ASCII.)  Now, you could always hope the user converts the PHP file to their locale of choice when they copy it to their server, but that's putting a bit too much trust in the user for my liking, and it's definitely not going to work if your hardcode the individual bytes.  Nevertheless, that's what we have to rely on for a few symbols in the original version.

Zenexer:
If you happen to have PCRE functions compiled in but not ctype for some crazy reason (looking at you, Gentoo), you can use a regex instead, but it will be slower:

``` php
preg_match('/\A[\pl\pn._@-]+\Z/ui', $string)
```

Synchro:
...and this is why I asked for your opinion! In that case I assume that in order to test this correctly, we should generate the character range from the current locale too, rather than using a fixed array? Perhaps test against all known locales? Or is that going too far?

Zenexer:
I'm working on a "one-size-fits-all" solution [here](https://gist.github.com/Zenexer/40d02da5e07f151adeaeeaa11af9ab36#file-isshellsafe-php).  It's completely untested, but that may be closer to what you're aiming for.  I'm not really sure how you'd generate the arrays in the current PR dynamically without relying on mbstring, iconv, or intl.

Synchro:
PHPMailer requires mbstring for UTF-8 support, which many use, though it's not the default. I don't think there's ever been much testing of running without it.

glensc:
@Zenexer this PR is about adding testing support. not actually changing (yet) the actual `isShellSafe` implementation suggested here: https://github.com/PHPMailer/PHPMailer/pull/937#issuecomment-269540929

so as i see it: first get unit tests in place, then go change implementation

Zenexer:
@glensc Then yes, mbstring would be the way to go for locale testing.  I don't know a whole lot about EUC, but it sounds like EUCKR would make for a good test, since it will likely break the existing function.

Synchro:
I'm just revisiting this. Zend mail had exactly the same vulnerability (also found by Dawid Golunski / @legalhackers-lab), but [their fix was quite different](https://framework.zend.com/security/advisory/ZF2016-04) - and also much simpler.  They simply reject any sender address that contains the sequence `\"`. I wonder if we would be better off using that solution instead - though I'm not 100% sure that it's entirely safe (compared to the more aggressive approach in PHPMailer), so your input would be appreciated!

Zenexer:
It's definitely not safe in all environments, and may be quite easy to
evade; I haven't tested it.  I don't recall needing that sequence to
exploit the vuln, but it's been a while since I did the testing.

Ultimately, there's no solution that's both reasonable and perfect if the
application is allowing end users to affect that parameter.  Applications
most likely to be affected regardless of what precautions we take are those
in countries that primarily write in non-Latin scripts and have
historically employed non-ASCII-compatible charsets, including stateful
charsets.  This is a fundamental flaw in the design of PHP's mail
function.  I've reported it, but no interest was shown in fixing it or
providing an alternative.

If pcntl_exec is available, it should be safe to use, as it doesn't rely on
the shell's parsing.  We should prioritize that method of invoking sendmail
when it's available.

glensc:
@Synchro you may close or merge or pick fragments from this PR. i'm not going to finish this whatever was planned. sorry :)

PR #940, opened 2016-12-29
composer: phpunit 4.8

Ramoonus:
switch from 4.7 to 4.8 (latest stable) of phpunit

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     },
     "require-dev": {
         "phpdocumentor/phpdocumentor": "*",
-        "phpunit/phpunit": "4.7.*"
+        "phpunit/phpunit": "4.8.*"
     },
     "suggest": {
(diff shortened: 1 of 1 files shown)


glensc:
PR should describe purpose, not just contents of commit message, which can be read out anyway.

Synchro:
We can't do this. We need to run tests on PHP 5.3 and 5.4, which are not supported in PHPUnit 4.8.

Ramoonus:
@Synchro according to the release notes not https://github.com/sebastianbergmann/phpunit/wiki/Release-Announcement-for-PHPUnit-4.8.0

Synchro:
Hm. You're right. I remember this came up before, but I can't remember what the problem was. 4.7 has the same minimum PHP version as 4.8. I know we're stuck with 4.8 in PHPMailer 6 because we need to run on PHP 5.5. Lets see if we can persuade this PR to run its tests...

PR #962, opened 2017-01-14
completed the example code

SscSPs:
it didn't had the closing tag

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,7 @@ if(!$mail->send()) {
 } else {
     echo 'Message has been sent';
 }
+?>
 ```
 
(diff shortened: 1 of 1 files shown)


SDKiller:
@SscSPs 

http://php.net/manual/en/language.basic-syntax.phptags.php

> If a file is pure PHP code, it is preferable to omit the PHP closing tag at the end of the file. This prevents accidental whitespace or new lines being added after the PHP closing tag, which may cause unwanted effects because PHP will start output buffering when there is no intention from the programmer to send any output at that point in the script.

SscSPs:
Okay! learned something new! Thanks @SDKiller

PR #983, opened 2017-02-01
change __autoload 

Flayor:
Change __autoload to spl_autoload_register for php 7.2 (depreced)
https://wiki.php.net/rfc/deprecations_php_7_2


Synchro:
No. 5.2 is already EOL, the autoloader is entirely gone in PHPMailer 6.0, and removing `__autoload` breaks older PHP versions, which is what the 5.2 branch is used in.

Synchro:
Also, that's a deprecation in 7.2, which isn't even released yet, and __autoload won't be removed until PHP 8.0.

PR #984, opened 2017-02-01
Convert use of each() to foreach() for performance

drbyte:
Convert use of each() to foreach() for performance
and also to avoid triggering deprecation alerts in PHP 7.2

This PR is against master, which is 5.2.22 presently. 
I notice this file isn't currently in 6.0 ... but if it "does" get added, then this change should be included there.


Synchro:
I'll allow this in, but htmlfilter (which is full of bad code anyway!) is being dropped completely from 6.0 - it's just out of scope for PHPMailer really.

drbyte:
Thanks!

PR #996, opened 2017-02-13
addAddress : Accept addresses like 'John Doe <john.doe@mail.com>'

lapsus63:
With the class PHPMailer, I could call method addAddress with two ways:
- $mail -> addAddress('john.doe@mail.com', 'John Doe');
- $mail -> addAddress('john.doe@mail.com');

I've updated the addAddress method to accept the following option :
- $mail -> addAddress('John Doe <john.doe@mail.com>');

Which is RFC-2822 compliant.

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -816,6 +816,11 @@ class PHPMailer
      */
     public function addAddress($address, $name = '')
     {
+	if (strpos($address, '<')) {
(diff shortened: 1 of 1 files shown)


Synchro:
Please see the functionality provided by the `parseAddresses()` method which your code largely duplicates. You're also missing quite a lot of error checking.

PR #1020, opened 2017-04-05
Mails

federicorosario:
Hi, my mails on wordpress doesnt works. 

Y have enable de smtp for mail server. My server is windows.  

When i tried to send a contact from de form i get an error.

what can i do?

thank you!


Synchro:
Ask questions like this on Stack Overflow. There's nothing anyone can do without seeing your error, and it's most likely a wordpress problem. Creating random pull requests is not helpful.

PR #1023, opened 2017-04-09
Date header should be more dynamic

rsdifu:
When a message is sent, unless the MessageDate is explicitly set, the Date header should always be set to the current rfcDate() value.

This is necessary for long-running processes that create a single PHPMailer instance, so all emails sent during the lifetime of the process don't have the same Date header value of the first email sent from the process.

This is important because Gmail in particular, uses the Date header value to show how long ago an email was sent when viewing the email.  (The list view in Gmail, however, uses the Received header value to show when an email was sent.)

These differing times will lead to confusion for users, often resulting in them trying to do several things like updating their timezone in Google's settings, or on their computers, which would result in no resolution.

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -2024,10 +2024,7 @@ class PHPMailer
     {
         $result = '';
 
-        if ($this->MessageDate == '') {
-            $this->MessageDate = self::rfcDate();
-        }
-        $result .= $this->headerLine('Date', $this->MessageDate);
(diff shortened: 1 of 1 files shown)


Synchro:
This commit makes no sense. Your code does exactly the same thing as the existing code, just rephrased as a less readable ternary expression.

Synchro:
I take that back - sorry, I see what you mean.

Synchro:
Thanks. I will merge this into master and 6.0 later.

Synchro:
Merged into master and 6.0

PR #1065, opened 2017-06-02
add parameter SMTPSecure to example smtp.phps

yantze:
No parameter SMTPSecure cause 
```
SMTP connect() failed
```

### DIFF
--- a/examples/smtp.phps
+++ b/examples/smtp.phps
@@ -17,9 +17,12 @@ $mail->isSMTP();
 // 0 = off (for production use)
 // 1 = client messages
 // 2 = client and server messages
+// 3 = Enable verbose debug output
 $mail->SMTPDebug = 2;
 //Ask for HTML-friendly debug output
 $mail->Debugoutput = 'html';
+//Set smtp secure [ssl, tls]
+$mail->SMTPSecure = 'ssl';
 //Set the hostname of the mail server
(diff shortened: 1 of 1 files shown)


Synchro:
Sorry, no. Setting ssl with port 25 is pretty much guaranteed to fail, and undermines this being a minimal SMTP example.

yantze:
Smtp also can use port 576 and 587

I use the port 576, if not add SMTPSecure to ssl, it will run fail, so do i push another example ?

yantze:
Also can indicate port 576 use ssl and port 587 use tls

Synchro:
No. There are plenty of other examples that demonstrate `SMTPSecure`, and PHPMailer will work with explicit TLS automatically anyway. SSL on port 465 has been deprecated since 1998, so we do not want to encourage it. There is no need to change this basic example to be compatible with your non-standard setup.

yantze:
Deprecated but developers must use it, because many(at least my country) mail providers only provide service ssl with 465

This is not changed by developers, and this problem took me a lot of time. may be others also.

So i suggest that add another example such as smtp_legacy_netease_sina.phps file.

It may be more convenient for developers.

Synchro:
It's not worth adding a whole new example that's only 5 bytes different, especially when it's already covered extensively in documentation.

PR #1073, opened 2017-06-13
Minor correction of attribute description

manzoorwanijk:
Corrected the description of Parameters to `$action_function` which otherwise gave me a lot of trouble. The corrected params are passed in `doCallback()` as arrays but here it was mentioned as strings.

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -440,9 +440,9 @@ class PHPMailer
      *
      * Parameters:
      *   boolean $result        result of the send action
-     *   string  $to            email address of the recipient
-     *   string  $cc            cc email addresses
-     *   string  $bcc           bcc email addresses
+     *   array   $to            email addresses of the recipients
+     *   array   $cc            cc email addresses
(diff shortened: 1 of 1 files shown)


Synchro:
Thanks

Synchro:
Incidentally I'd be interested in hearing how you're using these callbacks - it's something I'm considering removing from PHPMailer, or at least revising heavily as the current implementation is pretty bad! A simple function injection would be much better that what is here now, but I don't know that it's actually very useful at all anyway.

manzoorwanijk:
I'm using it as a trigger to process the sent emails in one of my WordPress plugins.

Synchro:
Since you know the addresses you're feeding in, could you not achieve the same thing by checking the return status from `send()`?

manzoorwanijk:
Emails are not sent by my plugin, rather they are sent by other plugins or by WordPress itself. So, I needed a hook by which I could monitor the sent email. I could not find anything better than this.

PR #1091, opened 2017-07-12
Add support for Sendmail's DeliveryMode option

balazslang:
We use sendmail to send different kinds of mails. When we send bulk mails, e.g. a newsletter, we do this in the background (-O DeliveryMode=b), because it is faster in general. But when we directly send a single mail and the DNS-Lookup gets no reply, the website is stuck for 30 seconds and then dies with a timeout. So it's important to us to have the posibility to change the DeliveryMode within the code.

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -625,6 +625,54 @@ class PHPMailer
     protected $uniqueid = '';
 
     /**
+     * Controls sendmail's delivery mode.
+     *
+     * @see http://sendmail.org/~ca/email/doc8.12/op-sh-4.html
+     * @see http://www.sendmail.org/~ca/email/man/sendmail.html
+     *
+     * DeliveryMode=x option described by sendmail's man:
+     * Set the delivery mode to x. Delivery modes are
+     * 'i' for interactive (synchronous) delivery,
+     * 'b' for background (asynchronous) delivery,
+     * 'q' for queue only - i.e., actual delivery is done
+     *     the next time the queue is run, and
+     * 'd' for deferred - the same as 'q' except that
+     *     database lookups (notably DNS and NIS lookups)
+     *     are avoided.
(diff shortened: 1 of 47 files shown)


Synchro:
This is a very nicely presented PR, however, I'm not sure it is worth incorporating into PHPMailer. If you want this functionality, it would be best implemented as a subclass, which you can maintain outside of PHPMailer without sacrificing the ability to upgrade. I don't really want to encourage the use of sendmail like this - if you need to send asynchronously, use a separate process and send with SMTP to localhost. By pushing this responsibility into sendmail, you're making obtaining error feedback much more difficult.

Also, I really don't want to accept new features into the 5.2 branch. It needs to be deprecated in favour of 6.0 as it is!

balazslang:
I see. The subclass-implementation is a possibility. Our problem is, that there is no unified way to control the parameters that will be passed as command to `popen()` or as `$params` to `mail()`. In case of `sendmailSend` we must inject our parameter into the `Sendmail` property, while with `mailSend` the `mailPassthru` method must be overwritten.
It would be nice to have a method, that builds this parameter list. That could be the point to inject some custom parameters.
Is there a way to make this method available in both 6.0 and 5.2? Is there a chance that it makes it into the release if we refactor the code in 5.2 and in 6.0 that way?

Synchro:
I don't think there's much point in unifying these. They are sufficiently different that there's little to be gained, and it's probably risky to try to make them act the same. I wouldn't recommend either the mail() nor sendmail transports. If you're doing anything serious, you should be using SMTP, and if you want asynchronous operations, you should be using separate PHP processes to manage it. I have no trouble sending around 700,000 messages/hour using this approach.

balazslang:
We've found a solution so this PR is obsolete. Thank you for your time and advice.

PR #1094, opened 2017-07-17
Correct bugs for multiple SMTP server insertion

eros23:
preg_match scheme error

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -1622,7 +1622,7 @@ class PHPMailer
 
         foreach ($hosts as $hostentry) {
             $hostinfo = array();
-            if (!preg_match('/^((ssl|tls):\/\/)*([a-zA-Z0-9:\[\]\.-]*):?([0-9]*)$/', trim($hostentry), $hostinfo)) {
+            if (!preg_match('/^((ssl|tls):\/\/)*([a-zA-Z0-9\[\]\.-]*):?([0-9]*)$/', trim($hostentry), $hostinfo)) {
                 // Not a valid host entry
                 continue;
             }


Synchro:
Can you provide test cases to show what this fixes please?

Synchro:
I've had more of a look at this: removing that `:` will break IPv6 literals, which is why it's there along with `[` and `]`. The match is not terribly thorough, but it's not meant to be.

eros23:
When i use ssl://smtp.name.it:25;tls://smtp.test2.it:586
The script creates this situation:
Connection: opening to tls://smtp.test2.it:586:25, timeout=300, options=array (\....

It does not find the fourth optional parameter (port) and add the standard port.

This is for domain name. For ipv6, i think i need to make a special regex

For this:
`^((ssl|tls):\/\/)*([a-zA-Z0-9\.-]*):?([0-9]*)$`

I have:
`
Full match | 0-23 | tls://smtp.test2.it:586
Group 1. | 0-6 | tls://
Group 2. | 0-3 | tls
Group 3. | 6-19 | smtp.test2.it
Group 4. | 20-23 | 586
`
For this:
`^((ssl|tls):\/\/)*([a-zA-Z0-9:\[\]\.-]*):?([0-9]*)$`

I have:
`
Full match | 0-23 | tls://smtp.test2.it:586
Group 1. | 0-6 | tls://
Group 2. | 0-3 | tls
Group 3. | 6-23 | smtp.test2.it:586
Group 4. | 23-23 | null

`

Synchro:
@eros23 You should find this is fixed now. I changed it to check for IPv6 separately.

PR #1116, opened 2017-08-02
IMAP example

NiklasBr:
An example on how to use Gmail and IMAP to save sent messages


Synchro:
This is nice, but it has so much in common with the existing gmail example I think it would be better merged into that - that way things relating to gmail are all in one place.

NiklasBr:
Absolutely.

PR #1126, opened 2017-08-12
league/oauth2-google is require

ashickurnoor:
league/oauth2-google is require to send mail. By Adding league/oauth2-google into composer.json it will download required library. 

https://github.com/PHPMailer/PHPMailer/blob/master/composer.json#L47


Synchro:
Thanks

PR #1155, opened 2017-09-11
Provide SMTP Transaction Id in Callback

deebuck:
SMTP now obtains the last transaction id from the reply to the DATA command. This change enables the transaction id to be provided through the callback mechanism to the caller. 
An additional parameter is added to the callback, which contains the smtp_transaction_id. 
This also adds a new pattern for the reply to the DATA command for Amazon SES.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -473,6 +473,8 @@ class PHPMailer
      *   string  $subject       the subject
      *   string  $body          the email body
      *   string  $from          email address of sender
+     *   string  $extra         extra information of possible use
+     *                          "smtp_transaction_id' => last smtp transaction id
      *
      * @var string
      */
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -1530,7 +1530,8 @@ class PHPMailer
                     $this->bcc,
                     $this->Subject,
                     $body,
-                    $this->From
+                    $this->From,
+                    []
                 );
                 if (0 !== $result) {
                     throw new Exception($this->lang('execute') . $this->Sendmail, self::STOP_CRITICAL);
@@ -1551,7 +1552,8 @@ class PHPMailer
                 $this->bcc,
                 $this->Subject,
                 $body,
-                $this->From
+                $this->From,
+                []
             );
             if (0 !== $result) {
                 throw new Exception($this->lang('execute') . $this->Sendmail, self::STOP_CRITICAL);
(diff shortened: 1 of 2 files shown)


src/PHPMailer.php
```diff
@@ -473,6 +473,8 @@ class PHPMailer
      *   string  $subject       the subject
      *   string  $body          the email body
      *   string  $from          email address of sender
+     *   string  $extra         extra information of possible use
```
Synchro: The type in that param tag should be `array` or `string[]`

src/PHPMailer.php
```diff
@@ -1638,11 +1642,11 @@ protected function mailSend($header, $body)
         if ($this->SingleTo and count($toArr) > 1) {
             foreach ($toArr as $toAddr) {
                 $result = $this->mailPassthru($toAddr, $this->Subject, $body, $header, $params);
-                $this->doCallback($result, [$toAddr], $this->cc, $this->bcc, $this->Subject, $body, $this->From);
+                $this->doCallback($result, [$toAddr], $this->cc, $this->bcc, $this->Subject, $body, $this->From,[]);
```
Synchro: Spaces after commas please

src/PHPMailer.php
```diff
@@ -1638,11 +1642,11 @@ protected function mailSend($header, $body)
...
             }
         } else {
             $result = $this->mailPassthru($to, $this->Subject, $body, $header, $params);
-            $this->doCallback($result, $this->to, $this->cc, $this->bcc, $this->Subject, $body, $this->From);
+            $this->doCallback($result, $this->to, $this->cc, $this->bcc, $this->Subject, $body, $this->From,[]);
```
Synchro: Spaces after commas please

src/PHPMailer.php
```diff
@@ -1726,20 +1730,31 @@ protected function smtpSend($header, $body)
...
             $this->smtp->reset();
         } else {
             $this->smtp->quit();
             $this->smtp->close();
         }
+        
+        foreach ($callbacks as $cb)
+        {
+            $this->doCallback($cb['issent'],array($cb['to']),array(),array(),$this->Subject,$body,$this->From,
+                ["smtp_transaction_id"=>$smtp_transaction_id]);
```
Synchro: Single-quote assoc keys, spaces around `=>` please.

src/SMTP.php
```diff
@@ -1291,6 +1292,7 @@ protected function recordLastTransactionID()
             foreach ($this->smtp_transaction_id_patterns as $smtp_transaction_id_pattern) {
                 if (preg_match($smtp_transaction_id_pattern, $reply, $matches)) {
                     $this->last_smtp_transaction_id = $matches[1];
+                    break;
```
Synchro: Good point!

PR #1159, opened 2017-09-14
Enforce PSR-4

Slamdunk:
`test/phpmailerLangTest.php` and `test/phpmailerTest.php` filenames don't match their classes, therefore violating PSR-4.

Adopt PSR-4, and enforce it through the dedicated coding-standard rule.

Along the way, a space fix emerged since you merged #1148 after one of your recent commits.


Synchro:
Thanks again.

PR #1161, opened 2017-09-14
Show debug log only when tests fail

Slamdunk:
As of yet the default debug log fills travis log with ten thousands line, unbearable to read and browse.

This PR shows the entire debug log, but only for failed tests.

I am willing to reorganize the source in `./test` dir, but better take one step at time.

### DIFF
--- a/test/PHPMailerTest.php
+++ b/test/PHPMailerTest.php
@@ -70,6 +70,7 @@ class PHPMailerTest extends \PHPUnit_Framework_TestCase
         }
         $this->Mail = new PHPMailer();
         $this->Mail->SMTPDebug = 3; //Full debug output
+        $this->Mail->Debugoutput = ['PHPMailer\Test\DebugLogTestListener', 'debugLog'];
         $this->Mail->Priority = 3;
         $this->Mail->Encoding = '8bit';
         $this->Mail->CharSet = 'iso-8859-1';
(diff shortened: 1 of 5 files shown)


test/PHPMailerTest.php
```diff
@@ -70,6 +70,7 @@ public function setUp()
         }
         $this->Mail = new PHPMailer();
         $this->Mail->SMTPDebug = 3; //Full debug output
+        $this->Mail->Debugoutput = ['PHPMailer\Test\DebugLogTestListener', 'debugLog'];
```
Synchro: That's a great idea

Synchro:
I've pushed a fix for what broke this build, please rebase.

The diffs that php-cs-fixer produces are a bit broken - after the last difference, they include the whole of the rest of the file. It would also be better if it included line numbers.

Slamdunk:
> The diffs that php-cs-fixer produces are a bit broken - after the last difference, they include the whole of the rest of the file. It would also be better if it included line numbers.

It do, but in newer versions.

I have in mind to do the coding standard checks first, before tests, and if they fail no test is executed in the build, so we can save a lot of times, and contributor notices them early. What do you think?

Rebased.

Synchro:
Fine by me - the tests take a long time as they include tests for timeouts - and it makes a point of doing those last.

PR #1199, opened 2017-10-09
Tibbel patch 2 alt body

Tibbel:
Before submitting your pull request, be sure your code adheres to PHPMailer

`./vendor/bin/php-cs-fixer fix`

And committing eventual changes.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -3758,7 +3758,7 @@ class PHPMailer
             }
         }
 
-        return $this->Body;        
+        return $this->Body;
     }
 
     /**
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -3745,24 +3745,20 @@ class PHPMailer
                 }
             }
         }
-         $this->isHTML(true);
-          // Convert all message body line breaks to LE, makes quoted-printable encoding work much better
+        $this->isHTML(true);
+        // Convert all message body line breaks to LE, makes quoted-printable encoding work much better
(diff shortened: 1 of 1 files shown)


Tibbel:
okay at this point `'' === $this->AltBody` can be replaced with `$this->alternativeExists()`
But the hole thing is that the AltBody is overwritten every time in the old code
and this case is such a stupid one, where `html2text(Body)` returns an empty string, it can be reached by setting Body only to Characters which will be replaced by `trim`.

Synchro:
But this is in `msgHTML`, which is specifically intended to set both `Body` and `AltBody` in one go, and you've got it doing an HTML to text conversion on something that should be plain text anyway? If you don't want AltBody overwritten, don't call `msgHTML`; just set `AltBody` explicitly. Can you explain the circumstances you're trying to address?

Tibbel:
okay, i think the problem is my script sets first `AltBody` and then calls `msgHTML`.
But from my point of view it's better to ignore the order of settings.

Synchro:
`msgHTML` is a convenience function that explicitly says it's going to overwrite both bodies. Applying `html2text` to an existing plain-text body is likely to cause damage - for example in my own apps I often wrap URLs in `<>` in plain-text bodies (because some clients treat them specially if you do that), and the `strip_tags` in `html2text` would destroy them. If you don't want the body-setting behaviour, don't call it, or set `AltBody` afterwards - which is also what the docs say.

PR #1203, opened 2017-10-14
Make idnSupported() static

fbonzon:
Upgrade guide and changelog say that idnSupported() is now static, but it actually isn't.

Probably this PR should wait v6.1 or other version where breaking changes are OK.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -973,7 +973,7 @@ class PHPMailer
         }
         $params = [$kind, $address, $name];
         // Enqueue addresses with IDN until we know the PHPMailer::$CharSet.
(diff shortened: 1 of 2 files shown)


Synchro:
How odd - I wonder if a commit got lost somewhere. I think this is safe to merge because calling a static method dynamically is permitted in PHP and does not generate any errors or warnings, and the method does not use `$this`, which is really the only restriction on static methods, so there's no BC break. Since it's documented that it already works that way, that it doesn't is effectively a bug that should be fixed. Anyone calling it statically, as documented, will find they no longer get errors for calling a dynamic method statically, so that makes it a transparent fix rather than a BC break. Thanks for the PR and for spotting this.

Issue #1212, opened 2017-10-27
phpmailer invalid address (punyEncode) root@localhost


cschivaz:
I'm just try tp use but i got Error "invalid address (punyEncode) root@localhost"
$mail = new PHPMailer(true);
	try{
	$mail->IsHTML(true);
	$mail->IsSMTP();
	$mail->SMTPAuth = true;
	$mail->SMTPSecure = 'tls';
	$mail->Host = 'smtp.gmail.com';
	$mail->Port = 587; 
	$mail->Username = '********@gmail.com'; 
	$mail->Password = ' *** '; 
	$mail->SetFrom = ('test@gmail.com'); 
	$mail->Subject = 'Test sending mail'; 
	$mail->Body = 'My Body & <b>My Description</b>';
	$mail->AddAddress('testEmail@gmail.com');
	$mail->send(); 
	} catch (Exception $e) {
    echo 'Message could not be sent.';
    echo 'Mailer Error: ' . $mail->ErrorInfo;
}

How can i solve this?

fbonzon:
Did you copy this code from an example? If yes, please give the reference, so it can be fixed.

    $mail->SetFrom = ('test@gmail.com');

This is incorrect usage of ```SetFrom```. It is a method, but your code treats it as a variable. Replace with:

    $mail->SetFrom('test@gmail.com');

cschivaz:
Thank you so much.

dileepverma107:
<?php
// Import PHPMailer classes into the global namespace
// These must be at the top of your script, not inside a function
use PHPMailer\PHPMailer\PHPMailer;
use PHPMailer\PHPMailer\Exception;

require 'src/Exception.php';
require 'src/PHPMailer.php';
require 'src/SMTP.php';
require 'src/OAuth.php';
require 'src/POP3.php';


//Load Composer's autoloader
require 'vendor/autoload.php';

$mail = new PHPMailer();                              // Passing `true` enables exceptions
try {
    //Server settings
    $mail->SMTPDebug = 2;                                 // Enable verbose debug output
    $mail->IsSMTP();                              // send via SMTP
$mail->Host = "smtp.gmail.com";
$mail->SMTPAuth = true;                       // turn on SMTP authentication
$mail->Username = "dileepverma107@gmail.com";        // SMTP username
$mail->Password = "aci@#4321";               // SMTP password
$webmaster_email = "dileepverma107@gmail.com";       //Reply to this email ID
                // Recipients email ID
[...]

PR #1213, opened 2017-10-27
Improve error message for invalid address

fbonzon:
Follow-up to https://github.com/PHPMailer/PHPMailer/commit/c2bde34e58caa500b8fb2f7e14fc9765596cdf6d. Will help users understand the error reason in cases like #1212.

- Adds the address kind (From, Sender, ConfirmReadingTo) instead of word punyEncode which is internal to PHPMailer and not so helpful.
- Similarly, removes word addAnAddress (internal) but keeps only address kind (to, cc, bcc, Reply-To).


Synchro:
Thanks

PR #1241, opened 2017-11-14
general code style fixes

glensc:
initially just wanted to submit `is_a` -> `instanceof` change, but resolved some other issues on the go. see individual commits of the changes.

if you're wondering then the strict comparison fixes were done only when both sides were non-empty strings, that is safe changes.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -4379,7 +4379,7 @@ class PHPMailer
     protected function doCallback($isSent, $to, $cc, $bcc, $subject, $body, $from, $extra)
     {
         if (!empty($this->action_function) and is_callable($this->action_function)) {
-            call_user_func_array($this->action_function, [$isSent, $to, $cc, $bcc, $subject, $body, $from, $extra]);
+            call_user_func($this->action_function, $isSent, $to, $cc, $bcc, $subject, $body, $from, $extra);
         }
     }
 
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -866,7 +866,7 @@ class PHPMailer
     {
         $ini_sendmail_path = ini_get('sendmail_path');
 
-        if (!stristr($ini_sendmail_path, 'sendmail')) {
+        if (false === stripos($ini_sendmail_path, 'sendmail')) {
             $this->Sendmail = '/usr/sbin/sendmail';
         } else {
             $this->Sendmail = $ini_sendmail_path;
@@ -881,7 +881,7 @@ class PHPMailer
     {
         $ini_sendmail_path = ini_get('sendmail_path');
 
-        if (!stristr($ini_sendmail_path, 'qmail')) {
+        if (false === stripos($ini_sendmail_path, 'qmail')) {
             $this->Sendmail = '/var/qmail/bin/qmail-inject';
         } else {
             $this->Sendmail = $ini_sendmail_path;
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -37,7 +37,7 @@ class PHPMailer
      *
      * @var int
      */
-    public $Priority = null;
+    public $Priority;
 
     /**
      * The character set of the message.
@@ -314,7 +314,7 @@ class PHPMailer
      *
      * @var OAuth
      */
-    protected $oauth = null;
+    protected $oauth;
 
     /**
      * The SMTP server timeout in seconds.
@@ -504,7 +504,7 @@ class PHPMailer
      *
      * @var SMTP
      */
-    protected $smtp = null;
+    protected $smtp;
 
     /**
      * The array of 'to' names and addresses.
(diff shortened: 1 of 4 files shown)


src/PHPMailer.php
```diff
@@ -362,7 +364,7 @@ class PHPMailer
      *
      * @see SMTP::$Debugoutput
      *
-     * @var string|callable|\Psr\Log\LoggerInterface
+     * @var string|callable|LoggerInterface
```
Synchro: Since this is optional, I prefer having this left as a FQCN; it's clearer where it comes from at the point of use.

src/PHPMailer.php
```diff
@@ -3380,7 +3383,7 @@ protected function cidExists($cid)
     public function inlineImageExists()
     {
         foreach ($this->attachment as $attachment) {
-            if ($attachment[6] == 'inline') {
+            if ($attachment[6] === 'inline') {
```
Synchro: This (and the next one) should be yoda-flipped

src/PHPMailer.php
```diff
@@ -1853,12 +1855,12 @@ public function smtpConnect($options = null)
             }
             $prefix = '';
             $secure = $this->SMTPSecure;
-            $tls = ('tls' == $this->SMTPSecure);
-            if ('ssl' == $hostinfo[2] or ('' == $hostinfo[2] and 'ssl' == $this->SMTPSecure)) {
+            $tls = ('tls' === $this->SMTPSecure);
```
Synchro: I know many go on about using strict comparisons, but this is often done dogmatically rather than for any specific reason. In cases like this we *really* don't care, and it *really* doesn't matter whether it's actually a string or something that can be coerced into one. The value submitted is never actually used directly, only compared against, and such unnecessary strictness runs counter to usability, very much an instance of Postel's law.

glensc: yeah, i picked only comparisons, that compare strings against strings as that is considered safe.

anyway, removing these two commits from this PR:
- Applying: strict in_array comparision
- Applying: strict string comparison

src/PHPMailer.php
glensc: rewritten the commit

src/PHPMailer.php
glensc: done, also more yoda.

altho my personal opinion it's counter to intuitive. forces one reading left to right to start reading from right to left.

and if it's intended to be protection against accidential assign, there's no correct way to solve if both sides are non-constants, you can still cause wrong behaviour by using wrong operator:

```php
                return $params[0] = $kind;
```

Synchro: I agree, and argued much the same on Symfony's code style recently - they argued you should do it everywhere, without exception, even where it made no difference! IMHO, the only real place worth living with the readability issue is in conditions with one constant - the payoff is that case is that inadvertent assignment becomes a compile-time error rather than a runtime mystery, and that's worth having.

glensc: having compile time error vs runtime error - it's already to late, the file is in production.

code quality and behavior should be tested with unit/integration-tests. broken code should never be merged to default branch (`master`) :)

Synchro: Of course, but it's all about reducing accidents, and compile-time errors are far easier to deal with whenever they happen, whether in dev or production, so I think yoda comparisons are a good defensive habit from that POV. Aside from that they are quite annoying...

glensc:
cs-fixer broken rule
https://travis-ci.org/PHPMailer/PHPMailer/jobs/303264823

```
   1) src/PHPMailer.php (phpdoc_to_comment, no_blank_lines_after_phpdoc, single_blank_line_before_namespace)
      ---------- begin diff ----------
--- Original
+++ New
@@ @@
              */
-            /** @noinspection PhpMissingBreakStatementInspection */
+            /* @noinspection PhpMissingBreakStatementInspection */
```



this is not good. PHPSTORM requires the `@noinspection` to be inside `/** */` to work. `/* *` will not work.

i'm going to remove the commit that changed that part to get this PR merged.

also the cs-fixer diff from that travis link is rather annoying to read, if it prints out whole file, not like regular diff which shows just few lines of context. off topic, but still annoying :)

Synchro:
I'm not entirely happy with the cs-fixer stuff either. It's good to be consistent in code style, and to enforce it (and the PRs to set it up were very welcome), but some of the rules in it (which claim to be "Symfony standard") do not appear to have a rational or documented basis. I've had trouble getting PHPStorm to recognise some of these annotations - or I think it may be that its own checkers conflict with those of cs-fixer, and you can't fix it for both at once. IMHO PHPStorm is more important, so whatever rule cs-fixer is enforcing there should be dropped. Save that for another PR...

glensc:
i  think there's bug in cs-fixer fixer, as some other annotation in same file doesn't get reported. likely related to consecutive comment blocks.

Synchro:
Thanks

PR #1249, opened 2017-11-20
Add Sender to MIMEHeader

adrium:
The `Sender` header is not sent. This PR fixes this.

PS:

1. I ran `./vendor/bin/php-cs-fixer`
2. I could not get PHPUnit to run, lots of errors. If somebody can tell me the exact command, I can test my commit. I used: `./vendor/bin/phpunit --bootstrap=test/bootstrap.php test`

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -2273,6 +2273,12 @@ class PHPMailer
 
         $result .= $this->addrAppend('From', [[trim($this->From), $this->FromName]]);
 
+        if ('' == $this->Sender || $this->From == $this->Sender) {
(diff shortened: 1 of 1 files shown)


Synchro:
Sorry, this is not how it works. The `Sender` property is used to set the SMTP-level envelope sender - the address that is used in a `MAIL FROM` SMTP command. It should never be put in a header, though it's very common for it to be the same value as what is in the "From" header, it's also common for it to be entirely different (e.g for VERP addressing). The envelope sender is added to a message in the form of a `Return-Path` header *by the receiver*, so again, this is not something you should add as a sender.

PR #1259, opened 2017-11-30
Update phpmailer.lang-eo.php

dknacht:
Adding the missing translation.
Due to the flexibility of Esperanto, there is actually many ways of saying "Extension missing". I selected this one, even tho it's not the literal translation (what would rather be more like "Manki etendo" as the "missing" is used as an adjective, not as a verb), because I think this way is much more understandable.


Synchro:
Thanks!

PR #1290, opened 2017-12-15
Added space between error

RAZERZ:
Easier to read

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ try {
     $mail->send();
     echo 'Message has been sent';
 } catch (Exception $e) {
-    echo 'Message could not be sent.';
+    echo 'Message could not be sent. ';
     echo 'Mailer Error: ' . $mail->ErrorInfo;
 }
 ```


README.md
```diff
@@ -124,7 +124,7 @@ try {
     $mail->send();
     echo 'Message has been sent';
 } catch (Exception $e) {
-    echo 'Message could not be sent.';
+    echo 'Message could not be sent. ';
```
Synchro: I agree that's not optimal - but it might be better to add a line break, or simpler, merge these two lines into one.

RAZERZ: Correct, I do agree on the line break, however it would be even better if we added a bold for the preprogrammed messages and then let the error message come after, like this:

echo '<b>Message could not be sent. </b>';

:)

Synchro: Bold doesn't exist in plain text!

RAZERZ: I added <br> tags but they didn't show up haha

Synchro: No, don't add tags - it's meant to work in plain text. I suggest changing it to:

    echo 'Message could not be sent. Mailer Error: ', $mail->ErrorInfo;

That will produce exactly the same output as your original PR, but it's slightly more efficient.

Synchro:
Implemented in 2834867

PR #1292, opened 2017-12-17
Update README.md

dknacht:
Added links to the recommended optional libraries.
Changed order of "Upgrading from 5.2" and "Legacy versions", i think this way is more logical to read.
Updated language quantity in the "Localization" section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Build status: [![Build Status](https://travis-ci.org/PHPMailer/PHPMailer.svg)](h
 Many PHP developers utilize email in their code. The only PHP function that supports this is the `mail()` function. However, it does not provide any assistance for making use of popular features such as HTML-based emails and attachments.
 
 Formatting email correctly is surprisingly difficult. There are myriad overlapping RFCs, requiring tight adherence to horribly complicated formatting and encoding rules - the vast majority of code that you'll find online that uses the `mail()` function directly is just plain wrong!
-*Please* don't be tempted to do it yourself - if you don't use PHPMailer, there are many other excellent libraries that you should look at before rolling your own - try SwiftMailer, Zend_Mail, eZcomponents etc.
+*Please* don't be tempted to do it yourself - if you don't use PHPMailer, there are many other excellent libraries that you should look at before rolling your own - try [SwiftMailer](https://swiftmailer.symfony.com/), [Zend/Mail](https://zendframework.github.io/zend-mail/), [eZcomponents](http://ezcomponents.org/) etc.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,7 +29,7 @@ Build status: [![Build Status](https://travis-ci.org/PHPMailer/PHPMailer.svg)](h
 Many PHP developers utilize email in their code. The only PHP function that supports this is the `mail()` function. However, it does not provide any assistance for making use of popular features such as HTML-based emails and attachments.
 
 Formatting email correctly is surprisingly difficult. There are myriad overlapping RFCs, requiring tight adherence to horribly complicated formatting and encoding rules - the vast majority of code that you'll find online that uses the `mail()` function directly is just plain wrong!
-*Please* don't be tempted to do it yourself - if you don't use PHPMailer, there are many other excellent libraries that you should look at before rolling your own - try SwiftMailer, Zend_Mail, eZcomponents etc.
+*Please* don't be tempted to do it yourself - if you don't use PHPMailer, there are many other excellent libraries that you should look at before rolling your own - try [SwiftMailer](https://swiftmailer.symfony.com/), [Zend/Mail](https://zendframework.github.io/zend-mail/), [eZcomponents](http://ezcomponents.org/) etc.
```
Synchro: This needs updating - ezComponents became zeta components, was an apache incubator project, then moved to github, so its link should be: https://github.com/zetacomponents/Mail

README.md
```diff
@@ -66,16 +66,16 @@ require 'path/to/PHPMailer/src/PHPMailer.php';
 require 'path/to/PHPMailer/src/SMTP.php';
 ```
 
-If you're not using the `SMTP` class explicitly (you're probably not), you don't need a `use` line for the SMTP class.
+If you're not using the `SMTP` class explicitly (you're probably not), you don't need a `require` line for the SMTP class.
```
Synchro: The original is correct here. You always need a `require` line for the SMTP class if you're using SMTP (without composer), but you don't need a `use` line unless your'e using it directly, i.e. creating your own SMTP instances.

dknacht:
Hi!
Thanks so much for your review.
I will make the corrections you mentioned and by the way I realized there is a couple of extra changes that I can make (Correct a little mistake in Spanish translation, complete the Swedish translation (I don't speak Swedish but my English teacher does)).
I don't manage really well git hub; I thought that I could make a pull request for new changes but it seems like it pulls again all the files so I will close this, make all the changes and make a new PR.

Synchro:
You don't need to make a new PR - You can make changes that revert/update/add to your earlier commits and push them to your fork and they will be added to this PR, then when it's merged it will squash them into a single commit to keep history tidy.

Issue #1298, opened 2017-12-19
Missing dependency on the PHP "filter" extension


orlitzky:
In https://github.com/PHPMailer/PHPMailer/blob/9d9021ebc399f519b9bba9bb5cf18702fb5f4ac0/src/PHPMailer.php#L1252 the `filter_var` function is called, and that function is provided by the [PHP filter extension](http://php.net/manual/en/book.filter.php).

That extension should be required in https://github.com/PHPMailer/PHPMailer/blob/master/composer.json, right next to where "ctype" is.

Synchro:
That's true, but the filter extension has been enabled by default since PHP 5.2. There are probably dozens of other extensions that are implicitly relied on too, but they're not listed for the same reason. Have you actually encountered an installation that doesn't include the filter ext? Ctype was specifically added because FreeBSD's PHP packages lack it. PRs welcome.

orlitzky:
In Gentoo and its derivatives, it's easy to build PHP without any one particular extension. In theory, all packages require the extensions that they need, so if the package manager lets you disable an extension, then doing so should be cool. Using that reasoning, I just built myself a PHP that lacked _filter_, and noticed the omission.

I've already added the extra dependency to the Gentoo package, but anyone building PHP from source and using Composer should be able to shoot himself in the foot in the same way.

orlitzky:
The PHP that I built has *only* filter and ctype enabled now, because all it does is run one contact form. If our web developer tells me that it doesn't work, I'll be happy to figure out which other (enabled by default) extensions are needed.

Synchro:
OK, I'll leave this open then, will be happy to merge a PR when you've got a list.

orlitzky:
The "openssl" extension is also needed for STARTTLS in SMTP. With that enabled, I was able to send via both SMTP and PHP's `mail()` function.

orlitzky:
Actually, there are checks in place for most of the uses of OpenSSL. In the places where the extension is required but absent, an exception is thrown. I've opened https://github.com/PHPMailer/PHPMailer/issues/1300 for one case that I think was simply overlooked.

The "filter" functions are however still used unconditionally and unavoidably, so I think my initial report was correct that was should depend on the filter extension (PR incoming).

PR #1301, opened 2017-12-21
composer.json: add dependency on the PHP "filter" extension.

orlitzky:
Many of the PHP extensions used by PHPMailer are optional, in one of
two senses. Either there is a fallback if the extension is not
present, or the use of the extension is avoidable entirely. For
example, there is a fallback parser if "imap_rfc822_parse_adrlist" is
not available, and OpenSSL is not needed unless encryption is used;
therefore neither are hard dependencies.

The filter extension, on the other hand, is unavoidable. It is used
unconditionally, with no fallbacks, in (for example) the "isValidHost"
function. This commit adds "ext-filter" to composer.json, to document
the dependency and to help out composer users.

Closes: https://github.com/PHPMailer/PHPMailer/issues/1298

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,8 @@
     ],
     "require": {
         "php": ">=5.5.0",
-        "ext-ctype": "*"
+        "ext-ctype": "*",
+        "ext-filter": "*"
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


Synchro:
Thanks for your research into this, and the PR.

PR #1317, opened 2018-01-05
fix code redundancy.

GaithD:
Fix code redundancy reduce process to one instead of two.


Synchro:
Thanks, and welcome to Github!

PR #1338, opened 2018-01-23
change array_key_exists to empty

leadboots5:
The array keys will always be set since post will just send them through with empty values. using empty checks if they actually have anything so its a better way to evaluate.

### DIFF
--- a/examples/simple_contact_form.phps
+++ b/examples/simple_contact_form.phps
@@ -9,18 +9,18 @@ use PHPMailer\PHPMailer\PHPMailer;
 
 require '../vendor/autoload.php';
 
-if (array_key_exists('to', $_POST)) {
+if (!empty($_POST['to'])) {
     $err = false;
     $msg = '';
     $email = '';
     //Apply some basic validation and filtering to the subject
-    if (array_key_exists('subject', $_POST)) {
+    if (!empty($_POST['subject'])) {
         $subject = substr(strip_tags($_POST['subject']), 0, 255);
     } else {
         $subject = 'No subject given';
     }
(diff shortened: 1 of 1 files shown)


ooxi:
Why would the keys be set, you could access the page with a `GET` request.

leadboots5:
Because that's how PHP is coded? I don't know why the developers took that direction when creating PHP.

Synchro:
I really don't like using `empty` like that because of its implicit error suppression, which means you can't tell why it's failing. For example:

```php
$a = ['a' => 1, 'b' => null];
echo empty($a['b']); //true
echo empty($a['c']); //also true, but you don't know why
//but:
echo $a['b']; //Outputs nothing
echo $a['c']; //Triggers unknown array index error
```

`array_key_exists()` does not share that ambiguity, because it checks whether the index exists, not whether the value it points to is empty.

I want that first test for `array_key_exists('to', $_POST)` to check specifically whether we are looking at a `POST` submission containing expected form fields before proceeding to input validation (which may quite reasonably use `empty` for checking field contents, not existence); checking for `!empty($_POST['to'])` is relying on a non-obvious peculiarity of PHP (that `empty` suppresses errors), and while that test will work in practice, it may be doing so for the wrong reason, and wrong reasons are the foundations of future bugs. I put it firmly in the same category as using `@` to suppress errors; it's worth avoiding.

leadboots5:
@Synchro I started this thinking of subject/query which will never reach their else/failure when blank and blanket applied that logic for the rest of the instances where it doesn't apply. Thanks for the review.

PR #1378, opened 2018-03-16
Readme composer

NiklasBr:
A small improvement to the Readme: Link to the Composer site, which hopefully helps somewhat when people want to install. Also capitalises the Composer name, and removes an odd line break.


Synchro:
Thanks, nice improvement.

PR #1383, opened 2018-03-21
Add support of _ in hostname

soywod:


soywod:
The _ character is allowed in URLs, but not in Hostname

PR #1390, opened 2018-03-26
Updated default language path

notian:
The default language path points to src/languages/ but the languages dir is one level down src/../languages/ for BC the default path should be updated, since many people probably point to the existing languages folder directly.

```
if (empty($lang_path)) {
            // Calculate an absolute path so it can work if CWD is not here
            $lang_path = __DIR__ . DIRECTORY_SEPARATOR . 'language' . DIRECTORY_SEPARATOR;
        }
```

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -1994,7 +1994,7 @@ class PHPMailer
         ];
         if (empty($lang_path)) {
             // Calculate an absolute path so it can work if CWD is not here
-            $lang_path = dirname( __DIR__ ) . DIRECTORY_SEPARATOR . 'language' . DIRECTORY_SEPARATOR;
+            $lang_path = dirname(__DIR__) . DIRECTORY_SEPARATOR . 'language' . DIRECTORY_SEPARATOR;
         }
         //Validate $langcode
         if (!preg_match('/^[a-z]{2}(?:_[a-zA-Z]{2})?$/', $langcode)) {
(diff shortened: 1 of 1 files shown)


src/PHPMailer.php
```diff
@@ -1994,7 +1994,7 @@ public function setLanguage($langcode = 'en', $lang_path = '')
         ];
         if (empty($lang_path)) {
             // Calculate an absolute path so it can work if CWD is not here
-            $lang_path = __DIR__ . DIRECTORY_SEPARATOR . 'language' . DIRECTORY_SEPARATOR;
+            $lang_path = __DIR__ . DIRECTORY_SEPARATOR . '..' . DIRECTORY_SEPARATOR . 'language' . DIRECTORY_SEPARATOR;
```
Synchro: Rather than building a path string containing a relative path that will need to be interpreted, I think it's probably better to calculate the parent folder directly, something like:

    $lang_path = dirname(__DIR__) . DIRECTORY_SEPARATOR . 'language' . DIRECTORY_SEPARATOR;

notian: Done.

Synchro:
Thanks. I'd spotted this problem a while ago but not got around to reporting/fixing it.

notian:
When do you plan to merge master into 6.0? I can update my composer to dev-master, but I'd rather stick to a numeric version.

Synchro:
`master` will never be merged into `6.0` - master *is* currently the 6.0 branch. The branch called 6.0 is from before 6.0 was released and merged into master; it should probably be deleted. Essentially I just need to tag a release; I'll aim to do that tomorrow.

Synchro:
I just tagged a 6.0.4 release containing this fix.

PR #1429, opened 2018-04-26
Add mail header constants, fixes #1428

### LINKED ISSUE #1428: Add constants for standard charsets, content-types and encodings
It would be helpful to have constants for the following values:
charset: 'iso-8859-1', 'utf-8'
content-type: 'text/plain', 'text/html',...
encoding: "8bit", "7bit", "binary", "base64" and "quoted-printable"

maxbeckers:
Add constants for mail headers.


src/PHPMailer.php
```diff
@@ -44,22 +65,22 @@ class PHPMailer
      *
      * @var string
      */
-    public $CharSet = MailHeaderConstants::CHARSET_ISO88591;
+    public $CharSet = self::DEFAULT_CHARSET;
```
Synchro: I think these should all use `static` rather than `self` so that they are easily overridable: `self` always refers to the base class, even in a child class ([docs](http://php.net/manual/en/language.oop5.late-static-bindings.php)), and uses like this are exactly the kind of thing you might want to override in a subclass.

The defaults don't need to do that since we don't expect the values for things like charsets to ever change - `utf-8` will always be `utf-8`.

Synchro: You're quite correct, thanks. Annoyingly that does somewhat defeat the point of using the `DEFAULT_*` values; they should probably go back to more like you originally had them, like `$CharSet = self::CHARSET_ISO88591;`, since you can't sensibly override the value of the DEFAULT constant in a subclass and have it be used there, so you should just override the value itself directly. Apologies for the detour.

maxbeckers: Yes you're right, I'll change

Synchro:
Thanks

PR #1463, opened 2018-05-26
Create new language file: Tagalog (Philippines)

StoneArtz:


StoneArtz:
The language code for Tagalog is 'tl'. I'm renaming it

Issue #1406, opened 2018-04-04
DKIM fails when i change the length of the subject 


lukasmedia:
Allright; this case is breaking my moral over here. Hope you guys can help.

I've setup PHPMailer on my webserver (Centos, PHP5.6) using Exim 4.89 to send mails. I've setup SPF and DKIM correctly. Using https://mxtoolbox.com/dkim.aspx i can validate the DKIM record. When i send an e-mail to http://dkimvalidator.com/ and/or https://www.mail-tester.com/ all is OK (10/10 @mail-tester.com!).

**# Problem description**
The current subject of the test e-mail is "Reminder give feedbäck to R1 R1 / Reminder" --> OK.
When i change it to: "Reminder give feedbäck to R1 R1 / Reminder to Ronald" the DKIM check fails.

**# Code to reproduce**
`date_default_timezone_set('Etc/UTC');
use PHPMailer\PHPMailer\PHPMailer;

$key = '-----BEGIN RSA PRIVATE KEY-----
-----END RSA PRIVATE KEY-----';
$selector = '1520190215.mytestdomain';
$to       = 'test-xxx@appmaildev.com'; 
$sender = 'mytestdomain.nl';

$mail = new PHPMailer;

$mail->setFrom('noreply@' . $sender, 'Mail from ' . $sender);
$mail->addReplyTo('noreply@' . $sender, 'Mail from ' . $sender);
$mail->addAddress($to);

// OK, 23 chars with 1 special char:
//$mail->Subject = "Reminder feedbäck to R1";

//  NOT OK, 1 char. too long? Strlen: 46
$mail->Subject = "Reminder give feedbäck to R1 R1 / Reminder to Ronald";

// NOT OK, 1 char too long. Strlen: 77
//$mail->Subject = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"; 

// Strlen 63, NOT OK, 1 char too long.
//$mail->Subject = "Reminder give feedback to R1 R1/ Reminder Reminder Reminder Remi"; 

$htmlMsg = '<html><body>This is a test!!!
I dont give ä damn!</body></html>';

$mail->msgHTML($htmlMsg);
//$mail->AltBody = strip_tags($htmlMsg);

$mail->DKIM_domain          = $sender;
$mail->DKIM_selector        = $selector; 
[...]

Issue #1352, opened 2018-02-06
Dkim fails with Altbody


NicolasINATIC:
Hi,

When I use phpmailer (6.0.3) without AltBody DKIM pass but when I add AltBody DKIM fail.

Would you know why ?

Best regards,

My code :

```php
<?php
$headers  = "MIME-Version: 1.0\r\n";
$headers .= "Content-type: text/html; charset=iso-8859-1\r\n";
$subject = 'TEST';
$HTMLmessage = '
<html>
	<body>
		<p>
			<b>Message</b> de test en html
		</p>
	</body>
</html>';

$TEXTmessage = 'Message de test en text';
$mail = new PHPMailer();
$mail->isSMTP(); // Set mailer to use SMTP
$mail->Host = 'xxxxxxxx'; // Specify main and backup SMTP servers   ns0.ovh.net
$mail->SMTPAuth = true; // Enable SMTP authentication
$mail->Username = 'xxxxxxxx'; // SMTP username
$mail->Password = 'xxxxxxxx'; // SMTP password
$mail->SMTPSecure = 'tls'; // tls sur port 587, ssl sur port 465
$mail->Port = 587; // TCP port to connect to
$mail->From = "xxxxxxxx";
$mail->FromName = "xxxxxxxx";
$mail->AddAddress("xxxxxxxx");
$mail->Subject = $subject;
$mail->IsHTML(TRUE);
$mail->Body = $HTMLmessage;
$mail->AltBody  =  $TEXTmessagetext;            //if I disable this, it pass
$mail->Encoding = "base64";
$mail->DKIM_domain = 'xxxxxxxxx';
$mail->DKIM_private = 'dkim/dkim.private';
$mail->DKIM_selector = 'xxxxxxx';
$mail->DKIM_passphrase = '';
$mail->DKIM_identity = $mail->From;
if(!$mail->Send())
{
        $error_message = "Mailer Error: " . $mail->ErrorInfo;
        echo $error_message;
} else 
{
        $error_message = "Successfully sent!";
        echo $error_message;
}
?>
```

NicolasINATIC:
"
$headers  = "MIME-Version: 1.0\r\n";
$headers .= "Content-type: text/html; charset=iso-8859-1\r\n";

$subject = 'TEST';

$HTMLmessage = '
<html>
	<body>
		<p>
			<b>Message</b> de test en html
		</p>
	</body>
</html>';

$TEXTmessage = 'Message de test en text';

[...]

PR #1468, opened 2018-06-04
Add flag for using copiedHeaderFields and config for using extra headers in DKIM signature

gwi-mmuths:
According to https://tools.ietf.org/html/rfc6376#section-3.5:
"Copied header fields (dkim-quoted-printable, but see description;
OPTIONAL, default is null). ..."
"Copied header field values are for diagnostic use."

Disabling this information can save network traffic.
For backward compatibility its enabled by default.

In addition you can add custom extra headers to the DKIM signature, for instance `List-Unsubscribe` for newsletter, which is recommend by [CSA](https://certified-senders.org/)

### DIFF
--- a/test/PHPMailerTest.php
+++ b/test/PHPMailerTest.php
@@ -1911,6 +1911,92 @@ EOT;
     }
 
     /**
+     * DKIM copied header fields tests.
+     *
+     * @group dkim
+     *
+     * @see https://tools.ietf.org/html/rfc6376#section-3.5
+     */
+    public function testDKIMOptionalHeaderFieldsCopy()
+    {
+        $privatekeyfile = 'dkim_private.pem';
+        $pk = openssl_pkey_new(
+            [
+                'private_key_bits' => 2048,
+                'private_key_type' => OPENSSL_KEYTYPE_RSA,
+            ]
+        );
+        openssl_pkey_export_to_file($pk, $privatekeyfile);
+        $this->Mail->DKIM_private = 'dkim_private.pem';
+
+        //Example from https://tools.ietf.org/html/rfc6376#section-3.5
+        $from = 'from@example.com';
+        $to = 'to@example.com';
+        $date = 'date';
+        $subject = 'example';
+
+        $headerLines = "From:$from\r\nTo:$to\r\nDate:$date\r\n";
+        $copyHeaderFields = "\tz=From:$from\r\n\t|To:$to\r\n\t|Date:$date\r\n\t|Subject:=20$subject;\r\n";
+
+        $this->Mail->DKIM_copyHeaderFields = true;
+        $this->assertContains(
+            $copyHeaderFields,
+            $this->Mail->DKIM_Add($headerLines, $subject, ''),
+            'DKIM header with copied header fields incorrect'
+        );
+
+        $this->Mail->DKIM_copyHeaderFields = false;
+        $this->assertNotContains(
+            $copyHeaderFields,
+            $this->Mail->DKIM_Add($headerLines, $subject, ''),
+            'DKIM header without copied header fields incorrect'
+        );
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


Synchro:
Could you squash and force-push all these commits so that I can look at a full diff please?

test/PHPMailerTest.php
```diff
@@ -1910,6 +1910,92 @@ public function testDKIMHeaderCanonicalization()
...
+        $this->Mail->DKIM_private = 'dkim_private.pem';
+
+        //Example from https://tools.ietf.org/html/rfc6376#section-3.5
+        $from = 'from@example.com';
+        $to = 'to@example.com';
+        $date = 'date';
+        $subject = 'example';
+        $anyHeader = 'foo';
+        $unsubscribeUrl = '<https://www.example.com/unsubscribe/?newsletterId=anytoken&amp;actionToken=anyToken' .
+                            '&amp;otherParam=otherValue&amp;anotherParam=anotherVeryVeryVeryLongValue>';
```
Synchro: This URL shouldn't use `&amp;`; it's not in an HTML or XML context where that would be appropriate.

test/PHPMailerTest.php
```diff
@@ -1910,6 +1910,92 @@ public function testDKIMHeaderCanonicalization()
...
+    {
+        $privatekeyfile = 'dkim_private.pem';
+        $pk = openssl_pkey_new(
+            [
+                'private_key_bits' => 2048,
+                'private_key_type' => OPENSSL_KEYTYPE_RSA,
+            ]
+        );
+        openssl_pkey_export_to_file($pk, $privatekeyfile);
+        $this->Mail->DKIM_private = 'dkim_private.pem';
```
Synchro: Use `$privatekeyfile` here. Also this file should be deleted after (and possibly before) the test run so we don't leave a mess.

test/PHPMailerTest.php
```diff
@@ -1910,6 +1910,92 @@ public function testDKIMHeaderCanonicalization()
...
+
+        $headerLines = "From:$from\r\nTo:$to\r\nDate:$date\r\n";
+        $headerLines .= "X-AnyHeader:$anyHeader\r\nBaz:bar\r\n";
+        $headerLines .= 'List-Unsubscribe:' . $this->Mail->encodeHeader($unsubscribeUrl) . "\r\n";
+
+        $headerFields = 'h=From:To:Date:Subject:Baz:List-Unsubscribe';
+
+        $result = $this->Mail->DKIM_Add($headerLines, $subject, '');
+
+        $this->assertContains($headerFields, $result, 'DKIM header with extra headers incorrect');
```
Synchro: This is a somewhat shallow test - it might be useful to go all the way through to `getSentMIMEMessage()` and check that the headers in there contain the properly encoded headers. I suspect the DKIM test coverage isn't great in this area, though this isn't specific to your code.

test/PHPMailerTest.php
gwi-mmuths: Done

test/PHPMailerTest.php
gwi-mmuths: That's true, but it's enough for me.

It would be better having an end2end test by checking DKIM signature is passing at all.

gwi-mmuths:
I also added the features to the example.

test/PHPMailerTest.php
Synchro: True - DKIM is tricky at the best of times. Are DKIM checks passing on live servers (e.g. on gmail) in your own tests using this code?

gwi-mmuths: Yes it's passing, this fork is already in production ;)

Synchro:
Thanks for your effort.

Synchro:
If you're looking into DKIM stuff, I could do with some help on #1352 and #1406

PR #1503, opened 2018-07-21
reverting member attribute to local variable

doganoo:
in preSend() method, I have seen that there is an member attribute created within an foreach loop. The member attribute (created with an dollar sign, $this->**$**address_kind) is only used within the loop and therefore, I think it does not make sense to use it globally (as an member attribute).

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -1412,16 +1412,16 @@ class PHPMailer
 
             // Validate From, Sender, and ConfirmReadingTo addresses
             foreach (['From', 'Sender', 'ConfirmReadingTo'] as $address_kind) {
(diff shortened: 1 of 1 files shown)


Synchro:
I think you’ve entirely misunderstood how this works. $address_kind takes values for to, bcc, cc etc, all of which are static properties. Addresd_kind is not used as a property at all.

doganoo:
@Synchro probably. But $this->$address_kind is never used anywhere else?!

Synchro:
You're missing something - it's `$this->$address_kind` **NOT** `$this->address_kind`. `$address_kind` is a local variable that holds the name of the property it's referring to.

doganoo:
Hm, ok. So then it is my fault, sorry. I close the pull request.

PR #1520, opened 2018-08-13
Allow for setting multiple emails through addAddress

agentd00nut:
Either by comma separated string to addAddress
Or by array to addAddresses.

 $mail->addAddress('test@test.com,anotherEmail@test.com')
 $mail->addAddresses( array( array('test@test.com', 'testName'), array('anotherEmail@test.com') ))
 $mail -> addAddresses( array('ThisIsSupported@too.com', 'but Not recommended') );

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -931,7 +931,37 @@ class PHPMailer
      */
     public function addAddress($address, $name = '')
     {
+        if(strpos($address, ",") !== false && $name == ''){
+            $state = True;
+            foreach( explode(",",$address) as $a){
+                if($this->addOrEnqueueAnAddress('to', $address, $name) == False){
+                    $state = False;
+                }
+            }
+            return $state;
+        }
         return $this->addOrEnqueueAnAddress('to', $address, $name);
+    }
+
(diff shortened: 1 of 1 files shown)


Synchro:
This is a bit more complex than you make out.

The split of addresses in `addAddress` by comma is too naïve - addresses can contain commas - for example `"smith,jane"@example.com`" is a valid address. Parsing these reliably is extremely difficult, so it's easier to not do so and limit this method to adding a single address, i.e. what it already does.

If you have one address that fails validation, you fail the whole collection, so you can't tell which one was bad, though if exceptions are enabled it will throw an exception (from `addOrEnqueue...`) as soon as it hits a bad address. Because the source format used for multiple addresses varies so much (comma-delimited, semi-colon delimited, line break-delimited, in an array (with or without names) etc), it's better to do that in user code, which is why this method has been rejected in the past. If we were going to support one format, it should be a simple array; PHPMailer is meant to be for assembling messages, not parsing user input.

torressam333:
******PHPMailer is meant to be for assembling messages, not parsing user
input.******

Well Said!

Issue #303, opened 2014-10-27
SMTP connect() failed while hosting on 000webhost.com


bshikhar13:
I am using a feedback form for my users on my website and currently hosting it on 000webhost.com Apache ver : 2.2.19 (Unix)
PHP version : 5.2

I am using google SMTP server (smtp.gmail.com) to send emails.
It shows SMTP connect() failed everytime.
Please respond.

PR #203, opened 2014-03-28
phpmailer.lang-mg

octavebenil:
Une traduction en Malgache

PR #1555, opened 2018-09-28
Create phpmailer.lang-mg.php

Hackinet:
Added translation for Malagasy.


Synchro:
Did you see the PR in #303? Yours is more complete, but more words seem to be translated in the other one - for example "SMTP error". I can't tell which is "better" as I don't speak Malagasy!

Hackinet:
@Synchro Sorry for the late reply. Yes, I saw the PR in #203 . While creating the new lang file I took into account, the old commit. The new one is definitely better.

Synchro:
Thanks. And yes I did mean #203!

PR #1577, opened 2018-10-16
Fix missing Arabic translations

MicroDroid:
You probably want to squash before merging. `extension_missing` was missing.


Synchro:
Thanks

PR #1589, opened 2018-10-29
add php 7.3 to travis

andreybolonin:


Synchro:
Thanks. 7.3 was effectively being tested in the `nightly` option anyway, but it's good to have it made explicit now it's available by name.

Issue #24, opened 2013-02-01
Missing translations


Synchro:
There are quite a few missing translations. Can anyone help complete them?

To get a list of wha needs fixing, run `test/phpmailerLangTest.php` with PHPUnit and it will give a list of all missing (and stray extra) translations.

Hassantravels:
yes
do

Hassantravels:
> There are quite a few missing translations. Can anyone help complete them?
> 
> To get a list of wha needs fixing, run `test/phpmailerLangTest.php` with PHPUnit and it will give a list of all missing (and stray extra) translations.fix

PR #1598, opened 2018-11-14
Add Serbian translation

mmilanovic4:
Hi, this commit adds Serbian translation in PHPMailer (issue #24).

Cheers.

### DIFF
--- a/language/phpmailer.lang-sr.php
+++ b/language/phpmailer.lang-sr.php
@@ -1,0 +1,26 @@
+<?php
+/**
+ * Serbian PHPMailer language file: refer to English translation for definitive list
+ * @package PHPMailer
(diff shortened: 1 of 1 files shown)


Synchro:
Thanks

mmilanovic4:
Hi, there is already Serbian translation under **rs** code (ISO 3166, codes for the names of the countries). I used ISO 639 (codes for the names of languages) **sr**. You can keep the old one, eventually apply the missing properties from mine.

Synchro:
Ah, good catch. I just looked at the mechanism for remapping incorrect language codes (in `lang()`) and I see that this was previously changed in the other direction! I will rename it correctly and update it in there.

PR #1610, opened 2018-11-28
Update PHPMailer.php

MESWEB:
Please don't make silent errors for php codes. This's not a solution. Never use this "@". Only noobs use this silencer for errors. Shame on U.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -1320,7 +1320,7 @@ class PHPMailer
         ) {
             $domain = substr($address, ++$pos);
             // Verify CharSet string is a valid one, and domain properly encoded in this CharSet.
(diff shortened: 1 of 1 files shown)


Synchro:
No. If you'd actually read the code instead of choosing to insult me, you'd have seen that the error status is retrieved correctly and returned to the calling function. The `@` operator is used to suppress a **duplicate** out-of-band warning that cannot otherwise be dealt with. If you want to deal with this a more involved way, the right way to do it is to instantiate a new error handler during this single call, something which is done elsewhere in the PHPMailer code, however, all that does is achieve *exactly* the same thing as using the `@` operator.

What's more important is that using the mail function **at all** is unsafe no matter what you do. It's both faster and safer to use SMTP to localhost, as the docs recommend.

PR #1646, opened 2019-01-25
ical method header

Paktusin:

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -146,6 +146,21 @@ class PHPMailer
     public $Ical = '';
 
     /**
+     *
+     * If MIME type equals 'text/calendar' it should allow method=REQUEST and method=CANCEL in the content-type header.
+     * filtering these and disallowing them to be injected into the header appears to prevent Outlook from processing the calendar invites and cancellations correctly.
+     *
+     * Outlook ICS file header for an invite request:
+     * Content-Type: text/calendar; charset="utf-8"; method=REQUEST
+     *
+     * Outlook ICS file header for an cancellation request:
+     * Content-Type: text/calendar; charset="utf-8"; method=CANCEL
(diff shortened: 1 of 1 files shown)


Synchro:
Not a bad idea - do you find that the iCal support in Outlook works at all?

It would be good to validate it so we can avoid header injection attacks - enforce that it is exactly one of `REQUEST` or `CANCEL` (and whichever other values are valid) before using it.

Also - please fix coding standards (see Travis output) and add test coverage please!

Paktusin:
I tested on Outlook. When Event must cancel and header method and ical file method are different, Outlook do not deleting event from calendar. If methods in header and in body the same (CANCEL) all work properly

Paktusin:
Yes. Its good Idea. I recode solution accroding  this rules:
https://tools.ietf.org/html/rfc5545#section-3.8.4.7

>      The "method" parameter is used to convey the iCalendar objec
>       method or transaction semantics for the calendaring and scheduling
>       information.  It also is an identifier for the restricted set of
>       properties and values of which the iCalendar object consists.  The
>       parameter is to be used as a guide for applications interpreting
>       the information contained within the body part.  It SHOULD NOT be
>       used to exclude or require particular pieces of information unless
>       the identified method definition specifically calls for this
>       behavior.  Unless specifically forbidden by a particular method
>       definition, a text/calendar content type can contain any set of
>       properties permitted by the Calendaring and Scheduling Core Object
>       Specification.  The "method" parameter MUST be specified and MUST
>       be set to the same value as the "METHOD" component property of the
>       iCalendar objects of the iCalendar stream if and only if the
>       iCalendar objects in the iCalendar stream all have a "METHOD"
>       component property set to the same value.

Synchro:
Why did you close this?

Paktusin:
Sorry I'm busy right now

PR #1663, opened 2019-02-14
corrected 'imap_getmailboxes' comment to include third parameter

olivermbs:
[**imap_getmailboxes expects a 3rd parameter:**](http://php.net/manual/en/function.imap-getmailboxes.php)

imap_getmailboxes ( resource $imap_stream , string $ref , **string $pattern** )


Synchro:
Thanks

PR #1699, opened 2019-03-31
Fixed current variable instance with double dollar sign

jukimv1986:


Synchro:
Sorry, this is not a bug, and your change breaks things. This syntax is making use of a PHP feature called variable variables, and is both valid and entirely intentional in this case.

PR #1719, opened 2019-05-05
Update composer license to `LGPL-2.1-only`

dregad:
As per SPDX License List, previously used `LGPL-2.1` is deprecated [[1]].

I'm not 100% sure, but assumed you indeed are under `LGPL-2.1-only` and not `LGPL-2.1-or-later`.

[1]: https://spdx.org/licenses/LGPL-2.1.html

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,5 +51,5 @@
             "PHPMailer\\Test\\": "test/"
         }
     },
-    "license": "LGPL-2.1"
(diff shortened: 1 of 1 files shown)


Synchro:
Thanks. There's really no scope for changing licenses - there's no way I can get permission from all contributors since 2001!

PR #1735, opened 2019-05-24
Gmail supports the standard IMAP, POP, and SMTP protocols. (Non-Gmail…

BryanKylan:
… Users)

Accessing mail using IMAP or POP and sending mail using SMTP is often done using existing IMAP and SMTP libraries for convenience. As long as these libraries support the Simple Authentication and Security Layer (SASL), they should be compatible with the SASL XOAUTH2 mechanism supported by Gmail.


Synchro:
I'm not sure why you wanted to say that - of course it's optional, but this is example code, so it makes use of lots of features which you can ignore/delete if you don't need them; but attaching things to messages is a very common request.

Issue #184, opened 2014-02-14
Can't disable X-Mailer header


bhall7:
I can't disable the X-Mailer header nor clear it. When I instantiate a new PHPMailer object , I even tried setting the XMailer header to `''`, but PHPMailer still includes this header in all emails sent.

```
$mail = new PHPMailer(true);
$mail->XMailer    = ''; 
```

Is there an easy way to disable this header field?

See also https://code.google.com/a/apache-extras.org/p/phpmailer/issues/detail?id=66

gaecom:
if ($this->XMailer == '') {
            $result .= $this->headerLine(
                'X-Mailer',
                'PHPMailer ' . $this->Version . ' (https://github.com/PHPMailer/PHPMailer)'
  );
you cant leave  $mail->XMailer empty,you should assign value.such as $this->XMailer='Microsoft Mailer';

Synchro:
@gaecom what? XMailer is never left empty, though it may be removed altogether, as all `X`-prefixed headers can.

rajeevacj:
Use `$mail->xmailer = '0'`

Zafpyr:
$mail->XMailer = '(;-_-) Nothing to see here';

PR #1764, opened 2019-06-27
Strict check XMailer with empty string

szemek:
Motivation:
https://github.com/PHPMailer/PHPMailer/issues/184
to disable X-Mailer header it is necessary to set `XMailer` to a whitespace.
```php
$mailer = new PHPMailer();
$mailer->XMailer = ' ';
```

Without breaking a backward compatibility too much, strict check with an empty string allows to use `null` in a similar manner to whitespace:

```php
$mailer = new PHPMailer();
$mailer->XMailer = null;
```

`trim(null)` return `''`, so code inside this condition won't get executed:

https://github.com/PHPMailer/PHPMailer/blob/511413a6eaeb167fa88282d437766d34510548b8/src/PHPMailer.php#L2390-L2392

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -527,9 +527,9 @@ class PHPMailer
 
     /**
      * What to put in the X-Mailer header.
-     * Options: An empty string for PHPMailer default, whitespace for none, or a string to use.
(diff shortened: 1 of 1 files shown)


Synchro:
Thanks.

PR #1776, opened 2019-07-11
addrFormat: Ensure RFC 2822 compliance for addresses without name.

apapsch:
Per https://tools.ietf.org/html/rfc2822#section-3.4, even a naked
address without display name needs to be enclosed in angle brackets.
At least spamassassin may not like naked addresses, e.g. if the rule
TO_NO_BRKTS_NORDNS_HTML is triggered.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -2133,7 +2133,7 @@ class PHPMailer
     public function addrFormat($addr)
     {
         if (empty($addr[1])) { // No name provided
-            return $this->secureHeader($addr[0]);
+            return '<' . $this->secureHeader($addr[0]) . '>';
         }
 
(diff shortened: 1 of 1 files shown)


Synchro:
The section you cite says:

> There is an alternate simple form of a mailbox where the addr-spec address appears alone, without the recipient's name or the angle brackets.

and that's what we're using. 2822 was obsoleted by 5322, but it didn't change this spec.

The spamassassin rule you mention would be triggered if you used this format, but only if you also sent an HTML-only message, and you did not have reverse DNS configured, which is a far more serious issue - in that configuration you'd be lucky to be able to send mail at all.

apapsch:
Ah well, I only looked at the BNF definition. Closing then.

> you did not have reverse DNS configured

It was a newly created server, which had a PTR for the IPv4 address, but the server connected using IPv6, for which a PTR did not exist. The woes of dual stack :-)

PR #1789, opened 2019-07-24
Update README.md - Use Clear* functions in example

dturner123:
Update the example script in the README.md file to include the following after $mail->send(); - $mail->ClearAddresses(); / $mail->ClearCCs(); / $mail->ClearBCCs();


Synchro:
No, this is pointless. There's no need to clear recipients unless you're re-using the instance (for example when [sending to a list](https://github.com/PHPMailer/PHPMailer/blob/master/examples/mailing_list.phps)), which this example script does not do.

PR #1794, opened 2019-07-26
PHP 7.4 Fix: Do not run get_magic_quotes_runtime anymore

Ayesh:
PHP 7.4 will start to throw deprecation notices when magic quote functions are use. A simple version check before calling them will fix this. 

I'd be grateful if you could review it. The PR tries to make the changes as minimally as possible.

Thaak you.

### DIFF
--- a/class.phpmailer.php
+++ b/class.phpmailer.php
@@ -2711,7 +2711,7 @@ class PHPMailer
             if (!self::isPermittedPath($path) or !file_exists($path)) {
                 throw new phpmailerException($this->lang('file_open') . $path, self::STOP_CONTINUE);
             }
-            $magic_quotes = get_magic_quotes_runtime();
+            $magic_quotes = PHP_VERSION_ID < 70400 && get_magic_quotes_runtime();
             if ($magic_quotes) {
                 if (version_compare(PHP_VERSION, '5.3.0', '<')) {
                     set_magic_quotes_runtime(false);


Synchro:
It's a bit odd that [the PHP docs don't mention this deprecation](https://www.php.net/manual/en/function.get-magic-quotes-runtime.php), though it is listed in [the 7.4 upgrade guide](https://github.com/php/php-src/blob/PHP-7.4/UPGRADING#L412).

That's all a bit academic though - `get_magic_quotes_runtime` only appears in the long-deprecated 5.2 branch. The _only_ reason for running PHPMailer 5.2 is if you are stuck on some old (< 5.5) version of PHP, so fixes for 7.4 make no sense - you can just use 6.x and you won't get the warnings because it doesn't contain the deprecated functions.

Ayesh:
Thanks for your reply @Synchro. We are working on WordPress's PHP 7.4 compatibility fixes, and WordPress currently bundles PHPMailer 5.2.17 (not the current 5.2.22 for some reason). I was thinking it would be a good idea to improve upstream rather than maintaining WordPress's own fork of it. 

Applying this patch shouldn't break existing installation, and I think this is purely a patch-level improvement rather than anything significant.

Synchro:
As far as I'm aware, the only reason that Wordpress has stuck with the 5.2 branch was because of the PHP version requirements of Wordpress itself - that has now [changed](https://wordpress.org/about/requirements/), and Wordpress now requires a minimum of PHP 5.6.20, which means that it can run PHPMailer 6.x without any issues. So a better solution than fixing 5.2 issues is to stop using 5.2 altogether and switch to 6.x. That should also mean you no longer need to maintain a fork for BC either.

Synchro:
Wordpress has started the process of switching to PHPMailer 6.x, so this change is no longer needed. See the Wordpress tracker for more info: https://core.trac.wordpress.org/ticket/41750

PR #1799, opened 2019-08-02
Change msgHtml to allow callback as image resolver

conrallendale:
Change the second parameter of the function msgHtml to accept a callable, beside the current string parameter. This way, I can implement my own way of getting the image data. In my case, using S3, I can request S3 for a object and return the image data, name and type. Eg:

```
$mail->msgHTML('<img src="my-image.jpg" />', function($url) use ($s3client) {
    $object = $s3client->getObject([ "Bucket" => "my-bucket", "Key" => $url ]);

    return [
        "name" => $url,
        "data" => (string) $object["Body"],
        "type" => $object["@metadata"]["headers"]["content-type"]
    ];
});
```


Synchro:
This all looks quite reasonable - could you add a test to cover this change please?

src/PHPMailer.php
```diff
@@ -3897,35 +3898,46 @@ public function msgHTML($message, $basedir = '', $advanced = false)
...
+                            $directory .= '/';
+                        }
+                        $added = $this->addEmbeddedImage(
+                            $imageResolver . $directory . $filename,
+                            $cid,
+                            $filename,
+                            static::ENCODING_BASE64,
+                            static::_mime_types((string) static::mb_pathinfo($filename, PATHINFO_EXTENSION))
+                        );
+                    } elseif (is_callable($imageResolver)) {
```
h00na, line 3928: If the passed param is not callable we can thrown an exception telling about it. Otherwise it will be hard to debug what exactly is happening

PR #1807, opened 2019-08-09
Update mail.phps

slykereven:
Modify $mail->AltBody by $mail->Body to resolve the error :
Mailer Error: Message body empty


Synchro:
No. I think you have misunderstood how `msgHTML()` works; what you have proposed doesn’t make sense in this context.

Issue #1824, opened 2019-09-05
Improve mail sending by reduce MIME encoding process


changyy:
# Problem description

When you want to send small changes mail content to a lot of receivers one by one, the same attachments would be MIME encoded repeatedly.

It would cost CPU time/credit on AWS, so I try to propose a solution.

# Code to reproduce

Look at sample code: https://github.com/PHPMailer/PHPMailer/blob/master/examples/smtp.phps

Use a while loop to reproduce situation:

```
foreach( [ /* .. */ ] as $receiver ) {
    //Create a new PHPMailer instance
    $mail = new PHPMailer;
    // ...
    $mail->addAddress($receiver);
    // ...
    $mail->Subject = 'Hi '.$receiver;
    // ...
    $mail->addAttachment('images/phpmailer_mini.png');
    // ...
}
```

# Debug output

None

# Solution:

Providing a MIME Encoding Cache Store to handle it.

Synchro:
I get what you're trying to do here, however, the implementation you wrote has some fairly major issues - it represents the only time that PHPMailer would attempt to write to disk, preventing it from working in read-only environments. It would be better to keep the cache in memory, or better, make it pluggable with [a PSR-6 instance](https://www.php-fig.org/psr/psr-6/).

That said, it's not clear whether this brings any significant performance improvement since instead of reading a file once, it would need to read it twice and write it once. Do you have benchmarks to back it up?

Your usage example in this ticket is already sub-optimal because it creates a new instance inside the loop, and there can be no benefit at all in the example file because it only sends one message - it also strikes me as unnecessary to add a complete example purely to add one line - if we're going to have a cache like this, it should probably be automatic and not require user code changes.

Overall, what I'd really like to see to improve attachment handling is to remove the need to store attachments in memory at all - have them encoded dynamically using a generator, which would be faster and _much_ more memory efficient. I suspect this would require a BC break and a PHP version bump, which I'm considering for future versions anyway.

changyy:
Hi Synchro,

Thanks for your review.

This solution could be running at Read-Only file system because using PHP array type to keep the MIME encoding result in memory.

```
//
// $this->ReduceMIMEEncodeCacheStore = [];
//
if (is_array($this->ReduceMIMEEncodeCacheStore)) {
	$cache_key = "$path\t$encoding";
	if (isset($this->ReduceMIMEEncodeCacheStore[$cache_key])) {
[...]

PR #1827, opened 2019-09-05
Try to help #1824

changyy:
This solution could be running at Read-Only file system because using PHP array type to keep the MIME encoding result in memory.

```
//
// $this->ReduceMIMEEncodeCacheStore = [];
//
if (is_array($this->ReduceMIMEEncodeCacheStore)) {
	$cache_key = "$path\t$encoding";
	if (isset($this->ReduceMIMEEncodeCacheStore[$cache_key])) {
		$file_buffer = $this->ReduceMIMEEncodeCacheStore[$cache_key];
		$this->edebug('encodeFile: use cache ['.$path.']['.$encoding.']');
	} else {
		$file_buffer = file_get_contents($path);
		if (false === $file_buffer) {
			throw new Exception($this->lang('file_open') . $path, self::STOP_CONTINUE);
		}
		$file_buffer = $this->encodeString($file_buffer, $encoding);
		$this->ReduceMIMEEncodeCacheStore[$cache_key] = $file_buffer;
		$this->edebug('encodeFile: set cache ['.$path.']['.$encoding.']');
	}
} 
```

Performance improvement:

```
$ php -v
PHP 7.1.23 (cli) (built: Feb 22 2019 22:19:32) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2018 Zend Technologies
$ git clone https://github.com/changyy/PHPMailer.git
$ cd PHPMailer
$ php ~/bin/composer.phar install --no-dev
$ cd examples
$ php examples/smtp_reduce_mime_encoding.phps 
$ php smtp_reduce_mime_encoding.phps 
Genrate 1000 receivers: 1000
Sending 1000 mails with creating a new PHPMailer instance: 0.41333603858948 sec
Sending 1000 mails with only one PHPMailer instance: 1.745913028717 sec
Sending 1000 mails with creating a new PHPMailer instance and ReduceMIMEEncodeCacheStore: 0.32201600074768 sec
Sending 1000 mails with only one PHPMailer instance and ReduceMIMEEncodeCacheStore: 1.6159191131592 sec
```

I think using an array key-value cache to reduce MIME encoding would nice to let CPU usage down.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,8 @@
     "require": {
         "php": ">=5.5.0",
         "ext-ctype": "*",
-        "ext-filter": "*"
+        "ext-filter": "*",
+        "psr/simple-cache": "^1.0"
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.2",
--- a/examples/benchmark_mime_cache.phps
+++ b/examples/benchmark_mime_cache.phps
@@ -18,7 +18,6 @@ require '../vendor/autoload.php';
 // $ curl https://sample-videos.com/img/Sample-jpg-image-200kb.jpg > /tmp/200kb.jpg
 // $ curl https://sample-videos.com/img/Sample-jpg-image-100kb.jpg > /tmp/100kb.jpg
 // $ curl https://sample-videos.com/img/Sample-jpg-image-50kb.jpg > /tmp/50kb.jpg
-//
 $image_files = array(
     '1.8kb' => 'phpmailer_mini.png',
     '5.7kb' => 'images/phpmailer.png',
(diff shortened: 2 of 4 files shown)


Synchro:
I think the payoff here is really pretty marginal - while the cache does help a small amount, it's completely dwarfed by the overhead of actually sending the message, so it can never amount to much of a gain overall. I'm actually much more curious about why reusing a single PHPMailer instance is slower (having a far greater impact than the cache), and I'd really like to profile that to find out why. It would also be interesting to compare memory consumption between these approaches too.

PR #1832, opened 2019-09-13
Update the translation for Persian/Farsi

MHM5000:
Include the last string (`extension_missing`) in Persian translation.


Synchro:
Thanks!

PR #1837, opened 2019-09-20
Fixed signing emails with pck7.

christiandavilakoobin:
When signing an email using a pck7 cert, it throws an error because openssl_pkcs7_sign is expecting a string, not a resource, failing the signature.


Synchro:
Thanks. Nice to know someone is using this code!

Issue #818, opened 2016-09-05
Trouble encoding subjects with isMail


boteha:
I'm experiencing trouble encoding mail subjects with isMail 

=?ISO-8859-15?Q?monsite.com_:_SUIVI_de_votre_devis_11, _proposition_?= d'échange 

This happens with isMail, not with isSMTP.

I've discovered that the simultaneous presence of a comma and an accented letter triggers the problem.

In other words the following subjects are treated well : 
/ / only commas
cordon RJ45 rouge, 10 m, LSZH 

/ / only apostrophes or accented letters
cordon RJ45 d'extérieur sans halogène 

/ / BUT if containing a comma and apostrophe OR accented letter, the following happens : 
=?ISO-8859-15?Q?monsite.com_:_SUIVI_de_votre_devis_11, _proposition_?= d'échange 

Is this a PHPMailer bug or a server bug ?

Synchro:
Is this in outlook or outlook.com? Is this how they are displayed, or how they appear when you view message source?

boteha:
Not tested with Outloook, only with Thunderbird under Win 10.

In the list of mails the display is correct :
monsite.com : :SUIVI de votre devis 11, proposition d'échange 

But the problem appears in the smaller header you can choose.

Il mean you see :
From : sender@sender.com
Subject : =?ISO-8859-15?Q?monsite.com_:SUIVI_de_votre_devis_11, _proposition?= d'échange 
To : receiver@receiver.com

Of course in the page source it is even worst :
Subject: =?ISO-8859-1?Q?monsite=2Ecom_=3A_SUIVI_de_votre_commande_4?=
 =?ISO-8859-1?Q?0379=2C_proposition_d=27=E9change?=

But this normal, no problem.

The problem is the header that you see under Thunderbird.

Is it clear ?

Synchro:
It's what it looks like in the source that's important! The bit you posted isn't exactly what I'd expect to see - there should be white space at the start of the second line because it's a folded header. Is it like that, or did you mispaste?

Can you post exact examples of working and non-working headers from the message source please?

There's one important difference - the source says `ISO-8859-1` (PHPMailer's default), but your decoded version says `ISO-8859-15` - that must be something that your client has done. In PHPMailer we only have control over the source, not how a client renders it. Those 2 charsets are very similar, but -15 includes the € symbol.

boteha:
Hello,

[...]

Issue #36, opened 2013-04-04
Header lines can be too long


Synchro:
The EncodeHeader function has two problems:
1. If a header needs Q or B encoding, it also gets folding applied, but if a line has no characters that need encoding, but is still too long, it will not get folded.
2. When a header does get folded, it only takes into account the length of the value, not the key, and the combination of the two can make the line too long.

This makes me think that the folding is applied at the wrong point. Header values should be encoded and returned, and folded later after combining with their keys. That said, I don't know if Q and B encoded headers really need their prefixes applied on every line, which is what they are doing at present.

Update: Q and B encodings _do_ need their prefixes repeating on each line because an RFC 2047 'encoded-word' may not contain white space, and thus cannot be folded.

maartenmachiels:
Hi, is there any update on this issue? Using PHPMailer in combination with Sendgrid via (SMTPAPI) and am encountering lots of issues, as the recipients are stored in a custom header. After 72 characters, newlines are inserted, which results in broken e-mail addresses...

Synchro:
Line breaks in headers are not a problem because they are folded, but that's not what this issue is about - it's about headers being allowed to get too long *before* they are folded.

Can you open a new ticket and be more specific about your problem, provide examples etc?

maartenmachiels:
Thanks so much for your follow-up. I've posted my issue here: https://github.com/PHPMailer/PHPMailer/issues/726

It's a long one. Sorry in advance :-)

PR #1840, opened 2019-09-23
Use Q-encode to wrap too long headers

### LINKED ISSUE #1469: Long filenames in "Content-Disposition: attachment; filename" breaks extension
**- Problem description**
I've been getting several complaints by customers about them receiving mail with attachments which they cannot open because of some malformed file extension.

It turns out the filenames are folded after 76 characters. This SEEMS RFC-compliant.
The extra spacing character is interpreted as a space and, of course, doesn't cause problems when not in the extension.

When sending the same, original, file through Outlook (client or web), GMail or any other hosted service, the file arrives with the complete and untouched extension.

**- Code/Steps to reproduce**
Send an email and include a file with a filename just over 76 characters long.

```
$filename = "WayTooLongFileNameWhichWillCorruptTheFileExtensionAtCharacterSeventySeven.pdf"; // filename is 77 characters
$mailer->AddAttachment($att->file, $filename, $att->encoding, $att->mimeType);
```

**- Debug Output**
```
--b1_6IjFX38R88zg8cQbwrrdGs9XKE4P6sRVpqGdadM00Is
Content-Type: application/pdf; name="WayTooLongFileNameWhichWillCorruptTheFileExtensionAtCharacterSeventySeven.pd f"
Content-Transfer-Encoding: base64
Content-ID: <WayTooLongFileNameWhichWillCorruptTheFileExtensionAtCharacterSeventySeven.pdf>
Content-Disposition: attachment; filename="WayTooLongFileNameWhichWillCorruptTheFileExtensionAtCharacterSeventySeven.pd f"
```

**- Possible solution**
As said, the current implementation SEEMS RFC-compliant. But [RFC 2184 Section 3](https://tools.ietf.org/html/rfc2184#section-3) speaks about Parameter Value Continuation and ways to break too long parameter values into smaller chunks without folding and altering the original values.

> RFC 2184 Section 3
>    ...
>    
>    The obvious solution, then, is to use multiple parameters to contain
>    a single parameter value and to use some kind of distinguished name
>    to indicate when this is being done.  And this obvious solution is
>    exactly what is specified here: The asterisk character ("*") followed
>    by a decimal count is employed to indicate that multiple parameters
>    are being used to encapsulate a single parameter value.  The count
>    starts at 0 and increments by 1 for each subsequent section of the
>    parameter value.  Decimal values are used and neither leading zeroes
>    nor gaps in the sequence are allowed.
> 
>    The original parameter value is recovered by concatenating the
>    various sections of the parameter, in order.  For example, the
>    content-type field
> 
>      Content-Type: message/external-body; access-type=URL;
>       URL*0="ftp://";
>       URL*1="cs.utk.edu/pub/moore/bulk-mailer/bulk-mailer.tar"
> 
>    is semantically identical to
> 
>      Content-Type: message/external-body; access-type=URL;
>       URL="ftp://cs.utk.edu/pub/moore/bulk-mailer/bulk-mailer.tar"

Shortening filenames is not an option because:
- clients include specific indentifiers into the filename which can differ in length
- regular email clients are able to send files with names exceeding 76 characters in length without folding/altering

**- System-info**
PHPMailer 6.0.5 (Composer)
PHP 7.1
CentOS 6.9/7.4

Can you look into this?
Thank you!

### LINKED ISSUE #1525: Long header folding adds additional spaces
If sending a long header without spaces, header folding is incorrectly.

Headers should fold with a CRLF, but after folding the CRLF is replaced with a LF and a space is introduced: 

    $encoded = trim(chunk_split($str, static::STD_LINE_LENGTH, static::$LE));

This is correct; but 

    $encoded = str_replace(static::$LE, "\n", trim($encoded));
    $encoded = preg_replace('/^(.*)$/m', ' \\1', $encoded);

corrupts the folding; order should be different.

caugner:
Fixes #1469.
Fixes #1525.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -3154,8 +3154,8 @@ class PHPMailer
                 $encoded = trim(chunk_split($encoded, $maxlen, "\n"));
             }
             $encoded = preg_replace('/^(.*)$/m', ' =?' . $this->CharSet . "?$encoding?\\1?=", $encoded);
-        } elseif ($matchcount > 0) {
-            //1 or more chars need encoding, use Q-encode
+        } elseif ($matchcount > 0 || strlen($str) > $maxlen) {
+            // 1 or more chars need encoding or header exceeds max line length, use Q-encode
             $encoding = 'Q';
             //Recalc max line length for Q encoding - see comments on B encode
             $maxlen = static::STD_LINE_LENGTH - $lengthsub - 8 - strlen($this->CharSet);
@@ -3163,15 +3163,6 @@ class PHPMailer
             $encoded = $this->wrapText($encoded, $maxlen, true);
             $encoded = str_replace('=' . static::$LE, "\n", trim($encoded));
             $encoded = preg_replace('/^(.*)$/m', ' =?' . $this->CharSet . "?$encoding?\\1?=", $encoded);
-        } elseif (strlen($str) > $maxlen) {
-            //No chars need encoding, but line is too long, so fold it
-            $encoded = trim($this->wrapText($str, $maxlen, false));
-            if ($str == $encoded) {
-                //Wrapping nicely didn't work, wrap hard instead
-                $encoded = trim(chunk_split($str, static::STD_LINE_LENGTH, static::$LE));
-            }
-            $encoded = str_replace(static::$LE, "\n", trim($encoded));
-            $encoded = preg_replace('/^(.*)$/m', ' \\1', $encoded);
         } else {
             //No reformatting needed
             return $str;
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -746,6 +746,16 @@ class PHPMailer
     protected static $LE = "\r\n";
 
     /**
+     * The maximum line length supported by mail().
+     *
+     * Background: mail() will sometimes corrupt messages
+     * with headers headers longer than 65 chars, see #818.
+     *
+     * @var int
+     */
+    const MAIL_MAX_LINE_LENGTH = 63;
+
+    /**
      * The maximum line length allowed by RFC 2822 section 2.1.1.
      *
      * @var int
@@ -3131,15 +3141,19 @@ class PHPMailer
                 break;
         }
 
-        //RFCs specify a maximum line length of 78 chars, however mail() will sometimes
-        //corrupt messages with headers longer than 65 chars. See #818
-        $lengthsub = 'mail' == $this->Mailer ? 13 : 0;
-        $maxlen = static::STD_LINE_LENGTH - $lengthsub;
-
         if ($this->has8bitChars($str)) {
             $charset = $this->CharSet;
         } else {
             $charset = static::CHARSET_ASCII;
+        }
+
+        // Q/B encoding adds 8 chars and the charset ("` =?<charset>?[QB]?<content>?=`").
+        $overhead = 8 + strlen($charset);
+
+        if ('mail' == $this->Mailer) {
+            $maxlen = static::MAIL_MAX_LINE_LENGTH - $overhead;
+        } else {
+            $maxlen = static::STD_LINE_LENGTH - $overhead;
         }
 
         // Select the encoding that produces the shortest output and/or prevents corruption.
@@ -3159,12 +3173,6 @@ class PHPMailer
 
         switch ($encoding) {
             case 'B':
-                //This calculation is:
-                // max line length
-                // - shorten to avoid mail() corruption
-                // - Q/B encoding char overhead ("` =?<charset>?[QB]?<content>?=`")
-                // - charset name length
-                $maxlen = static::STD_LINE_LENGTH - $lengthsub - 8 - strlen($charset);
                 if ($this->hasMultiBytes($str)) {
                     // Use a custom function which correctly encodes and wraps long
                     // multibyte strings without breaking lines within a character
@@ -3177,8 +3185,6 @@ class PHPMailer
                 $encoded = preg_replace('/^(.*)$/m', ' =?' . $charset . "?$encoding?\\1?=", $encoded);
                 break;
             case 'Q':
-                //Recalc max line length for Q encoding - see comments on B encode
-                $maxlen = static::STD_LINE_LENGTH - $lengthsub - 8 - strlen($charset);
                 $encoded = $this->encodeQ($str, $position);
                 $encoded = $this->wrapText($encoded, $maxlen, true);
                 $encoded = str_replace('=' . static::$LE, "\n", trim($encoded));
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -3136,6 +3136,12 @@ class PHPMailer
         $lengthsub = 'mail' == $this->Mailer ? 13 : 0;
         $maxlen = static::STD_LINE_LENGTH - $lengthsub;
 
+        if ($this->has8bitChars($str)) {
+            $charset = $this->CharSet;
+        } else {
+            $charset = static::CHARSET_ASCII;
+        }
+
         // Select the encoding that produces the shortest output and/or prevents corruption.
         if ($matchcount > strlen($str) / 3) {
             // More than 1/3 of the content needs encoding, use B-encode.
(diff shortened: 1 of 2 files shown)


caugner:
There is [one test case](https://travis-ci.org/PHPMailer/PHPMailer/jobs/588482898#L331) that expects the previous behaviour:

grafik

caugner:
As for the test case, this is fixed in 8701f3b.

Regarding the charset selection, isn't this a separate issue of `PHPMailer::encodeHeader`? Currently it always Q-encodes any value using the `PHPMailer::$CharSet`, whereas it could be more clever and prefer less expensive charsets such as `us-ascii` or `7bit` if applicable.

Synchro:
Yes, exactly that (though 7bit is a CTE, not a charset).

src/PHPMailer.php
```diff
@@ -3163,15 +3163,6 @@ public function encodeHeader($str, $position = 'text')
             $encoded = $this->wrapText($encoded, $maxlen, true);
             $encoded = str_replace('=' . static::$LE, "\n", trim($encoded));
             $encoded = preg_replace('/^(.*)$/m', ' =?' . $this->CharSet . "?$encoding?\\1?=", $encoded);
```
caugner: @Synchro Can we simply use `us-ascii` here instead of `$this->CharSet` if `!$this->has8bitChars($encoded)`?

Synchro: I think so - the only hole would be if it uses some unusual 7-bit charset that was not ascii compatible.

Synchro: It would be cool to add a test to create headers in both formats.

caugner:
> UTF-8 is excessive for something that doesn't contain 8-bit chars

Can you illustrate why UTF-8 is more excessive in this case:

```
=?UTF-8?Q?eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee?=
  =?UTF-8?Q?eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee?=

vs.

=?us-ascii?Q?eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee?=
  =?us-ascii?Q?eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee?=
```

Isn't `us-ascii` a subset of `UTF-8`?

Synchro:
While the content is indeed identical and compatible, marking it as UTF-8 requires firing up a complete unicode engine to decode it on the receiver, which is unnecessary overhead if we know in advance it's plain ASCII. In PHP for example, we would not be able to decode it easily if we didn't have the `mbstring` extension enabled, whereas it would present no problems in us-ascii.

 It's like putting literal IPs first in SPF records; it's not a requirement, but it makes life easier for receivers.

src/PHPMailer.php
caugner: > I think so - the only hole would be if it uses some unusual 7-bit charset that was not ascii compatible.

Do we need to handle this case somehow, if we can handle it all?

> It would be cool to add a test to create headers in both formats.

Added in 9e1f8f1 .

Synchro: Great stuff, thanks. Anyone that's using a non-ascii-compatible charset probably has bigger problems, so I don't think we need to worry about it...

caugner:
Unfortunately, CI has [some unrelated problems](https://travis-ci.org/PHPMailer/PHPMailer/jobs/588540025#L250):

grafik

src/PHPMailer.php
```diff
@@ -3154,24 +3155,20 @@ public function encodeHeader($str, $position = 'text')
                 $encoded = trim(chunk_split($encoded, $maxlen, "\n"));
             }
             $encoded = preg_replace('/^(.*)$/m', ' =?' . $this->CharSet . "?$encoding?\\1?=", $encoded);
-        } elseif ($matchcount > 0) {
+        } elseif ($matchcount > 0 || strlen($str) > $maxlen) {
             //1 or more chars need encoding, use Q-encode
```
caugner: TODO Update comment

caugner:
> I wonder if this will fix #1525 as well?!

I think so.

Maybe you could have a ~~second~~ third look? I have now refactored my PR as follows:

* Aligned commit messages with the usual style in this repository (consistency).
* Separated encoding selection from encoding (readability, separation of concerns).
* Applied charset selection to B-encoding as well (consistency).

src/PHPMailer.php
```diff
@@ -3134,47 +3135,57 @@ public function encodeHeader($str, $position = 'text')
         //corrupt messages with headers longer than 65 chars. See #818
         $lengthsub = 'mail' == $this->Mailer ? 13 : 0;
         $maxlen = static::STD_LINE_LENGTH - $lengthsub;
```
Synchro: This assumption may need looking at too - This came up in #1525 - it's OK to have a line longer than 76 chars if it contains no embedded spaces that it can be folded on - we should only force RFC2047 encoding if a line is over 998.
I think we need another check to see if the line has no 8-bit chars and contains an unspaced sequence longer than `MAX_LINE_LENGTH` chars (I guess something like `/\w{MAX_LINE_LENGTH,}/`), then we should force Q-encoding, which will wrap the whole header to `STD_LINE_LENGTH`. This isn't absolutely needed as there's no significant harm in wrapping earlier other than a mild increase in size, but it would be a "nice to have".

There's another bug related to this in #36, in that the header label is not taken into account in the length - that's not been critical since it only becomes a problem at the 998 char limit, but we should perhaps allow for that by wrapping a bit before 998, in a similar way that we try to avoid folding bugs in `mail()`.

So I think here we might rename `$maxlen` as `$foldlen` and set `$maxlen` to `static::MAX_LINE_LENGTH - $lengthsub`, then use them in appropriate places. This will affect test outcomes.

Everything else in here is great!

caugner: > This assumption may need looking at too

Actually, I just extracted this from the two branches (B and Q) for the sake of deduplication. Or did I miss anything?

> There's another bug related to this in #36, in that the header label is not taken into account in the length 

You're right that #36 (which I [did notice](https://github.com/PHPMailer/PHPMailer/issues/1469#issuecomment-534108894)) could be fixed as part of this PR, but in general I would opt against also fixing that one in this PR (which already does more than just *Us[ing] Q-encode to wrap too long headers*).

I'm not sure fixing #36 is as trivial since it would require touching several places where the encoded header *value* is just appended to the header name (e.g. `Content-Type`, `Content-Disposition`, etc.), like here:

https://github.com/PHPMailer/PHPMailer/blob/3c1c6e03603364ab9ba774b062ca4a5270b90adb/src/PHPMailer.php#L2941-L2954

Or am I missing anything?

Synchro: Indeed – we don't need to *fix* #36 in this PR, but a fudge here will certainly help. If we force Q+folding at say `MAX_LINE_LENGTH - 50` (which isn't quite RFC, but is harmless), to have a problem related to it you would need to have both a) a header label longer than 48 chars and b) a header value > 998 chars with no spaces, which is a very unlikely combo, sufficiently so that we could probably forget about #36.

Synchro: I'm having second thoughts about my last comment. We don't need to worry about lines between 76 and 998 chars long in an SMTP context, but in PHP and sending via `mail()`, we do, because of the bug noted mentioned the source, and on here in #818, where lines longer than 65 chars cause issues. Because of that we may as well not bother allowing long lines to exist and leave the encoding behaviour as it stands in your PR.

caugner: I committed an intermediate state (c4b7f6a) where I distinguished `$foldlen` and `$maxlen`, but I reverted this to f71240d, deduplicating the `$maxlen` calculation as originally intended (and hopefully making it clearer).

Synchro:
This was a great bit of work, solving several long-standing bugs that I'd been procrastinating about! Thank you very much for all your effort. This set of changes and other things that have happened lately mean that I will release this as PHPMailer 6.1 soon. Would you be interested in joining the PHPMailer org?

caugner:
@Synchro Thank you for your reviews and the merge. And sure, feel free to add me to the org. I might like to tackle the header name vs. header value line length limitation at some point.

Issue #1780, opened 2019-07-15
Cancel ICS file corrupted on Office 365


bashirmurtaza93:
There seems to be a recent issue with Office 365 when sending out cancellation Ical emails.
The main issue is that within the core code, when Ical is set, it's setting the method header to REQUEST. However, if you're sending a cancellation, you would need to set the method to CANCEL instead. 

Is there anyway within PHPMailer to set this?

To be clear i'm referring to this snippet 
```PHP
   case 'alt':
   $body .= $mimepre;
   $body .= $this->getBoundary($this->boundary[1], $altBodyCharSet, 'text/plain', $altBodyEncoding);
   $body .= $this->encodeString($this->AltBody, $altBodyEncoding);
   $body .= $this->LE . $this->LE;
   $body .= $this->getBoundary($this->boundary[1], $bodyCharSet, 'text/html', $bodyEncoding);
   $body .= $this->encodeString($this->Body, $bodyEncoding);
   $body .= $this->LE . $this->LE;
   if (!empty($this->Ical)) {
      $body .= $this->getBoundary($this->boundary[1], '', 'text/calendar; method=REQUEST, ''); //NEED TO CHANGE THIS TO CANCEL
      $body .= $this->encodeString($this->Ical, $this->Encoding);
      $body .= $this->LE . $this->LE;
      
     }
    $body .= $this->endBoundary($this->boundary[1]);
    break;
```

Synchro:
Not at the moment, no. I guess the best way to handle this would be to inspect the contents of the `Ical` property and figure out what method it should be using. Ical is generally very difficult to support because pretty much every client does not conform with RFCs (for example, Outlook should not be relying on this header at all - the status value is visible inside the Ical data), and their inconsistency makes it mostly impossible to create an ical email that works everywhere.

bashirmurtaza93:
Thanks for the quick response! I'll investigate further, and if I come up with a workaround to solve this issue, i'll post it here. Thanks!

GeorgeWells:
I hit this problem and found a way to get it working. Use addStringAttachment to attach the ical and specify the method in 4th parameter i.e.

`$mail->addStringAttachment($ical, 'ical.ics', 'base64', 'text/calendar; method=CANCEL');`

Synchro:
[...]

PR #1842, opened 2019-09-24
Cancel ICS file corrupted on Office 365 #1780

puhr-mde:
Since Office 365 with Outlook for web respects the Contenttype-header attribute für ICalendar-Method there is a need to set the property  suitable to your ICal-Method. ```$this->Encoding``` can be used to change the method value (Default = REQUEST).

In order to prevent literal errors the propery can be set by contants matching to the RFC.

    ```const ICAL_METHOD_REQUEST = "REQUEST";
    const ICAL_METHOD_PUBLISH = "PUBLISH";
    const ICAL_METHOD_REPLY = "REPLY";
    const ICAL_METHOD_ADD = "ADD";
    const ICAL_METHOD_CANCEL = "CANCEL";
    const ICAL_METHOD_REFRESH = "REFRESH";
    const ICAL_METHOD_COUNTER = "COUNTER";
    const ICAL_METHOD_DECLINECOUNTER = "DECLINECOUNTER";```

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -162,7 +162,7 @@ class PHPMailer
      *
      * @var string[]
      */
-    private static $IcalMethod = [
+    protected static $IcalMethods = [
         self::ICAL_METHOD_REQUEST,
         self::ICAL_METHOD_PUBLISH,
         self::ICAL_METHOD_REPLY,
@@ -170,7 +170,7 @@ class PHPMailer
         self::ICAL_METHOD_CANCEL,
         self::ICAL_METHOD_REFRESH,
         self::ICAL_METHOD_COUNTER,
-        self::ICAL_METHOD_DECLINECOUNTER
+        self::ICAL_METHOD_DECLINECOUNTER,
     ];
 
     /**
@@ -2617,13 +2617,13 @@ class PHPMailer
                 $body .= static::$LE;
                 if (!empty($this->Ical)) {
                     $method = static::ICAL_METHOD_REQUEST;
-                    foreach (static::$IcalMethod as $imethod) {
-                        if (stripos($this->Ical, 'METHOD:'.$imethod) !== false) {
+                    foreach (static::$IcalMethods as $imethod) {
+                        if (stripos($this->Ical, 'METHOD:' . $imethod) !== false) {
                             $method = $imethod;
                             break;
                         }
                     }
-                    $body .= $this->getBoundary($this->boundary[1], '', static::CONTENT_TYPE_TEXT_CALENDAR . '; method='.$method, '');
+                    $body .= $this->getBoundary($this->boundary[1], '', static::CONTENT_TYPE_TEXT_CALENDAR . '; method=' . $method, '');
                     $body .= $this->encodeString($this->Ical, $this->Encoding);
                     $body .= static::$LE;
                 }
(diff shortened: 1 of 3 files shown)


src/PHPMailer.php
```diff
@@ -148,6 +157,13 @@ class PHPMailer
      */
     public $Ical = '';
 
+    /**
+     * Value of "method" in Contenttype header "text/calendar"
+     *
+     * @var string
+     */
+    public $IcalMethod = self::ICAL_METHOD_REQUEST;
```
Synchro: I think it would be safer and simpler to skip this property and extract the method from the `Ical` property instead, using the permitted values declared in the constants. Otherwise you're leaving it open to the user setting a method that does not match the content.

Synchro: Also all references to these constants should be `static` rather than `self` to allow overriding.

puhr-mde: If you say skippen, do you mean remove it from the Header or ignoring it in processing the icalendar?

src/PHPMailer.php
```diff
@@ -158,11 +158,20 @@ class PHPMailer
...
 
     /**
-     * Value of "method" in Contenttype header "text/calendar"
+     * Value-array of "method" in Contenttype header "text/calendar"
      *
-     * @var string
+     * @var string[]
      */
-    public $IcalMethod = self::ICAL_METHOD_REQUEST;
+    private static $IcalMethod = [
```
Synchro: I was wondering why you hadn't used a constant array here - then I checked and found that you couldn't do that prior to PHP 5.6, and we are meant to retain compatibility back to PHP 5.5 in the 6.x branch, so that will have to wait. That said, this should be `protected` rather rather than `private` so that users are free to override it to add or remove Ical methods. I'd also call it `IcalMethods` since it's an array, not a single value.

puhr-mde: Yes, the constant thing will not work by now. I also changed it to case-insensitive. I worked with many groupwaresystems, i think it was Zarafa or Groupwise, in some of these ich came up with all lowercase icalendars. I will change it to protected, good idea. The thing with IcalMethods came to me too, just after commiting, sorry for that one.

puhr-mde:
I've added 3 testcases to check correct method an use default method in case of missing or invalid method attribute. 

> See my other comment

What do you mean with this? Am I still missing something?

test/PHPMailerTest.php
```diff
@@ -2675,6 +2675,131 @@ public function testOAuth()
...
+    /**
+     * Test ICal method
+     */
+    public function testICalMethod()
+    {
+        $this->Mail->Subject .= ': ICal method';
+        $this->Mail->Body = '<h3>ICal method test.</h3>';
+        $this->Mail->AltBody = 'ICal method test.';
+        $this->Mail->Ical = 'BEGIN:VCALENDAR'
+            . '\r\nVERSION:2.0'
```
Synchro: These `\r\n` sequences won't work because you're using single quotes, and those chars won't get unescaped; changing to double quotes will fix that.

test/PHPMailerTest.php
```diff
@@ -2675,6 +2675,131 @@ public function testOAuth()
...
+            . '\r\nORGANIZER;CN="Example, Test":MAILTO:test@example.com'
+            . '\r\nDTSTAMP:20190925T075546Z'
+            . '\r\nCREATED:20190925T075709Z'
+            . '\r\nLAST-MODIFIED:20190925T075546Z'
+            . '\r\nEND:VEVENT'
+            . '\r\nEND:VCALENDAR';
+        $this->buildBody();
+        $this->Mail->preSend();
+        $this->assertRegExp(
+            '/Content-Type: text\/calendar; method=CANCEL; charset=utf-8/',
```
Synchro: The default charset is `iso-8859-1`, so this won't match. Either change the match, or add `$this->Mail->CharSet = PHPMailer::CHARSET_UTF8;` to the test script.

You'll need to change the other instances of these problems too.

Synchro:
It's failing the coding standard check because you double quoted some strings that didn't need it.

puhr-mde:
> 
> 
> It's failing the coding standard check because you double quoted some strings that didn't need it.

Sorry, I thought its not wanted to mix quotes in one concatenation of a string. It should now match the cs.

Synchro:
Thank you very much!

puhr-mde:
Thank you. This was my first contribution/pull to a project on Github, thanks for your patience with all my bugs or conflicts with your cs.

Synchro:
Congratulations, and I'm honoured that you chose to do that in this project!
Coding standards can be annoying sometimes, but if you use an IDE that supports them (like PHPStorm), it will flag everything before you even commit. Sticking to a predefined CS scheme also removes the need to think about it!

PR #1844, opened 2019-09-24
add ssl smtp check to examples PHPMailer#1843

Marat555:
this fixes PHPMailer#1843


Synchro:
Sorry, adding this whole example to illustrate a 1-line change isn't really worth it.

Issue #1846, opened 2019-09-26
Travis-CI runs an unnecessary job


Synchro:
The [`.travis.yml` file](https://github.com/PHPMailer/PHPMailer/blob/master/.travis.yml) defines 8 jobs in the `test` stage:
```yaml
    - stage: test
    - php: 5.6
    - php: 7.0
    - php: 7.1
    - php: 7.2
    - php: 7.3
      env: CODE_COVERAGE=1
    - php: 7.4snapshot
    - php: nightly
    - php: hhvm
```
But when travis runs [a build](https://travis-ci.org/PHPMailer/PHPMailer/builds/589799513), it runs 9 jobs in this stage:

![image](https://user-images.githubusercontent.com/81561/65669545-0e127780-e044-11e9-9447-f8b5522fd1c4.png)

The extra one is the first one, called `PHP`. It uses a default PHP env that is currently PHP 7.2, which we are already including explicitly.

Why is this extra job being run? Is it an effect of the first line in the config which says `language: php`? How can it be prevented?

JalanJiang:
Hi, Synchro!

I think that is because you didn't choose PHP version in `.travis.yml`, so it used a PHP default version.

You should modify `.travis.yml` like this:

1. First, choosing PHP version:

```yml
language: php
php: '5.6'
```

2. Secondly, delete the PHP version which you choose:

```yml
jobs:
  include:
    - stage: coding-standard
      before_install:
      before_script:
      script: ./vendor/bin/php-cs-fixer --diff --dry-run --verbose fix
      after_script:
      php: 5.6
      env: CS_CHECK=1
    - stage: test
    # - php: 5.6 (Delete this line)
    - php: 7.0
    - php: 7.1
    - php: 7.2
    - php: 7.3
      env: CODE_COVERAGE=1
    - php: 7.4snapshot
    - php: nightly
    - php: hhvm
      dist: trusty
```

It runs 8 jobs in this stage:

![travis-ci](https://user-images.githubusercontent.com/7787716/65781840-f986ca00-e17e-11e9-8535-fecf5ffa5a99.png)

I create a pull request: https://github.com/PHPMailer/PHPMailer/pull/1848

Synchro:
Fixed in #1848, thanks!

PR #1848, opened 2019-09-27
Fix the problem about travis-ci.

JalanJiang:


JalanJiang:
About this issue: https://github.com/PHPMailer/PHPMailer/issues/1846

Synchro:
Excellent, that's exactly what I was looking for, thank you!

Issue #1484, opened 2018-06-27
Add an AJAX/Javascript example


Synchro:
We often get questions asked on Stack Overflow about sending mail and returning the response as JSON to a javascript function, including meaningful errors. For the experienced, it's easy to see how to separate these things into its separate parts, but it's hard for beginners, so we could do with an example they can use as a reference.

I'd recommend starting with one of the existing examples (such as the gmail one), converting it to produce JSON responses, and then write some Javascript to go with it.

praneethkatragadda:
Are you asking to provide some kind of confirmation whether mail is delivered or not ? 
I am not able to understand the requirement. It would be great you post those stack overflow questions  where users posted their queries. Please elaborate if possible.

Synchro:
No, that's a far more complex thing. What I'm suggesting is to provide an example which shows a request originating from a Javascript Ajax request (with and/or without jQuery), that is handled by PHP and uses PHPMailer to send a message using content supplied in the request, and then returns appropriate data (or an error) that can be handled correctly in JS.

praneethkatragadda:
I got your point. It's clear now.  I will do it. Thanks for eloborating and for quick response.

Synchro:
An example of [this kind of question](https://stackoverflow.com/questions/51583281/novice-developer-trouble-with-php-contact-form-templates-probably-something) popped up today.

its-tas:
https://gist.github.com/its-tas/7190b274f12a750901c68b7bac19459c

PR #1868, opened 2019-10-18
Add an AJAX/Javascript example #1484

udhaykumarbala:

### DIFF
--- a/examples/mail_through_ajax_js.html
+++ b/examples/mail_through_ajax_js.html
@@ -1,0 +1,39 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Contact form</title>
+    <script
+			  src="https://code.jquery.com/jquery-3.4.1.min.js"
+			  integrity="sha256-CSXorXvZcTkaix6Yvo6HppcZGetbYMGWSFlBw8HfCJo="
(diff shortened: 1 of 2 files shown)


Synchro:
Thanks for the PR. The JS side of this is fine, but the PHP side could probably use one of the existing examples since it's essentially the same. The only other thing I'd note is that there is no error response handling since the PHP script doesn't return anything particular to an ajax request, so you're left with parsing the text of the response rather than, for example, returning a 400 code along with a status message in JSON.

shikkaba:
Is this going to be added to the main branch ever? Searching to find the example is a slight inconvenience. Slight. Very slight.

Synchro:
@shikkaba You could always have a go at fixing the PR to address the issues I raised on it. I don't want to merge code that doesn't set a good example, which is after all the entire point of these examples!

shikkaba:
@Synchro Maybe! That's a good idea. I'll see what I can do. 👍

Synchro:
I know I said before that the JS in this was OK, but I'd really prefer to see an example without jQuery, i.e. using fetch.

Synchro:
Closed in favour of another PR which provided a good implementation, see https://github.com/PHPMailer/PHPMailer/blob/master/examples/contactform-ajax.phps

PR #1895, opened 2019-11-15
FIX int type

kiros32:
Hello!

This part `0 !== $max`
of expression `if (0 !== $max && $timeout > $max) {`
will always be true

Because ini_get returns string in `$max = ini_get('max_execution_time');`

Please merge my fix to correct use of strict comparison operator.

### DIFF
--- a/src/SMTP.php
+++ b/src/SMTP.php
@@ -372,7 +372,7 @@ class SMTP
         // SMTP server can take longer to respond, give longer timeout for first read
         // Windows does not have support for this timeout function
         if (strpos(PHP_OS, 'WIN') !== 0) {
-            $max = (int)(ini_get('max_execution_time'));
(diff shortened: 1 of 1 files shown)


src/SMTP.php
```diff
@@ -372,7 +372,7 @@ public function connect($host, $port = null, $timeout = 30, $options = [])
         // SMTP server can take longer to respond, give longer timeout for first read
         // Windows does not have support for this timeout function
         if (strpos(PHP_OS, 'WIN') !== 0) {
-            $max = (int)(ini_get('max_execution_time'));
+            $max = (int) (ini_get('max_execution_time'));
```
Synchro: You don't need those parentheses around the ini_get anyway.

kiros32: ok

Synchro:
Thanks!

PR #1896, opened 2019-11-15
Automated native_function_invocation fixes

draco2003:
Reproduce with:
php php-cs-fixer --rules=native_function_invocation fix ./ --allow-risky=yes
(php-cs-fixer from : https://cs.symfony.com/)

Details:
https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -22,7 +22,7 @@ return PhpCsFixer\Config::create()
         'phpdoc_summary' => false,
         'semicolon_after_instruction' => true,
         'simplified_null_return' => true,
-        'native_function_invocation' => false,
+        'native_function_invocation' => true,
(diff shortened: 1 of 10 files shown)


Synchro:
I get the underlying mechanism, but wow is that ugly! A single network request would completely dwarf any gain from this - SMTP servers often take whole seconds or even minutes to respond, so I really don't know that it's worth the readability hit.

draco2003:
Other libraries have opted for the other syntax of:
use function is_array

If that is preferred/easier to read happy to put in a PR with that style. 
100% agree on any I/O delay network/disk/etc.. it'll overshadow any gains like that. Doesn't mean things like this aren't beneficial.

PR #1909, opened 2019-12-03
Feature improve host check

petski:
Let isValidHost() determine validness of host

- PHP way of doing so (php/php-src#826)
- Improved and shortened regexp a bit
- Give more specific debug message in case host(entry) is invalid


petski:
Open for feedback. I need to know what maintainers thoughts are about the strictness in `isValidHost`, i.e. hostname "123" is perfectly valid, but is tested as a "bad" hostname.

src/PHPMailer.php
```diff
@@ -3857,6 +3857,10 @@ protected function serverHostname()
      */
     public static function isValidHost($host)
     {
+        if (version_compare(PHP_VERSION, '7.0.0') >= 0) {
+            return is_scalar($host) && filter_var($host, FILTER_VALIDATE_DOMAIN) !== false;
```
Synchro: `is_scalar` returns true for any string, but `FILTER_VALIDATE_DOMAIN` has quite a few problems, for example [this bug report](https://bugs.php.net/bug.php?id=77331). It also allows things that are definitely not usable hostnames, like `192.168.0.1/x` and `example`, even when the `FILTER_FLAG_HOSTNAME` is added for the latter example, but if you do add that flag it rejects IPv6 addresses. This isn't looking very reliable!

src/PHPMailer.php
```diff
@@ -1942,33 +1942,33 @@ public function smtpConnect($options = null)
         foreach ($hosts as $hostentry) {
             $hostinfo = [];
             if (!preg_match(
-                '/^((ssl|tls):\/\/)*([a-zA-Z\d.-]*|\[[a-fA-F\d:]+]):?(\d*)$/',
+                '/^(?:(ssl|tls):\/\/)?(.+?)(?:(\d*))?$/',
```
Synchro: The initial non-capturing group change is a good idea, but the rest of it loses most of the basic filtering for things that look vaguely domain-like, for example by allowing `exam ple.com`. I think there may well be a problem with the original matching on the port number, in that it will match trailing numbers even if they're not preceded by a `:`, but I don't think this new version deals with that either - for example it messes up on `example.com123`. To be fair, I'm not sure that domains ending in numbers are valid, but port numbers *definitely* need a preceding `:`. This needs lots of test cases!

src/PHPMailer.php
```diff
@@ -3857,6 +3857,10 @@ protected function serverHostname()
      */
     public static function isValidHost($host)
     {
+        if (version_compare(PHP_VERSION, '7.0.0') >= 0) {
```
Synchro: While I know that this constant was added in 7.0.0, I prefer to check for the feature directly, so `defined('FILTER_VALIDATE_DOMAIN')` instead of `version_compare` check. It's more specific, and a little safer (see also why we needed to do this for checking whether the openssl extension was loaded) .

src/PHPMailer.php
```diff
@@ -3874,7 +3878,7 @@ public static function isValidHost($host)
             //Is it a valid IPv4 address?
             return (bool) filter_var($host, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4);
         }
-        if (filter_var('http://' . $host, FILTER_VALIDATE_URL)) {
+        if (filter_var('http://' . $host, FILTER_VALIDATE_URL) || filter_var('http://' . str_replace('_', '-', $host), FILTER_VALIDATE_URL)) {
```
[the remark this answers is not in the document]
petski: Wow, you're quick :)

My initial thought was to allow hostnames like "my_container_name.docker", but I later came to the conclusion an underscore indeed isn't allow in a hostname. Therefor I fixed this up in a later commit.

src/PHPMailer.php
petski: I added the "is_scalar", so filter_var isn't fed with a class/array/etc.

Didn't knew about the unreliability of  FILTER_VALIDATE_DOMAIN/FILTER_FLAG_HOSTNAME .... "Great"

src/PHPMailer.php
petski: Sharp, regex should have been `'/^(?:(ssl|tls):\/\/)?(.+?)(?::(\d*))?$/'`, fixed that now

src/PHPMailer.php
petski: Clear, fixed this.

Synchro:
Regarding short ("dotless") names for hosts, while the RFCs don't technically prohibit them for email, they were [banned by ICANN](https://features.icann.org/dotless-domains) some time ago. Dotless names are not resolvable on the public internet, and there's no way to tell if a dotless name is a real hostname or a local alias - so the solution is to not use them directly, but to resolve them to an IP first, so the mailer is no longer responsible for establishing their validity. The most common example of this approach is of course `localhost`, which is a local alias - and that's why you'll see it defined in every `hosts` file.

Synchro:
Another quirk that relates to my earlier note about underscores: underscores are allowed in DNS records, they're just not allowed in DNS host name records that are meant to resolve to host IPs (`A`, `AAAA` and similar). A good example of this is things like `_dmarc` in TXT records.

src/PHPMailer.php
Synchro: That's good, thanks - I still don't see value in downgrading the match from `([a-zA-Z\d.-]*|\[[a-fA-F\d:]+])` to the much broader `(.+?)` - that will match pretty much anything at all, so it's not much use as a validation. It could be shortened by using the `/i` pattern modifier I guess.

petski: The reason I did, is because the "is a host valid"-responsibility now **only** lies in the isValidHost method:

a) Method isValidHost is public/static; checking the validness of a host for an outside caller isn't the complete set of checks that are performed within PHPMailer.
b) One could now derive from this class and only overload the isValidHost method to implement it's own check(s).

Do you agree that moving the `([a-zA-Z\d.-]*|\[[a-fA-F\d:]+])` regex to the isValidHost method will be a good solution?

src/PHPMailer.php
petski: Removed usage of FILTER_VALIDATE_DOMAIN/FILTER_FLAG_HOSTNAME

src/PHPMailer.php
petski: Did just that.

Synchro: I agree that moving responsibility there is a good idea. While we're there, checks like:

            return (bool) filter_var(trim($host, '[]'), FILTER_VALIDATE_IP, FILTER_FLAG_IPV6);

are probably better done as:

            return filter_var(trim($host, '[]'), FILTER_VALIDATE_IP, FILTER_FLAG_IPV6) !== false;

Though I spot a potential hole there too - `[[<ipv6>]]` would pass this test.

petski: Fixed. See latest commits

petski: Hmmmm, this has a side effect. See https://travis-ci.org/PHPMailer/PHPMailer/jobs/621083134?utm_medium=notification&utm_source=github_status

Synchro:
Could you remove the `composer` file that's ended up in here somehow, then I'll merge.

petski:
> Could you remove the `composer` file that's ended up in here somehow, then I'll merge.

I just removed the `composer` file and force pushed a new version. Thanks in advance (and sorry for the superfluous file)

Synchro:
Thanks for your effort on this.

PR #1934, opened 2020-01-05
Rename .php_cs to .php_cs.dist

voyula:


Synchro:
`.dist` is typically used for files that act as a template a real settings file, but one that is usually not used directly. This is not the case here: it's a real config file containing the settings for phpcs.

PR #1935, opened 2020-01-05
Create .editorconfig

voyula:


Synchro:
This is unnecessary bloat; it's implicit in the PHPCS config.

PR #1939, opened 2020-01-11
Add throwExceptions method

thisispiers:
If non-null value is passed, sets whether exceptions are thrown for errors. Always returns the current exception setting.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -837,6 +837,7 @@ class PHPMailer
         if (null !== $exceptions) {
             $this->exceptions = (bool) $exceptions;
         }
+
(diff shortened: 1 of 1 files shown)


Synchro:
I'm not sure I see the point in this; why would you need to turn this on and off dynamically, given that doing so would require you to completely rewrite the rest of your script, in which case the existing mechanism is likely sufficient?

thisispiers:
It means exceptions can be guaranteed without having to configure a dependency container (especially if the class gets passed around the codebase). If it's not needed that's cool

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

PR #1994, opened 2020-02-26
Make DKIM canonicalization configurable

rudiedirkx:
Microsoft/Hotmail/Outlook's DKIM check breaks with `relaxed/simple`, and works with `relaxed/relaxed`, because it loves to forward and fix and mess up white-space.

Doesn't change anything by default, but let the developer choose.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -502,6 +502,13 @@ class PHPMailer
     public $DKIM_domain = '';
 
     /**
+     * DKIM canonicalization to use.
+     *
+     * @var string
+     */
+    public $DKIM_canonicalization = 'relaxed/simple';
+
+    /**
      * DKIM Copy header field values for diagnostic use.
      *
      * @var bool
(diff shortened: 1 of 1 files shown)


Synchro:
This is a purely cosmetic change - you can't just *say* that you're using a different canonicalisation, you actually have to *do* the different canonicalisation, and the two approaches available are not the least bit compatible with each other. I can't merge this without an actual implementation of relaxed/relaxed.

rudiedirkx:
Ah crap. I thought `c` was just to let the recipient know how strict to check the hash. I'll look into how DKIM really works, finally.

Synchro:
The canonicalisation has to be done by both parties - the sender to generate the signature, and the receiver to verify it.

The simple algorithm tries to deal with the content as-is, with no changes, but this makes it less able to accept likely legitimate changes. The relaxed algorithm requires more processing, but is more tolerant of changes as a result.

PR #2001, opened 2020-03-01
Translate messages

jdruedaq:

### DIFF
--- a/src/POP3.php
+++ b/src/POP3.php
@@ -19,6 +19,7 @@
  */
 
 namespace PHPMailer\PHPMailer;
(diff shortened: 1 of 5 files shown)


Synchro:
Sorry, but I think this is a bit of a mess and breaks quite a few things. 

Constants are not a good choice in this situation as it means you cannot switch languages.

The naming scheme is inconsistent. 

Loading classes explicitly is a no-no too; if you’re going to add such classes, make them PSR-4 autoload. 

Adding a mandatory class would also be a significant BC break, and it’s not worth doing for this.

Issue #2002, opened 2020-03-04
PHPMailer::ENCRYPTION_STARTTLS


ueen:
Updated to the latest version via composer and set everything up according to the documentation, this line always provoked a 504 gateway timeout
`$mail->SMTPSecure = PHPMailer::ENCRYPTION_STARTTLS; `

so i changed it to
`$mail->SMTPSecure = 'ssl';`

and suddently it works, whats that about?

Synchro:
It just means that _your_ mail server supports SMTPS and doesn't support SMTP+STARTTLS for encryption; it's nothing to do with PHPMailer. The correct constant to use is:

    $mail->SMTPSecure = PHPMailer::ENCRYPTION_SMTPS;

ueen:
I do not think, that this is correct, my mail server DOES use SSL/TLS on Port 465
And why would you prefer STARTTLS over SSL?

Synchro:
SMTPS and SMTP+STARTTLS are **both** encryption mechanisms for SMTP, and both use TLS they are essentially equally secure. As I said, SMTPS on port 465 was deprecated in 1998, because it was replaced by SMTP+STARTTLS, however, it is still in common use (mainly because of Microsoft continuing to use it on Exchange), it has also been mostly resurrected by [RFC8314](https://tools.ietf.org/html/rfc8314).

PHPMailer use SMTP+STARTTLS as the default, because that's what the standards have said to use for the last 20 years. If _your_ mail server does support that, it's fine, but you should amend *your* code to deal with the difference, not PHPMailer itself.

ueen:
Thanks for your input, I learned quite a bit researching this, but as you yourself said, in respect to RFC8314, nowadays the port 465 is used for implicit SSL/TLS which is superior to STARTTLS, because it doesn't require cleartext communication, as clearly laid out in the very document you linked!
```
Connections to Mail Submission Servers and Mail Access Servers be made using "Implicit TLS" (as defined below), in preference to connecting to the "cleartext" port and negotiating TLS using the STARTTLS command or a similar command.
```

ueen:
But what why is it unadvisable to use the latest and most secure option: implicit TLS on port 465 as unambiguously stated in RFC8314?

Synchro:
Just because a server supports old SMTPS does not mean that it supports RFC8314 submission via implicit TLS on port 465 because they are not the same thing. Just because a server accepts connections on port 465 doesn't mean it supports submission.

[...]

PR #2003, opened 2020-03-04
Update README.md

ueen:
refer to https://github.com/PHPMailer/PHPMailer/issues/2002

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ try {
     $mail->SMTPAuth   = true;                                   // Enable SMTP authentication
     $mail->Username   = 'user@example.com';                     // SMTP username
     $mail->Password   = 'secret';                               // SMTP password
-    $mail->SMTPSecure = PHPMailer::ENCRYPTION_STARTTLS;         // Enable TLS encryption; `PHPMailer::ENCRYPTION_SMTPS` also accepted
+    $mail->SMTPSecure = 'ssl';                                  // Enable TLS encryption; `PHPMailer::ENCRYPTION_SMTPS` also accepted
     $mail->Port       = 587;                                    // TCP port to connect to
 
     //Recipients


Synchro:
This is a) not the right way to change this setting - use the `PHPMailer::ENCRYPTION_SMTPS` constant, and b) SMTPS was deprecated in 1998, so I'm not about to advise people to use it as the default now.

ueen:
Well I think you're wrong there, actually, TLS is considered obsolete https://tools.ietf.org/html/rfc8314 and SMTP with implicit SSL encryption is recommended! So please consider changing the example to
```
	    $mail->SMTPSecure = 'ssl';
	    $mail->Port       = 465;
```
Best regards! :)

Synchro:
I referred you to the same RFC in #2002. Bear in mind that the obsolete SMTPS on port 465 (what you're proposing) is not the same as what is described in RFC8314.

ueen:
I'm not proposing the obsolete port 465, but instead the very latest always SSL/TLS encrypted port 465, as recommended in the RFC document.

Synchro:
Please read what I said. Both SMTPS and RFC8314 use implicit TLS on port 465, so they look superficially identical. But you can't tell whether a server supports RC8314 except by trying to use it for submission and having it fail. This is not an approach that results in happy users, and is why it's not used as the default. Everyone who *does* have an RFC8314-compliant server is still entirely free to use it.

If on the other hand you're volunteering to change the default and then step in to diagnose every SMTP TLS problem it causes that is posted on here and Stack Overflow, you'd be very welcome.

ueen:
So obviously there is some confusion regarding standards and naming of them.
The only thing I was trying to do, was to kindly suggest changing the documentation to reflect, that the use of implicit TLS on port 465 according to RFC8314 is quite common today and can, actually should be used with PHPMailer, as I ran into troubles exactly because the documentaion did not examplify the use implicit TLS on port 465 according to RFC8314.

Synchro:
Had you read [this](https://github.com/PHPMailer/PHPMailer/wiki/Troubleshooting#encryption-flavours)?

ueen:
Not entirely, because i wasn't familiar with the differences and expected the example to work with the most recent standarts, which turned out to be a bit more tricky than expected.

Issue #2008, opened 2020-03-08
Multipart - missing extra EOL


lubosdz:
# Problem description

I believe there is missing one extra EOL in:
https://github.com/PHPMailer/PHPMailer/blob/master/src/PHPMailer.php#L2609
`$mimepre = 'This is a multi-part message in MIME format.' . static::$LE;`

The line should read properly:
`$mimepre = 'This is a multi-part message in MIME format.' . static::$LE. static::$LE;`

Missing extra LE causes, that when downloaded email like EML file, MS Outlook (tested at least 2010) will not recognize properly first boundary. So it will show up in plaintext. Once added extra line, it shows up properly HTML formatted. 

Currently rendered raw output is:

~~~
...
Content-Transfer-Encoding: 8bit
This is a multi-part message in MIME format.
--b1_7ZBRdlbIDcTVJNKZ2n35zPu19aYncKrNKjrpFKy6yM
Content-Type: text/plain; charset=utf-8
...
~~~

correctly should be (not extra line before first boundary):

~~~
...
Content-Transfer-Encoding: 8bit
This is a multi-part message in MIME format.

--b1_7ZBRdlbIDcTVJNKZ2n35zPu19aYncKrNKjrpFKy6yM
Content-Type: text/plain; charset=utf-8
...
~~~

The line "This is a multi-part message in MIME format." has been added about 3 yrs ago (~2016) according to git-blame and before that there had been there also 2 EOLs which worked fine. I found this issue after upgrading from 5.2.9 to newest version 6.1.4.

Synchro:
Neither of those examples are correct because there should always be a blank line between headers and body. [RFC2046 says](https://tools.ietf.org/html/rfc2046#section-5.1.1) that the boundary string must be preceded by a line break. The RFC has this to say about the preamble (the bit you're talking about):

> There appears to be room for additional information prior to the
   first boundary delimiter line and following the final boundary
   delimiter line.  These areas should generally be left blank, and
   implementations must ignore anything that appears before the first
   boundary delimiter line or after the last one.
   NOTE:  These "preamble" and "epilogue" areas are generally not used
   because of the lack of proper typing of these parts and the lack of
   clear semantics for handling these areas at gateways, particularly
   X.400 gateways.  However, rather than leaving the preamble area
   blank, many MIME implementations have found this to be a convenient
   place to insert an explanatory note for recipients who read the
   message with pre-MIME software, since such notes will be ignored by
   MIME-compliant software.

[...]

PR #2009, opened 2020-03-09
Fix issue #2008 - MS Outlook corrupts formatting multipart EML files

lubosdz:
Added extra EOL into preamble (before the very first boundary) for multipart messages.


Synchro:
Thanks

PR #2018, opened 2020-03-26
Fixed Timeout in stream_select (get_lines())

romkoll:
In stream_select must be used $this->Timeout, not $this->Timelimit


Synchro:
There is a mistake here, but it's in the subsequent debug output, not the `stream_select` call. `Timeout` is a timeout for initial connection and network outage, not the per-command timeout, which is what `Timelimit` is, so the code is correct but the debug output is not.

Issue #1579, opened 2018-10-20
Getting error message by error code


AkinOlawale:
Hi.

I am trying to implement a resend mail feature in my application. Basically should there be a connection issue or other related problems, the application saves the failed message to a database and then retries again after some minutes. 

Is there anyway to get an error code via errorInfo variable? For example, if it is an authentication problem, error code can be 3. etc?

XL-2000:
Although this seems a valid question at first glance, you want to resend regardless of the error right?
Where do you need the errorcode for? Next to this, keep in mind that auto-resending might be a bad idea i.e. if you run into quotum or abuse limitations.
Also, any error once the email is delivered to the mail server can not be caught on application-level, so keep in mind this will not be full proof.

Synchro:
I think this is a reasonable request. PHPMailer is usually set up in a fixed configuration, where it is always sending through the same mail server, and so any problems tend to be dealt with interactively during development, at which time descriptive error messages are appropriate.

Send-time SMTP errors already provide SMTP error codes, and bounces (which PHPMailer doesn’t attempt to handle) have embedded error information. Other runtime errors are harder to deal with because they do indeed lack clear error codes that would help you handle them programmaticly.

Something like an `ErrorCode` property that contains a numeric error code that could be used alongside the `ErrorInfo` property. Note that the `Exception` class already provides a value for that, but PHPMailer does not require the use of exceptions - the way exceptions are implemented is a little strange, but you can get more info from them than you can with `ErrorInfo` alone.

AkinOlawale:
@Synchro Yes that's exactly what i have in mind. The ErrorInfo property could be improved to contain numeric code. This should be a feature request

lsoutello:
I agree this should be a feature request. I need different treatment for each of the exceptions and the only way to do it right now is parsing the `ErrorInfo` message.

PR #2023, opened 2020-03-27
Add an option to get the send error in array format

christiandavilakoobin:

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -99,6 +99,13 @@ class PHPMailer
     public $ErrorInfo = '';
 
     /**
+    * Holds an array with the STMP send errors
+    *
+    */
+
+    public $ErrorDetails = [];
+
+    /**
      * The From email address for the message.
      *
      * @var string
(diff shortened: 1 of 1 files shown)


christiandavilakoobin:
I'm not sure if this is the best approach, but I need to get the error code returned by the smtp server to control it manually, e.g., to control the AWS SES limits (https://docs.aws.amazon.com/ses/latest/DeveloperGuide/troubleshoot-smtp.html) . If you reach the limit, they return a 454 error code. If all the error is concatenated in the ErrorInfo var, is more difficult to control it.

Synchro:
I think this is the same thing that was requested in #1579 some time ago.

christiandavilakoobin:
Yes, you're right. I can change the $STMPerror varname to ErrorCode to fit your suggestion, if you prefer it.

Synchro:
Can you show examples of what you're getting in `ErrorCode` and why you think an array is appropriate / needed? From its name I would expect `ErrorCode` to contain an error code, not an array!

christiandavilakoobin:
I changed it to `ErrorDetails`. I need it because with the SES daily and by second limits, I have to control it to retry sending emails later.

Here is an example:

```
try
{
    $mail->send();
}
catch (Exception $e)
{
   // If a limit is reached, AWS SES retruns a 454 error code
   if ($mail->ErrorDetail['code'] == 454)
   {
      // According AWS docs, we have the following error description.
      if ($mail->ErrorDetail['detail'] == 'Throttling failure: Maximum sending rate exceeded')
      {
         // Retry in a few seconds
      }
      elseif ($mail->ErrorDetail['detail'] == 'Throttling failure: Daily message quota exceeded')
      {
         // Wait until tomorrow
      }
      else
      {
         // Unkown SES issue
      }
   }
}
```

PR #2028, opened 2020-04-13
remove x-mailer from dkim sign headers

TheKito:
remove x-mailer from dkim sign headers because some exit smtp (proxy between corporate network and public network) remove al headers X-* to protect internal network information from public


Synchro:
I don't think this is needed - you can prevent the X-mailer header being set in the first place by telling PHPMailer to not add an x-mailer header, like this:

     $mail->XMailer = null;

Otherwise this represents a security downgrade.

Issue #2048, opened 2020-05-20
Integrate with google-api-php-client


sadashivdalvi:
I see that we are using OAuth and we have the phpLeague library, but I am already using [Google Api PHP Client](https://github.com/googleapis/google-api-php-client) and I think we can have a wrapper which supports the google api php client as well.

I can work on this just want to know all of your thoughts

Thanks,
Sadashiv.

Synchro:
There is already [an article about doing exactly this in the wiki](https://github.com/PHPMailer/PHPMailer/wiki/Gmail-XOAUTH2-Using-Google-API-Client). If you have any updates or improvements to that, go ahead and edit it; it's entirely open.

sadashivdalvi:
Thanks for the update, I missed that. 

Thanks,
Sadashiv.

PR #2051, opened 2020-05-21
Removing parameter type so that code works with PHP strict_types decl…

sadashivdalvi:
…aration and Gmail XOAUTH2

I was updated in #2048 about the [wiki page](https://github.com/PHPMailer/PHPMailer/wiki/Gmail-XOAUTH2-Using-Google-API-Client) to use google_api_php_client and I found that due to the parameter type php 7 strict types throws exceptions that it expected OAuth and found GmailXOAuth2 class object.

I need this fixed so that other devlopers using composer install can install phpmailer and use the smtp module of drupal cms directly without this modification or code break.

Thanks,
Sadashiv.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -4814,7 +4814,7 @@ class PHPMailer
     /**
      * Set an OAuth instance.
      */
-    public function setOAuth(OAuth $oauth)
+    public function setOAuth($oauth)
     {
         $this->oauth = $oauth;
     }


Synchro:
The OAuth class is more like an interface and doesn't have much in it – is there a way you could build a subclass that wraps the google_api_php_client to make it conform to that, rather than relaxing this type constraint?

sadashivdalvi:
Ok,

I got it working by overwriting the setOAuth function in the PHPMailerOAuth class as
```
  /**
   * Set an OAuth instance.
   */
  public function setOAuth(GmailXOAuth2 $oauth) {
    $this->oauth = $oauth;
  }

```

Ideally it doesn't match the core function definition but it's ok till we don't have a Interface defining the type of parameter, if you do that in future please include the GmailXOAuth2 as separate function or something else it will break for GmailXOAuth2 developers.

Thanks,
Sadashiv.

Synchro:
While that would work, it's not what I meant and is not safe, because there is no guarantee that the GmailXOAuth2 class is at all compatible with the PHPMailer OAuth class you're replacing. It will avoid that type exception, but will fail at some later point instead.

I was thinking more along the lines of:

```php
class PHPMailer\PHPMailer\GmailXOAuth2 extends PHPMailer\PHPMailer\OAuth
{
    //Implement mapping of all the functions of the OAuth class that are different for the Gmail client here
}
```

That way the gmail client class will act exactly as the provided OAuth base class expects. Another advantage here is that there are no BC issues.

sadashivdalvi:
Ok, that works, I think we should also make this change on wiki page because any one refering that can run into same problem.

Thanks,
Sadashiv.

PR #2053, opened 2020-05-22
Provide a content id to string attachments

emmanuel-deloget:
Since it's possible to create a string attachment with an 'inline'
$disposition we cannot blindly set what will become the attachment CID
to 0 otherwise we'll not be able to retrieve it using the 'cid:...' URL.

As with addAttachment(), set the future CID to the filename.

Signed-off-by: Emmanuel Deloget <logout@free.fr>

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -3487,7 +3487,7 @@ class PHPMailer
                 4 => $type,
                 5 => true, // isStringAttachment
                 6 => $disposition,
-                7 => 0,
+                7 => $filename,
             ];
         } catch (Exception $exc) {
             $this->setError($exc->getMessage());


Synchro:
I'm not sure I see the point of this. If you want to set a content ID on an inline attachment from a string, use `addStringEmbeddedImage` which does the same thing you're proposing here, except that it also preserves the semantics in the MIME structure. Also, the filename is not the same thing as a CID.

emmanuel-deloget:
Well, this is what addAttachment() is already doing and I was under the impression that addStringAttachment() should match the same semantics (just like addEmbeddedImage() and addStringEmbeddedImage() should also have the same semantics). 

Anyway, there is a problem in the existing code, and this is an attempt to solve it: either $disposition = 'inline' is legal when calling addStringAttachment() and in this case the CID should be set to something that's explicitely not 0 (I used the filename since addAttachment() is also using it) or $disposition = 'inline' is not legal on addStringAttachment() and the code should not allow it (and in this case you can set the CID to 0). The current situation is a mix of these two cases, and that prompted a good debug session on my side. 

For the record I have been using version 6.0.5 for monthes and I recently upgraded to version 6.1.5. Since my images are stored in a .phar archive I was using file pathes that are phar:// URLs. This no longer works in 6.1.5 (for a good reason) so I replaced my various uses of addAttachment() with "load the file + addStringAttachment()" - but the later forces the CID to 0 and thus does not behave the same way as addAttachment(), despite the name proximity.

Synchro:
Comparing the 4, I think:

* The `$name` parameter to `addStringEmbeddedImage` should be called `$filename` for consistency with `addStringAttachment`.
* `addAttachment` should not use `$name` as the CID since CIDs are meaningless for `disposition = attachment` parts.
* There's no good reason to allow setting disposition directly on these methods.

While you can indeed set the disposition for any of them, only the `EmbeddedImage` methods support setting a CID, so trying to get `addAttachment` to do it isn't what was intended. As far as I can see, using the `EmbeddedImage` variants solves your problem by using the method specifically intended for what you're doing, rather than altering one that doesn't?

emmanuel-deloget:
It should solve the problem, as I see it. 

In that case, it would be a good idea to not allow setting the content disposition on these methods (but that breaks the API compatibility) as allowing one to change it has adverse effects.

Synchro:
'inline' and 'attachment' are the only two dispositions in common use and the two sets of functions use their default values correctly. Given that trying to change them doesn't work well, I suspect you might be the only one trying to do so!

I think the cleanest way to resolve this particular issue is to **not** set a CID in either of the `Attachment` methods. That way it renders the ability to change disposition pointless (as inline without a CID, or a CID without inline make no sense), and I'd guess that essentially nobody is trying to do that anyway.

PR #2060, opened 2020-05-29
Split SMTP::connect in 2 functions to have more flexibility

BafS:
Connect is a pretty big function, here is a PR to split the logic to be able to overrides the way we connect to the SMTP server and have more flexibility.


Synchro:
Oops, merged too soon – this has a bug that causes tests to fail. I just pushed a fix for it. I'm not sure why that didn't show up in the original PR.

PR #2074, opened 2020-06-29
Retry stream_select if interrupted by signal

### LINKED ISSUE #1227: stream_select(): unable to select [4]: Interrupted system call (max_fd=7) in SMTP.php on 1116
Received two of the same errors in my error queue this evening and, for the life of me, I can't figure out how to troubleshoot. The relevant errors and stack trace from sending the email are:

```
WARNING: stream_select(): unable to select [4]: Interrupted system call (max_fd=7) in phpMailer/SMTP.php on 1116

Backtrace

phpMailer/PHPMailer.php 1310 calling postSend()

phpMailer/PHPMailer.php 1468 calling smtpSend()

phpMailer/PHPMailer.php 1712 calling smtpConnect()

phpMailer/PHPMailer.php 1867 calling connect()

phpMailer/SMTP.php 372 calling get_lines()

phpMailer/SMTP.php 1116 calling stream_select()
```

Sending emails has worked ample times before so I don't know what was special about this particular email. I've reviewed [#604](https://github.com/PHPMailer/PHPMailer/issues/604) and am wondering if it's related. Can anyone suggest how to troubleshoot or what may cause such an error? PHP v.7.0.25, PHPMailer 6.0.1, Debian 9. Thanks.

bartbutler:
Fixes #1227 by retrying interrupted stream_select calls and suppressing the associated warnings. This approach works regardless of original error handler status.

### DIFF
--- a/src/SMTP.php
+++ b/src/SMTP.php
@@ -1194,13 +1194,40 @@ class SMTP
         $selW = null;
         while (is_resource($this->smtp_conn) && !feof($this->smtp_conn)) {
             //Must pass vars in here as params are by reference
-            if (!stream_select($selR, $selW, $selW, $this->Timelimit)) {
+            //solution for signals inspired by https://github.com/symfony/symfony/pull/6540
+            set_error_handler([$this, 'errorHandler']);
+            $n = stream_select($selR, $selW, $selW, $this->Timelimit);
+            restore_error_handler();
+
+            if ($n === false) {
+                $message = $this->getError()['detail'];
+
+                $this->edebug(
+                    'SMTP -> get_lines(): select failed (' . $message . ')',
+                    self::DEBUG_LOWLEVEL
+                );
+
+                //stream_select returns false when the `select` system call is interrupted by an incoming signal, try the select again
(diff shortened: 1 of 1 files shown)


Synchro:
That's cool, thanks. Do you have any idea how this could be tested? How do you generate the error condition we've previously run into?

bartbutler:
So, we've testing it extensively now in production (we have a system which sends SIGTERMs if job run too long, and that was triggering the error for us).

For development what I did was turn on the low level debugging and add an `echo` immediately before the stream_select, and then hit Ctrl-C, which sends a SIGINT.

The error handler approach here works better than the `error_get_last()` approach in the Symfony PR because `error_get_last()` is empty in PHP 7 if an error handler is installed so it introduces dependency on external context (whether you have an error handler installed) whereas this works everywhere.

This also does nothing to suppress the signal, if you don't have a signal handler installed the program will exit on signal as usual, whereas if you do it'll wait for `pcntl_signal_dispatch()` as desired.

bartbutler:
Thanks!

PR #2085, opened 2020-07-15
Remove old links from SMTP test server suggestions

blueblots:
FakeSMTP's repo is read-only and hasn't been updated since 2016, so I removed it. The link to FakeEmail's repo was pointing at a fork that also hasn't been updated since 2016, so i changed the link to one that has more activity. Also smtp4dev has been moved to Github and works on Linux, so I updated its entry to reflect that.


Synchro:
While you're twiddling that, could you add a link to https://usehelo.com ?

Issue #85, opened 2013-07-23
Improve tests


Synchro:
Unit test code coverage is currently around 70%, and could do with being improved.

One major aspect of testing is not covered at all - while we're happily generating emails, there are no tests that check whether message content is what it is meant to be at the receiving end. This will require coordination with a local fake mail server (already part of the testing config).

mymas:
Hi,
Beforehand, I thank you for this project.

By the way, I’d like to contribute this issue and I have a little improvement.

However, this is my first challenge OSS. 
So, please advise me. For example, what the naming policy for create new branch.

I understand how to use git but I don’t know contributing github project.

Synchro:
When you create a fork, github will name it automatically. Typically it will be merged into master when a PR is accepted, so it doesn't matter what you call it.

mymas:
OK! I try it like you say. Thanks  a lot!

yuhanyuhan:
I realized that the test coverage is still at 73% and would love to improve on this lovely library!

mymas:
@yuhanyuhan 

Thanks you for your realizing this issue's PR!
 However, I need to fix a little bug like travis build failed, don't you?

jrfnl:
Just a general FYI as it is loosely related:

I'm currently working on pulling the previous, 5000+ line test class file apart into smaller, more manageable test classes with data providers and `@covers` tags, as well as improving the automated test runs done via CI.

This should make the existing tests more easily maintainable, make it easier to find where to add a test and to see what code is covered by tests and where new tests are needed.

I've been working with @Synchro on getting this refactor merged bit by bit. There have been scores of PRs over the past few weeks and more are expected still. (PR #2372 and later)

While this little project doesn't directly address the issue raised here (verification of mail as send on the receiving end), I figured it was worth mentioning in this issue anyhow.

PR #2118, opened 2020-08-24
#85 improvement/test improvement

criri99:

### DIFF
--- a/test/PHPMailerTest.php
+++ b/test/PHPMailerTest.php
@@ -3054,8 +3054,13 @@ eddiamnonumy
      * @test
      */
     public function encodedText_utf8CharBoundary_returnsCorrectMaxLength(){
-        $encodedText = 'H=E4tten';
-        $this->assertEquals(1, $this->Mail->utf8CharBoundary($encodedText, 3));
+        $encodedWordWithMultiByteCharFirstByte = 'H=E4tten';
+        $encodedSingleByteCharacter = '=0C';
+        $encodedWordWithMultiByteCharMiddletByte = 'L=C3=B6rem';
+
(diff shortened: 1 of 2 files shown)


test/PHPMailerTest.php
```diff
@@ -1234,7 +1234,7 @@ public function testHTMLAttachment()
         //Make sure that trying to attach an existing but unreadable file fails
         touch($filename);
         chmod($filename, 0200);
-        self::assertFalse($this->Mail->addAttachment($filename));
+        self::assertTrue($this->Mail->addAttachment($filename));
```
Synchro, line 1237: I'm not sure where you're going with these tests, but this change is wrong. `chmod 0200` will result in the file being marked as write-only (i.e. it exists, but it can't be read), so `addAttachment` is expected to fail and thus return false.

criri99: Ok, sorry. The test failed because of this assertion, so i found it an obvious error without reading the comment.

Synchro: I guess track down why that's failing for you – it's passing in the build on here.

Synchro: Ah, cool, that's very welcome. I see you added it to the title, but it doesn't get linked from there, needs to be in the text somewhere like #85

criri99: Ah, okay. Thanks for the hint. This is my first try on github, so i'm not sure how to efficently communicate.

Synchro:
Why did you close this? It looked like you were heading in the right direction.

criri99:
I tried to rename the branch to a more meaningful title. This was only possible by remplacing it with a new one 'increase_code_coverage'. Sorry for the circumstances. So i can move ahead with this issue?

PR #2130, opened 2020-09-03
Add AVIF file extension to mime list

Ayesh:
I'm not sure if the current mime list comes from anther software/project, but I'd like to propose to add AVIF image support to the mimes list. Chrome and Firefox support them in the latest versions, and I support their adoption rates will go up soon. We already have the relatively new heic extension, so I think adding avif makes sense too. 

Thank you.


Synchro:
Good addition, thanks

Issue #2137, opened 2020-09-18
SMTP connection not automatically reset on error when using keepalive


mscdex:
Please check these things before submitting your issue:

- [x] *Read the error message* you're seeing - it often tells you what is wrong, and may contain useful links & instructions
- [ ] Make sure you're using the latest version of PHPMailer
- [x] Check that your problem is not dealt with in [the troubleshooting guide](https://github.com/PHPMailer/PHPMailer/wiki/Troubleshooting), especially if you're having problems connecting to Gmail or GoDaddy
- [ ] Include sufficient code to reproduce your problem
- [ ] If you're having an SMTP issue, include the debug output generated with `SMTPDebug = 2` set
- [ ] If you have a question about how to use PHPMailer (rather than reporting a bug in it), tag a question on Stack Overflow with `phpmailer`, but [**search first**](http://stackoverflow.com/questions/tagged/phpmailer)!

# Problem description

We encountered this issue on an older version of PHPMailer (5.2.6 to be precise) running on PHP 5.6, however after reviewing the latest code in both the 5.2-stable and master branches, it appears the same bug still exists.

If a connection is set to utilize keepalive with SMTP, an RSET is not automatically sent when an error is encountered, which can cause artificial issues when additional mail sends are attempted on the same connection. One example would be if the server replies with an error in response to a MAIL FROM. In this case PHPMailer immediately throws/returns directly back to the `Send()` callsite without performing any kind of cleanup along the way.

I think it might be a good idea to automatically reset the session for the user upon receiving an error, especially since I was not able to even find any warning about end users needing to explicitly call `$mailer->smtp->Reset()` when `Send()` fails and keepalive is being used.

The way I've fixed this (for now) is by patching class.phpmailer.php to add the following to the exception handler inside `PostSend()`:

```php
// PHPMailer 5.2.x
if ($this->Mailer == 'smtp' && $this->SMTPKeepAlive == true) {
  $this->smtp->Reset();
}
```

# Code to reproduce

# Debug output

Synchro:
Good catch, thank you. Any chance of a PR?

PR #2138, opened 2020-09-18
Reset SMTP session after error with keepalive

mscdex:
Fixes: https://github.com/PHPMailer/PHPMailer/issues/2137

I couldn't figure out a way to trigger the issue with the current testing mechanism in the project.


Synchro:
Thanks. I know what you mean about testing – what we could really do with is a simple fake mail server that can generate specific errors on demand, along the same lines as what I've been trying to do in [the test classes for DKIMValidator](https://github.com/PHPMailer/DKIMValidator/tree/dev/tests).

PR #2147, opened 2020-10-01
Also use validateAddress in addOrEnqueueAnAddress

e1mo:
Prior to this commit E-Mails are if they contain an `@` sign in
`PHPMailer\addOrEnqueueAnAddress` before reaching the configurable
`validateAddress` in `PHPMailer\addAnAddress`. In 99.9999% of the
cases this will be perfectly fine. But however, it is possible to send
mail only to a user on the local system by their username. This is
especially neat for local development where you just have a sendmail
delivering mails to local user. Depending on your sendmail backend, it
would require additional configuration to create aliases with a domain
(e.g. something with `@something.tld`).

And if you provide a custom validator for that, I don't see any reason
to not use the configurable filter in that scenario.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -1069,7 +1069,7 @@ class PHPMailer
         $name = trim(preg_replace('/[\r\n]+/', '', $name)); //Strip breaks and trim
         $pos = strrpos($address, '@');
         // Allow custom validators to stop an email from being invalid just because
-        // a missing @ (See PR <placeholder>)
+        // a missing @ (See PR 2147)
         if (!static::validateAddress($address) && $pos === false) {
             $error_message = sprintf(
                 '%s (%s): %s',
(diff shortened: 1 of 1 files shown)


Synchro:
Two problems with this. Firstly, you can't validate IDNs at that point because you have not got reliable charset info – you only have that after `send()` is called, which is why addresses are queued and validated later, hence the name of the method. Secondly, bare usernames are not considered valid addresses in SMTP or RFC5322. If you want to send to a local user send it to `username@localhost.localdomain`. It's also an unhelpful edge case – if they are allowed at all, then they are permitted everywhere, and every contact form in the world that uses PHPMailer will allow addresses that won't work...

If it matters that much, I'd recommend subclassing for your use case.

e1mo:
Hi, thanks for your feedback. I'm wondering, what do you mean by sub-classing? Creating a wrapper extending the PHPMailer class and just overwriting the specific function?

Synchro:
Yes, exactly that.

PR #2151, opened 2020-10-02
Add MailHog in the code example as a tool for testing email sending

jenovateurs:
Hi!
MailHog is a tool that can be installed with Docker or GitHub Action to test email sending. I propose to add it in the README of the example folder.
The repository that I suggest shows how to use PHPMailer and to configure it, in order to work with MailHog easily.
Thanks for your review.


Synchro:
That looks good, thanks. What would be *even* better is a major revision to the PHPMailer test suite so that we can easily apply tests to sent messages; at the moment it relies on `getSentMIMEMessage()`, which isn't quite as good.

PR #2157, opened 2020-10-05
Add DSN mail option

vjardin:
It can be used to support the section 4.3 and 4.4 regarding
RET={FULL|HDRS} and ENVID=xyzIDxyz of the RFC:
  https://tools.ietf.org/html/rfc3461

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -372,6 +372,17 @@ class PHPMailer
     public $Timeout = 300;
 
     /**
+     * DSN Sender Extensions
+     * 'RET=FULL'  shall return the complete message in case of any error
+     *   xor
+     * 'RET=HDRS'  shall return only the headers of the failed e-mail.
+     * 'ENVID=xyz' shall include the ID xyz on the way back of this e-mail if any error.
+     *
+     * @see https://tools.ietf.org/html/rfc3461 See section 4.3 and 4.4 regarding RET and ENVID
+     */
+    public $dsn_ret = '';
+
+    /**
      * Comma separated list of DSN notifications
      * 'NEVER' under no circumstances a DSN must be returned to the sender.
      *         If you use NEVER all other notifications will be ignored.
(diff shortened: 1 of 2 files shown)


Synchro:
While this could be useful for some, I see a few issues with this implementation.

1. No consideration for [RFC6533](https://tools.ietf.org/html/rfc6533). While PHPMailer currently lacks support for SMTPUTF8, it does have support for IDNs, which have a bearing on this.
2. It's dangerous as there is no validation at all, making it an obvious point for SMTP and header injection attacks should someone be so foolish as to pass user-provided data into a property that ends up in the DSN.
3. How widespread is the adoption of this SMTP extension? I don't see it in any of gmail, yahoo, hotmail, but it is present in office365.

vjardin:
Regarding 1, it follows the same design than the support of recipient() https://github.com/PHPMailer/PHPMailer/blob/master/src/SMTP.php#L912 ; RFC6533 it a work to be done, but it shall be part of another pull request serie.

Regarding 2, you are right, I'll think about a better API, close to the support of $dsn (https://github.com/PHPMailer/PHPMailer/blob/master/src/PHPMailer.php#L375).

Regarding 3, of course it is widely supported for a while. On the sender side, for instance, postfix does support it too: https://manpages.debian.org/testing/postfix/sendmail.1.en.html

vjardin:
Regarding a DSN/ENVID reference, see: https://github.com/mozilla/releases-comm-central/blob/master/mailnews/compose/src/nsSmtpProtocol.cpp#L149

vjardin:
WIP with this previous commit. I'll provide an update using getServerExt()

vjardin:
@Synchro thanks for your comments. See enclosed this update of this pull request.

vjardin:
Dam'd ! I forgot php-cs-fixer, hold on please. I'll post an update.

vjardin:
done

vjardin:
@Synchro : please do you have some comments about this serie ? I included the features you requested.

thank you,

vjardin:
Please, do you have some comments ?

Synchro:
Sorry, lots of other things going on.

For the tests, could you reformulate that as tests within the test suite (in `tests/PHPMailerTests.php`)? Rather than having the values to encode and then checking them visually, write code that checks them.

vjardin:
OK, I'll give a try by end of the week.

PR #2160, opened 2020-10-06
Add ajax contact form

### LINKED ISSUE #1484: Add an AJAX/Javascript example

Sohib:
closes #1484 sorry i didn't check other pull requests.

### DIFF
--- a/examples/contactform-ajax.phps
+++ b/examples/contactform-ajax.phps
@@ -6,11 +6,11 @@
 //Import PHPMailer class into the global namespace
 use PHPMailer\PHPMailer\PHPMailer;
 
-$msg = '';
 //Don't run this unless we're handling a form submission
 if (array_key_exists('email', $_POST)) {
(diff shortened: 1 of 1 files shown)


Synchro:
This is very nice (thank you!), apart from one thing – it will make a bit of a mess if JS is disabled. Could you tweak the PHP to detect the XHR and use it to choose whether to provide a response as JSON or HTML, progressive-enhancement style?

Sohib:
@Synchro it's fixed now.

Synchro:
Thanks, that's great

PR #2162, opened 2020-10-08
Update README.md

nathanvanbakel:
Loading the class files manually needs the SMTP class as well: "use PHPMailer\PHPMailer\SMTP;"


Synchro:
No it doesn't. The SMTP class needs to be _loaded_ (either via a manual `require` or autoload), but it's very unlikely you will need to create an instance of it in your own script, so there's no need to import it into your namespace.

nathanvanbakel:
Thanks for the explanation and the very quick response!

PR #2164, opened 2020-10-08
Create phpmailer.lang-ua.php

Petro-Ivvysoft:

### DIFF
--- a/language/phpmailer.lang-ua.php
+++ b/language/phpmailer.lang-ua.php
@@ -1,0 +1,27 @@
+<?php
+/**
+ * Ukrainian PHPMailer language file: refer to English translation for definitive list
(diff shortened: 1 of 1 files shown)


Synchro:
[PHPMailer already has a Ukrainian translation](https://github.com/PHPMailer/PHPMailer/blob/master/language/phpmailer.lang-uk.php), but I see it has the wrong language code (`uk`, should be `ua`). Can you take a look at that and rebase your changes on top of that file, merging them as appropriate?

This misnaming has happened before and PHPMailer has [a remapping mechanism for renamed languages](https://github.com/PHPMailer/PHPMailer/blob/master/src/PHPMailer.php#L2108) that I can update for this (or you could add it to your PR if you like).

PR #2166, opened 2020-10-10
Add missing french translations

laurent35240:
Hi,

Here is some missing french translations and fix in sl file that I found when running `./vendor/bin/phpunit test/PHPMailerLangTest.php` command


Synchro:
Merci!

PR #2167, opened 2020-10-13
Rename phpmailer.lang-uk.php to phpmailer.lang-ua.php and checked the translation

Petro-Ivvysoft:


Synchro:
Thanks. I will close #2164 in favour of this one.

Synchro:
Ahem. I should have checked this properly. `uk` *is* [the correct ISO639-1 abbreviation for the Ukrainian language](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes#UK). I confused it with [the ISO3166 *country* code](https://en.wikipedia.org/wiki/ISO_3166-2:UA), which is indeed `UA`. So I've pushed a correction for this. Oh well, you get a @hacktoberfest credit anyway!

PR #2171, opened 2020-10-14
Replace deprecated curly brace access syntax

dungahk:
See https://wiki.php.net/rfc/deprecate_curly_braces_array_access


Synchro:
The 5.2-stable branch is no longer maintained. Switch to the 6.x version, where this issue was fixed long ago.

PR #2175, opened 2020-10-18
[POP3.php] small refactor: add constants DEBUG* like in SMTP class

proggga:
Add constants like in SMTP.php, add descriptions and getter/setter like in smtp

### DIFF
--- a/src/POP3.php
+++ b/src/POP3.php
@@ -63,12 +63,16 @@ class POP3
     const DEFAULT_TIMEOUT = 30;
 
     /**
-     * Debug display level.
-     * Options: 0 = no, 1+ = yes.
+     * POP3 class debug output mode.
+     * Debug output level.
+     * Options:
+     * @see POP3::DEBUG_OFF: No output
+     * @see POP3::DEBUG_SERVER: Server messages, connection/server errors
+     * @see POP3::DEBUG_CLIENT: Client and Server messages, connection/server errors
      *
      * @var int
      */
-    public $do_debug = 0;
+    public $do_debug = self::DEBUG_OFF;
 
     /**
      * POP3 mail server hostname.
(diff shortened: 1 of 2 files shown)


proggga:
Build passed, except nightly. After looking at error in build - I think thats ok
https://travis-ci.org/github/PHPMailer/PHPMailer/builds/736797707
image

src/POP3.php
```diff
@@ -404,6 +430,26 @@ public function getErrors()
...
     }
 
+    /**
+     * Set debug output level.
+     *
+     * @param int $level
+     */
+    public function setDebugLevel($level = 0)
+    {
+        $this->do_debug = $level;
```
Synchro, line 440: I don't think there is any point on having getters and setters if it's not going to do any validation and the property is public – it's easier to simply use the property directly, as it is in the other classes.

Synchro (commented):
I don't know that it's worth doing anything with the POP3 class. It's not a complete POP3 implementation, just a bare-bones class that can be used to do POP-before-SMTP authentication, which I've not seen anyone use for > 20 years. I did a quick search, and I can't find any projects that use this class at all, despite thousands using PHPMailer. I'm very tempted to drop it altogether, so I doubt tidier debugging is going to actually help anyone!

src/POP3.php
proggga: Smtp class have same public getter/setter with public do_debug field

proggga:
> I don't know that it's worth doing anything with the POP3 class. It's not a complete POP3 implementation, just a bare-bones class that can be used to do POP-before-SMTP authentication, which I've not seen anyone use for > 20 years. I did a quick search, and I can't find any projects that use this class at all, despite thousands using PHPMailer. I'm very tempted to drop it altogether, so I doubt tidier debugging is going to actually help anyone!

Delete pop3 file? Or close PR? What I can do for help
Sorry for disturbing, I want to really help

src/POP3.php
Synchro: You'r right! I don't think I've ever seen anyone use it because it's almost always set via the `SMTPDebug` property in the PHPMailer class, and [the one SMTP example](https://github.com/PHPMailer/PHPMailer/blob/master/examples/smtp_check.phps#L22) doesn't use it either. Still, I doubt anyone will be writing new code that uses this. I don't think there's any advantage in changing the API (adding the getter and setter), but changing to using constants for the debug levels is transparent and good practice, so I will accept the PR it if you drop the getter and setter and leave the constants in.

Synchro:
Dropping the POP3 class will need to wait for a new major version as it's a BC break, and I will need to mark it as deprecated ahead of time anyway.

src/POP3.php
proggga: Fixed

Synchro:
Thanks, good luck for @Hacktoberfest!

proggga:
> Thanks, good luck for @Hacktoberfest!

Thanks, you too!

PR #2182, opened 2020-10-26
Improve PHPCS config

jrfnl:
### PHPCS: rename config file

... to `phpcs.xml.dist` to allow devs to locally overload the file by using a `.phpcs.xml` or `phpcs.xml` file, to, for instance, test out some new rules.

Includes:
* Adding the local overload files to `.gitignore`.
* Adding the standard config file to `.gitattributes`.

### PHPCS: scan missing file

The `get_oauth_token.php` file in the project root seems to have been overlooked when configuring the PHPCS ruleset.

Fixed now by adding it to the file to be scanned.

Includes minor fixes to make the file comply with the configured standard.

### CS: fix two files

Two minor CS fixes.

### PHPCS: miscellaneous changes

* Don't fix the PHPCS/external standards version restraints.
    :point_right: **_I only just saw the standard PR text which refers to it being important to use a fixed version. Would you mind telling me why ? Generally speaking upgrading PHPCS to the latest version is always a good idea, so I'm wondering about this. If needs be, I can remove this specific change._**
* Add the PHPCS cache file to `.gitignore`.
* Removing the no longer existent `.php_cs` file from `.gitattributes`.


Synchro:
Thanks. I have removed that comment about phpcs versions in the PR template. That was probably a leftover from the early days of having issues with PHP-CS-Fixer.

Synchro:
I have no idea why I had fixed version deps for phpcs and phpcompatibility!

jrfnl:
Well, at least it's sorted now in that case ;-)

Issue #2179, opened 2020-10-22
Fix Travis no longer reporting on PRs


jrfnl:
It looks like Travis is slowly closing down their old interface. I have seen a number of repos running into issues with Travis not reporting back on branches/pull requests anymore and by the looks of it, this repo is affected as well.

Based on my experiences with other repos, this is what needs to happen:

> How to get TravisCI working again if it's currently not reporting:
> 1. On GitHub/ repo Settings: Remove the Webhook integration.
> 2. On GitHub/ marketplace: Install the Travis GH App
> 3. On travis-ci.com: sign up for the migrations beta for the organisation
> 4. On travis-ci.com: migrate all relevant repos
> 5. For PRs without status: force push problem PRs (no change needed in the content of the PR)
> 6. Once a first build has finished under the new system, change the protected branches to require "Travis CI - Pull Request" and un-require "continuous integration"

Source: https://twitter.com/jrf_nl/status/1255854838597984262

Synchro:
I've done 1-4 successfully.
For 5, PRs are not in repos I control, so how can I force-push them?
For 6, I don't think I have any protected branches, and I've not seen those settings – so do I need to do anything here?

jrfnl:
> For 5, PRs are not in repos I control, so how can I force-push them?

In most cases, PR branches are open for maintainers of the original repo to edit, so you can check them out and push back to them: 

![image](https://user-images.githubusercontent.com/663378/97217190-2dd81a00-17c7-11eb-8d51-13c50cf41773.png)


> For 6, I don't think I have any protected branches, and I've not seen those settings – so do I need to do anything here?

If you don't have protected branches, you can skip that step, though you _may_ want to set it up anyway - "Settings" -> "Branches" in the GH repo.

Synchro:
Oh, I've never noticed that before, from either end!

Synchro:
Is this working for you now?

jrfnl:
@Synchro We won't know until a new PR has been pulled ;-)

jrfnl:
Just saw #2186, which does show a status, so I guess it's working.

Synchro:
Cool.

PR #2183, opened 2020-10-26
Travis: simplify and improve setup and run tests on php 5.5

jrfnl:
As the changes are extensive, it will be easiest to review the changes by looking at the individual commits.

---

### Travis: reorder the config

No functional changes.

### Travis: remove redundant env variable

The "CS check" is being run in its own stage with its own script, so there is no need for this environment variable anymore.

### Travis: disable Xdebug before running Composer

Composer can be slower when Xdebug is enabled, so when Xdebug isn't needed, let's disable it _before_ running Composer.

### Travis: remove duplicate build definitions

The default stage is the `test` stage and all builds defined in the `php` key, possibly combined with an `env` key, will automatically be build in the `test` stage.

No need to define them twice.

The only builds to declare explicitly in `jobs` are those which need custom `env`, `dist`, `script`s etc.

So:
* Remove those builds which don't need anything "custom" from the `jobs` key.
* Remove the one build which _does_ need something "custom" from the `php` key.

### Travis: re-enable testing against PHP 5.5

PHP 5.5 is not available on the default `xenial` distro, but by explicitly telling Travis to use the `trusty` distro, we can still run the tests against PHP 5.5.

### Travis: run code coverage in a separate stage

Code coverage builds are "expensive" builds and if any of the tests would be failing, we are wasting time and resources by running them.

So, let's move the coverage build(s) to a separate stage, which will only run once the `test` stage has passed.

As this is now a separate stage, we don't need to set an environment variable to toggle code coverage on/off anymore. We can base conditions on the Travis stage name.

Let's also run code coverage for PHP 5.5, so both high and low PHP code coverage output can be combined for the most realistic results, taking PHP version specific conditions into account.

To make this work, we do need to tell Scrutinizer to expect reports from two separate code coverage runs.

And if I'm editing that file anyway, let's remove the redundant `filter` setting for a Scrutinizer native code coverage run which is turned off anyway.

### Tests: fix test failing on PHP 5.5

... due to the use of a PHPUnit method which is not available on PHPUnit 4.x.

These kind of things will be addressed better in a future PR, but this will get the build passing for now.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -9,6 +9,7 @@ tools:
     external_code_coverage:
         enabled: true
         timeout: 300
+        runs: 2
         filter:
             excluded_paths:
                 - 'docs/*'
@@ -19,13 +20,6 @@ tools:
 
     php_code_coverage:
         enabled:              false
-        filter:
-            excluded_paths:
-                - 'docs/*'
-                - 'examples/*'
-                - 'extras/*'
-                - 'test/*'
-                - 'vendor/*'
 
     php_code_sniffer:
         enabled:              true
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,11 @@
 language: php
 dist: xenial
 os: linux
+
+cache:
+  directories:
+    - $HOME/.composer/cache
+
 php:
   - 5.6
   - 7.0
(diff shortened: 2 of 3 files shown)


Synchro:
You're a star! Thanks!

jrfnl:
You're very welcome, my friend. Now might be a good time to also address #2179. From what I can see, the slow build problems we are seeing on .org are not happening on .com ;-)

PR #2186, opened 2020-11-01
Change Debugoutput to accept 'error_log' and other debug functions passed by string

wilfaustino:
Previous code exclude 'error_log' config for debug option.


Synchro:
This doesn't actually do anything useful. Because it's in the constructor and not from a parameter, `Debugoutput` will only ever contain its default value, which is `echo`, so making this change would prevent it from generating HTML output by default in a web environment. If you want to change it to something else, set it after creating the instance:

```php
$mail = new PHPMailer();
$mail->Debugoutput = 'error_log';
```

PR #2188, opened 2020-11-06
Always use CRLF if php version is 8.0 or newer

yaslama:
Since https://git.php.net/?p=php-src.git;a=commit;h=6983ae751cd301886c966b84367fc7aaa1273b2d
which solved the bug https://bugs.php.net/bug.php?id=47983, the headers
have to always be CRLF separated.
Without this fix, a space is prepended to the header names.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -1462,7 +1462,7 @@ class PHPMailer
     {
         if (
             'smtp' === $this->Mailer
-            || ('mail' === $this->Mailer && stripos(PHP_OS, 'WIN') === 0)
+            || ('mail' === $this->Mailer && (PHP_VERSION_ID >= 80000 || stripos(PHP_OS, 'WIN') === 0))
         ) {
             //SMTP mandates RFC-compliant line endings
             //and it's also used with mail() on Windows


Synchro:
I'm not sure it's as simple as that. The behaviour I see is not just in the headers that PHP adds (which is what the patch you refer to is about), but that on unixish platforms, `mail()` is expected to use `LF` for *all* line breaks, not just those added by PHP, which this patch has no bearing on. Your proposed change would result in `CRLF` being used on platforms that are expecting `LF`; the result of this is often "double spaced" emails.

Synchro:
It may be the case that sendmail binaries on those platforms will accept either line break format, so long as it's consistent, but that we have historically had to use `LF` because of the headers added by PHP so we have had to ensure that any other breaks we added (e.g. in message body) stuck to the same format. PHPMailer goes to quite some lengths to ensure that line break format is consistent either way.

yaslama:
To add more context why this PR is needed, I compiled php8 rc3 in debian using the official debian repository  https://salsa.debian.org/php-team/php/-/tree/debian/main/8.0 and used it in a clean debian installation. By default exim4 is installed.
And I used the following code to send a test email:
```
$mail = new PHPMailer(true);

try {
	$mail->AddReplyTo("email@example.org", "Sender");
	$mail->SetFrom("email@example.org", "Sender");
	$mail->AddAddress("name@example.org", "Name");

	$mail->Subject = 'Test subject';

   
	
	$mail->isHTML(true);
	$mail->Body    = '<h1>Hello World</h1>';
	$mail->AltBody = 'text';

	$mail->Send();
	echo 'Message has been sent';
} catch (Exception $e) {
    echo "Message could not be sent. Mailer Error: {$mail->ErrorInfo}";
}
```
Without the PR, when using php8.0, the headers shown by the client which received the email are:
```
Subject: Subject
Date: Thu, 5 Nov 2020 14:58:19 +0200
 From: <deleted>
 Reply-To: <deleted>
 Message-ID: <03O5vF6ZdLnJq9fM2f87k9VKmztHuTsDOV631ONm9g@upgrade24>
 X-Mailer: PHPMailer 6.1.8 (https://github.com/PHPMailer/PHPMailer)
 MIME-Version: 1.0
 Content-Type: multipart/alternative;
  boundary="b1_03O5vF6ZdLnJq9fM2f87k9VKmztHuTsDOV631ONm9g"
 Content-Transfer-Encoding: 8bit
Message-Id: <E1kaeqJ-006B6E-U6@upgrade24>

```
As you can see, there is a space before some headers and the client cannot display it correctly.
With php 7.4 everything is ok.
Wen using this PR, the headers are ok in both php7.4 and php8.0rc3.
When using this PR and sending a text only body with LF between lines like:
```
$mail->Body    = "line1\nline2\nline3";
```
I don't see any "double spaced" lines using both php7.4 and php8.0
Do you have another suggestion fixing the bug with php8.0?

Synchro:
The double spacing can occur if you use CRLF on a platform expecting LF – so in your example do you see extra breaks if your test string is `$mail->Body    = "line1\r\nline2\r\nline3";`? Are the CRLF breaks preserved correctly?

yaslama:
> The double spacing can occur if you use CRLF on a platform expecting LF – so in your example do you see extra breaks if your test string is `$mail->Body = "line1\r\nline2\r\nline3";`? Are the CRLF breaks preserved correctly in the delivered message?

When using `$mail->Body = "line1\r\nline2\r\nline3";`, I don't see any double spaced lines using both 7.4 and 8.0.
The email is stored in the recipient server (exim + dovecot) using LF between lines, and in thunderbird, the lines are shown as CRLF separated.

Synchro:
OK, that sounds like a good thing overall. I expect that PHPMailer is enforcing consistent LF breaks for local delivery, and exim is translating the LFs to CRLFs when doing onward delivery, as it should. It would be good to test this on Windows to confirm that your fix is correct.

yaslama:
>It would be good to test this on Windows to confirm that your fix is correct.

For sure, the fix doesn't change anything in Windows, because it changes the behavior of php 8.0+ in all platform to be the same as the behavior when using Windows.

Synchro:
Ah, yes, that's true. Thanks for bearing with me on this, I've merged it now.

yaslama:
> Ah, yes, that's true. Thanks for bearing with me on this, I've merged it now.

:+1:

barhun:
Hey,

I think this change breaks otherwise working behavior on PHP8 with opensmtpd on Linux. opensmtpd rejects the input with a message saying `smtp failed-command command="DATA" result="550 5.7.1 Delivery not authorized, message refused: Message is not RFC 2822 compliant"`.

I replaced the sendmail binary with a script that writes the input sent by the PHP-FPM process to a file and did some debugging with it, here is what I found: This change writes all line breaks as CR+LF while opensmtpd expects an empty line consisting only a LF where the headers end. If a delete the CR from the empty line after the headers, everything works as expected. So what I suggest is that CR+LF must only be used to terminate the lines used for headers, and all other line breaks should be the ones expected by the platform: LF on UNIX, CR+LF on Windows.

Synchro:
That just sounds like a bug in the mail server. Some mail servers will work with LF breaks, but I don't see any way that *requiring* mixed line break formats is anything but a bug. There may be a problem, but I very much doubt that's it.

barhun:
@Synchro Hmm, after reading the [RFC](https://datatracker.ietf.org/doc/html/rfc2822#section-2), I now think you're right. However, it also mentions that all line breaks before the body **must be** CR+LF. So, your implementation is correct for both PHP8 on Linux and PHP on Windows; but some servers might still reject messages from earlier versions of PHP 8 on Linux.

I'll also report this bug to the opensmtpd team asap. Thanks for your response, btw.

Synchro:
This is also due to numerous problems related to PHP's `mail()` function. Email standards are quite clear - **all** breaks should be CRLF. But in PHP, the mail function expects LF on unix-like platforms, CRLF on windows, and it has a long history of messing things up, along with line length problems, and shell escaping.
One way of avoiding such problems is to use SMTP to localhost rather than using `mail()` as the transport. This is the recommended approach anyway because it's faster, safer, and easier to debug.

PR #2190, opened 2020-11-07
Removed missing example link

jeremyVignelles:
The code_generator.phps example seems to have been removed. Remove the associated link in the README


Synchro:
Quite right, thank you

PR #2198, opened 2020-11-19
Preference to NTLM before LOGIN type.

prasad83:
NTLM should be chosen over LOGIN type of authentication.


Synchro:
This is a legacy branch that's no longer supported, the NTLM authentication never actually worked anyway, and it's been removed altogether in the current version.

PR #2200, opened 2020-11-21
Sun Nov 22 05:08:40 IST 2020

optiwariindia:
Before submitting your pull request, check whether your code adheres to PHPMailer coding standards (which is mostly [PSR-12](https://www.php-fig.org/psr/psr-12/)) by running [PHP CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer):

    ./vendor/bin/phpcs

Any problems reported can probably be fixed automatically by using its partner tool, PHP code beautifier:

    ./vendor/bin/phpcbf
I have tried to make translation more hindi than urdu... If you are providing hindi translation don't prefer to use urdu words...

### DIFF
--- a/language/phpmailer.lang-hi.php
+++ b/language/phpmailer.lang-hi.php
@@ -3,25 +3,25 @@
 /**
  * Hindi PHPMailer language file: refer to English translation for definitive list
  * @package PHPMailer
- * @author Yash Karanke <mr.karanke@gmail.com>
+ * @author Om Prakash Tiwari <optiwari.india@gmail.com>
  */
 
-$PHPMAILER_LANG['authenticate']         = 'SMTP त्रुटि: प्रामाणिकता की जांच नहीं हो सका। ';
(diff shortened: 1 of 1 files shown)


Synchro:
I'm not a Hindi speaker, however, comparing the output of google translate of this PR with what's already there, the original seems to be more accurate and better formatted. For example, the original has `connect_host` as `त्रुटि SMTP: अनुपलब्ध आतिथेय` which google translate tells me means "error SMTP: unavailable hospitality" (Note the lack of correct capitalisation too). This contrasts with the original version of `SMTP त्रुटि: SMTP सर्वर से कनेक्ट नहीं हो सका।` which translates back as "SMTP error: Could not connect to SMTP server.", which is formatted very nicely and corresponds exactly with what the original English says.

In almost every case, the translations of the original are better than what you have proposed. Now I know that Google Translate is not perfect, but it doesn't usually get things that badly wrong.

Why did you feel the need to retranslate every string?

PR #2203, opened 2020-11-25
Use fully-qualified constant names for PHP versions

Ayesh:
This is a micro performance improvement. When PHP opcode generates opcodes, it can remove PHP version-specific `if` blocks if it sees `PHP_MAJOR_VERSION` or `PHP_VERSION_ID`. However, if the code belongs to a namespace, it cannot make that optimization because the code under namespace can declare the same constants.
This PR updates such PHP constants to be fully-qualified (with back-slash), which enables PHP to make the improvement.

To compare, this is the VLD opcode without fully-qualified constant names:

```php
namespace X;

if (\PHP_VERSION_ID < 80000) {
    echo "hi";
}
```

-->

```
line     #* E I O op                           fetch          ext  return  operands
-------------------------------------------------------------------------------------
   5     0  E >   FETCH_CONSTANT                                   ~0      'X%5CPHP_MAJOR_VERSION'
         1        IS_SMALLER_OR_EQUAL                              ~1      ~0, 8
         2      > JMPZ                                                     ~1, ->4
   6     3    >   ECHO                                                     'hi'
   7     4    > > RETURN                                                   1
```

The same snippet, with the fully-qualified constants, PHP can simply eliminate and optimize the `if` block:

```
line     #* E I O op                           fetch          ext  return  operands
-------------------------------------------------------------------------------------
   5     0  E > > JMPZ                                                     <true>, ->2
   6     1    >   ECHO                                                     'hi'
   7     2    > > RETURN                                                   1
```

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -1462,7 +1462,7 @@ class PHPMailer
     {
         if (
             'smtp' === $this->Mailer
-            || ('mail' === $this->Mailer && (PHP_VERSION_ID >= 80000 || stripos(PHP_OS, 'WIN') === 0))
+            || ('mail' === $this->Mailer && (\PHP_VERSION_ID >= 80000 || stripos(PHP_OS, 'WIN') === 0))
         ) {
             //SMTP mandates RFC-compliant line endings
             //and it's also used with mail() on Windows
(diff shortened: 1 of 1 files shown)


Synchro:
Thanks, that's a neat optimisation. I think someone proposed a similar thing for global functions names a while ago, but it is really bad for readability. For these few constants I think it's fine though.

Ayesh:
Thanks a lot @Synchro. I understand the readability concerns, and sometimes it's not worth the loss of it for the little performance gain. Thanks for your time and decision to merge this 🙏🏼.

PR #2218, opened 2020-12-09
Migrate from TravisCI to GitHub Actions

codedge:
Hey,

I just read [your tweet](https://twitter.com/SynchroM/status/1336452160180117508) about asking for help with Github Actions. This PR adds support for GH checking the same things as with Travis CI. I tried to resemble all the settings from your `.travis.yml`, hopefully I did not fail.

**First stage:**
Coding standards run before the tests are run. If that stage fails, it does not continue with testing the code. This stage is run with PHP 7.4.

**Second stage:**
Here the phpunit tests are run and code coverage is generated. After that the coverage is uploaded to Scrutinizer, which fails in my fork, as I don't have the permissions to upload coverage to Scrutinizer,

All stages are run on Ubuntu 16.04. You can see the result here in my fork: https://github.com/codedge/PHPMailer/actions/runs/410752607

I checked the unit test results with the results that are currently in Travis CI which seem to match. Let me know, if I can help with anything more regarding Github Actions.


Synchro:
This is great, thank you. There are two omissions, but I'll have a go at dealing with them. Firstly PHP 8.0 support was in Travis, but hiding behind "nightly". Secondly there are builds for Power architecture which were only added recently, however, I don't think GH Actions supports that yet.

codedge:
Yes, PowerPC is not supported by Github Actions. That's why I left that out. I wasn't sure about the nightly, thanks for clarifying. PHP 8.0 I am going to add.

PR #2233, opened 2021-01-04
Allow to specify content ID in addStringAttachment

werbfred:
I propose to add a 6th optional parameter to specify the content ID in function PHPMailer::addStringAttachment.

In current version content ID it is not set, disallowing to add an attachement of an image into the HTML body via addStringAttachment.

In my personal use case I'm allowing to select a mailer among several ones. Adding right now PHPMailer. Base64 encoding seems to be the most common and low level attachment way. 

Pushing things forth and back, I don't see a valid reason why a base64 encoded attachment could not be used with a content ID reference.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -3485,6 +3485,8 @@ class PHPMailer
      * @param string $encoding    File encoding (see $Encoding)
      * @param string $type        File extension (MIME) type
      * @param string $disposition Disposition to use
+     * @param string $cid         Content ID of the attachment; Use this to reference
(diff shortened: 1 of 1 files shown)


Synchro:
This is already possible in [`addStringEmbeddedImage`](http://phpmailer.github.io/PHPMailer/classes/PHPMailer.PHPMailer.PHPMailer.html#method_addStringEmbeddedImage)?

werbfred:
Ouch... Had seen `addEmbeddedImage `but not `addStringEmbeddedImage`

Thanks... Pull request is usells though 👍

PR #2271, opened 2021-02-22
Implement Sender in message header

kuzin-shoptet:
Add option to add 'Sender' type with 'Reply-To'


Synchro:
Simple question – why? The sender field is mostly used when there are multiple `from` addresses to identify one as the sender – but PHPMailer doesn't support that anyway, so it's not much use. What are you hoping to do with this that you can't already do with From, Reply-To, and (envelope) Sender addresses?

Issue #2298, opened 2021-03-30
Copying the envelope sender from the From header causes failures


robertmathews:
A recent change in commit 09bc3065290431cad8c345b852386789ca852396 makes PHPMailer pass the `-f` flag to sendmail in cases where it didn't use `-f` (i.e., where it previously let the system choose the envelope sender address / return path). This leads to failures in situations where it previously worked because the sendmail default envelope sender was used.

There's discussion of this in [this WordPress bug report](https://core.trac.wordpress.org/ticket/52822), but the overview is that (for example) many WordPress plugins send mail with a "From" header address that they don't really own.

This is obviously not best practice because it means DKIM won't work, but as long as `-f` is *not* passed to sendmail, the envelope sender address is likely to be valid. Either the system's default sendmail envelope sender address or the php.ini sendmail_from address will be used, and assuming one of those is correctly configured (which is usually the case with competent hosting companies, for example), the resulting mail will pass SPF checks and hopefully be delivered.

With this change, PHPMailer and WordPress send that message with `-f same_address_as_from_header`, overriding sendmail's default envelope sender. That means that SPF won't match for the message, either, and it's much more likely to be rejected as a forgery. (I've seen this problem happening in the wild after people updated to the version of WordPress that includes this PHPMailer change. Messages that used to be delivered are now rejected outright.)

In case an example helps, imagine a message was sent with "From: example@gmail.com" by a script that is running as user "user123" at "hostingcompany.com". Since the WordPress wp_mail() function calls PHPMailer without specifying a "sender", that message would previously be sent as:

```
Return-Path: user123@hostingcompany.com
From: example@gmail.com
```

This would probably work: While the "sending with a From header address you don't own" is not recommended for DKIM/DMARC purposes, at least the Return-Path envelope address is a valid address that can be matched using SPF, and it would likely be delivered. Now, however, the same message would be sent as:

```
Return-Path: example@gmail.com
From: example@gmail.com
```

Neither SPF or DKIM will be valid, so this message is much more likely to be rejected as a forgery.

It doesn't seem that the new "generating a default -f envelope sender from the header From address" behavior when no explicit sender is provided is a good idea. It may fix problems where the default PHP/sendmail binary envelope sender is misconfigured, but those have failed all along... and this change introduces new failures where the sendmail envelope default was properly configured.

If the envelope sender is unspecified when calling PHPMailer, it seems better to let PHP and the sendmail binary choose the default.

Synchro:
[...]

PR #2300, opened 2021-03-31
Don't copy an empty envelope sender from From header for sendmail

robertmathews:
Revert change that made the `mail()` and sendmail transports set the envelope sender if one isn't explicitly provided, as it causes problems described at <https://github.com/PHPMailer/PHPMailer/issues/2298>


Synchro:
Thanks, I'll tag this for release shortly

Synchro:
6.4.0 is now tagged, hope that fixes things for you!

robertmathews:
This is awesome, thank you for being so responsive.

audrasjb:
Thanks all for the super quick resolution.
Kudos from the WordPress Core Team.

PR #2305, opened 2021-04-08
Updated callback.phps

nicolomonili:
Updated example

### DIFF
--- a/examples/callback.phps
+++ b/examples/callback.phps
@@ -26,13 +26,13 @@ function callbackAction($result, $to, $cc, $bcc, $subject, $body)
 {
     echo "Message subject: \"$subject\"\n";
     foreach ($to as $address) {
-        echo "Message to {$address[1]} <{$address[0]}>\n";
+        echo "Message to {$address} <{$address}>\n";
     }
     foreach ($cc as $address) {
-        echo "Message CC to {$address[1]} <{$address[0]}>\n";
+        echo "Message CC to {$address} <{$address}>\n";
     }
     foreach ($bcc as $toaddress) {
-        echo "Message BCC to {$toaddress[1]} <{$toaddress[0]}>\n";
+        echo "Message BCC to {$toaddress} <{$toaddress}>\n";
     }
     if ($result) {
         echo "Message sent successfully\n";
@@ -40,8 +40,6 @@ function callbackAction($result, $to, $cc, $bcc, $subject, $body)
         echo "Message send failed\n";
     }
 }
-
-require_once '../vendor/autoload.php';
 
 $mail = new PHPMailer();
 


Synchro:
That doesn't look right - the `$to` param [is an array](https://github.com/PHPMailer/PHPMailer/blob/master/src/PHPMailer.php#L1989), so I would expect this change to result in array-to-string conversion errors. That said, the example uses `$address[0]`, which will work, but also `$address[1]`, which doesn't exist, and should contain the to name. That's an omission in the callback code. I think that this line:

                $callbacks[] = ['issent' => $isSent, 'to' => $to[0]];

should become

                $callbacks[] = ['issent' => $isSent, 'to' => $to];

so that it contains all info about the recipient.

nicolomonili:
Ohh yes, that was the problem!

nicolomonili:
It should be so yes

`$callbacks[] = ['issent' => $isSent, 'to' => $to];`

The example would only be correct if both information were present (array with address and name).

Synchro:
It's actually a bit more involved – the use of `doCallback` is inconsistent. Sometimes it will contain (as in this case):

    ['address@example.com`]

but in other places the whole of `$mail->to` is passed in, which may contain structures like:

    [['address@example.com`, 'User Name'], ['address2@example.com`, 'User2 Name']]

The first case should be amended to match the second pattern.

nicolomonili:
There is also another problem

```
$callbacks = [];
//Attempt to send to all recipients
foreach ([$this->to, $this->cc, $this->bcc] as $togroup) {
    foreach ($togroup as $to) {
        if (!$this->smtp->recipient($to[0], $this->dsn)) {
            $error = $this->smtp->getError();
            $bad_rcpt[] = ['to' => $to[0], 'error' => $error['detail']];
            $isSent = false;
        } else {
            $isSent = true;
        }

        $callbacks[] = ['issent' => $isSent**, 'to' => $to[0]**];
    }
}
```
By doing this, the parameters **cc** and **bcc** aren't updated in the callback. Only the **to** parameter is updated.

Synchro:
I have pushed a change in master that should make calls to `doCallback` consistent; addresses will now always have the second format, as documented, and expected by this example. Please give it a try.

nicolomonili:
There still seems to be the problem. Maybe the change should be made [here](https://github.com/PHPMailer/PHPMailer/blob/master/src/PHPMailer.php#L1750) too?

Synchro:
No, because the `to` property is already in the right format. What problem are you seeing?

nicolomonili:
Ok found the problem
[Here](https://github.com/PHPMailer/PHPMailer/blob/master/src/PHPMailer.php#L2000), **to** is not an array.

Synchro:
Thanks, I missed that one, fix pushed

nicolomonili:
Tested, now it works!
Now only [this](https://github.com/PHPMailer/PHPMailer/pull/2305#issuecomment-815601773
) problem remains to be solved.

Synchro:
What problem? The original callback example should now work correctly?

nicolomonili:
The original callback example now it works, now **to** it's an array of arrays.
But

> the parameters cc and bcc aren't updated in the $callbacks array. Only the to parameter is updated.

Synchro:
> By doing this, the parameters cc and bcc aren't updated in the $callbacks array. Only the to parameter is updated.

I don't think that really applies here. At the SMTP level there is no distinction between to, cc and bcc – they are all just `RCPT TO` commands.

nicolomonili:
Ok, so in the callback i will have all the addresses in the parameter **to**? Even if they were added as **CC** or **BCC**?

Synchro:
No, with SMTP you will get a separate callback for each message recipient. I don't think it's important whether the address was originally in to, cc or bcc, because you were the one that set them!

The mail() transport behaves differently because you don't get that opportunity, and you will get all of to, cc, and bcc in a single hit, broken out into those params, and you can't tell which address failed. If it acted the same way for SMTP, you would not be able to get the separate error status for each recipient that you were originally looking for.

nicolomonili:
Maybe i explained myself wrong, sorry.

**With SMTP a separate callback for each message recipient it's perfect.**

But in this case :
```
...
$mail->addAddress("email1@gmail.com");
$mail->addBCC("email2@gmail.com");
$mail->addCC("email3@gmail.com");
...
```
in the callback function the parameters **to**, **bcc** and **cc** will they be like this?
```
$to = [["email1@gmail.com", ""],["email2@gmail.com", ""],["email3@gmail.com", ""]]
$bcc = []
$cc = []
```
and not so?
```
$to = [["email1@gmail.com", ""]] 
$bcc = [["email2@gmail.com", ""]]
$cc = [["email3@gmail.com", ""]]
```

Synchro:
No, you will get three separate callbacks, each time with a different `$to` value:
```
$to = [["email1@gmail.com", ""]]

$to = [["email2@gmail.com", ""]]

$to = [["email3@gmail.com", ""]]
```
Whether an address is to, cc, or bcc has no bearing on the sending status – you're never going to have an address that fails in `cc` but works in `to`, because at the point that happens, the receiving server can't tell them apart.

nicolomonili:
I have explained myself badly once again, perfect 🤣
However now i understand that in any case the address will be in the **to** parameter, thanks

PR #2340, opened 2021-05-17
Fix DKIM's error when there is S/MIME-sign

anhtunguyen:
In $parts[0] there is MIME-Version and Content-Type delimited by "\n".
Need to separate them, otherwise error with DKIM: "dkim=neutral (body hash did not verify)" (The reason is in "h=Date:To:From:Reply-To:Subject:Message-ID:X-Mailer;" without MIME-Version and Content-Type)

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -2933,7 +2933,14 @@ class PHPMailer
                     @unlink($signed);
                     //The message returned by openssl contains both headers and body, so need to split them up
                     $parts = explode("\n\n", $body, 2);
-                    $this->MIMEHeader .= $parts[0] . static::$LE . static::$LE;
+                    //In $parts[0] there is MIME-Version and Content-Type delimited by "\n".
+                    //Need to separate them, otherwise error with DKIM: "dkim=neutral (body hash did not verify)"
+                    //(The reason is in "h=Date:To:From:Reply-To:Subject:Message-ID:X-Mailer;" without MIME-Version and Content-Type)
+                    $subparts = array_map('trim', explode("\n", $parts[0]));
+                    foreach($subparts as $subpart) {
+                        $this->MIMEHeader .= $subpart . static::$LE;
+                    }
+                    $this->MIMEHeader .= static::$LE;
                     $body = $parts[1];
                 } else {
                     @unlink($signed);


Synchro:
Thanks for this – I don't think anyone has actually used this combination before, so I'm happy to receive fixes like this!
Do you know for sure that openssl always uses `\n` line breaks on Windows, or should your substitution normalise line breaks first?

anhtunguyen:
@Synchro Openssl uses \n\n to separate headers and content. This is what everyone knows. I'm not sure if openssl uses the unique \n to split the lines of headers. Maybe \r\n too? But \n is present in all cases. That's why I split the lines with \n, then use the function TRIM for those lines. This way, if \r is present, it will be removed as well.

Synchro:
Thanks for clarifying. Do you have tests to cover this?

anhtunguyen:
@Synchro Yes, I've tested it many times already. And this solves the problem of incorrectly displaying the message with this configuration: PHP 8.x, PHPmail, S/MIME without DKIM (the same reason: MIME-Version and Content-Type not found)

Synchro:
No, I mean could you add tests to the PHPMailer test suite to ensure it works and prevent regression in future.

anhtunguyen:
@Synchro Same test (environment: php 8.x, PHPmail, S/MIME) :
https://github.com/PHPMailer/PHPMailer/blob/master/examples/smime_signed_mail.phps

Synchro:
I mean the *automated* unit tests [in here](https://github.com/PHPMailer/PHPMailer/blob/master/test/PHPMailerTest.php). What you pointed at is example code which has to be adapted and run manually. Ideally there would be something similar that generates a message, but then verifies it as part of the test suite, which is run and checked every time a commit it made to the repo. Unfortunately there is no PHP-based test suite for DKIM (though I'm working on it), but S/MIME should be easier to write a test for – generate a signed message, then check if its signature matches the key.

src/PHPMailer.php
```diff
@@ -2933,7 +2933,14 @@ public function createBody()
...
                     //The message returned by openssl contains both headers and body, so need to split them up
                     $parts = explode("\n\n", $body, 2);
-                    $this->MIMEHeader .= $parts[0] . static::$LE . static::$LE;
+                    //In $parts[0] there is MIME-Version and Content-Type delimited by "\n".
+                    //Need to separate them, otherwise error with DKIM: "dkim=neutral (body hash did not verify)"
+                    //(The reason is in "h=Date:To:From:Reply-To:Subject:Message-ID:X-Mailer;" without MIME-Version and Content-Type)
+                    $subparts = array_map('trim', explode("\n", $parts[0]));
+                    foreach($subparts as $subpart) {
+                        $this->MIMEHeader .= $subpart . static::$LE;
+                    }
```
promzybankz, line 2942: Ok, this is great here

Synchro: There is already a function to normalise breaks in `static::normalizeBreaks()` which would make for a cleaner implementation here.

kmk142789 (approved):

PR #2347, opened 2021-05-28
Add detection support for Haraka transaction ID

LeonMelis:
A typical SMTP transaction ID for Haraka looks like this:

```
250 Message Queued (14490C56-76FB-4932-A59B-A8299DB2B693.1)
```

This regex will detect and extract this transaction ID


Synchro:
Nice, thanks! I use Haraka on a site.

PR #2363, opened 2021-06-20
GH Actions: start testing against PHP 8.1

jrfnl:
The first alpha of PHP 8.1 has been released, so now seems like a good time to start running the tests against PHP 8.1.

For now, I've configured it to allow builds against PHP 8.1 to fail, while PHP 8.1 is still unstable.

Also: PHPUnit doesn't officially support PHP 8.1 yet, so to install PHPUnit 9.x on PHP 8.1, we need to use `--ignore-platform-reqs`, as otherwise PHPUnit 4.8 would be installed (last PHPUnit version without strict PHP version constraints).

---

**Edit**: I've added a second commit to this PR.

### GH Actions: set error reporting to E_ALL

Turns out the default setting for `error_reporting` used by the SetupPHP action is `error_reporting=E_ALL & ~E_DEPRECATED & ~E_STRICT` and `display_errors` is set to `Off`.

For the purposes of CI, I'd recommend running with `E_ALL` and `display_errors=On` to ensure **all** PHP notices are shown.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -34,7 +34,16 @@ jobs:
       matrix:
         php: ['5.5', '5.6', '7.0.', '7.1', '7.2', '7.3', '7.4', '8.0']
         dependency-version: [prefer-stable]
+        experimental: [false]
+        include:
+          # Experimental builds. These are allowed to fail.
+          - php: '8.1'
+            dependency-version: 'prefer-stable'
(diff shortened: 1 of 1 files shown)


jrfnl:
Turning on `E_ALL` exposed a deprecation notice on PHP 8.1. A fix for this has been pulled in #2370

PR #2366, opened 2021-06-21
Composer: update version constraints for PHPUnit Polyfills

jrfnl:
PHPUnit Polyfills 1.0.0 has been released.

Ref: https://github.com/Yoast/PHPUnit-Polyfills/releases/tag/1.0.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
         "phpcompatibility/php-compatibility": "^9.3.5",
         "roave/security-advisories": "dev-latest",
         "squizlabs/php_codesniffer": "^3.5.6",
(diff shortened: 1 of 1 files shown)


remicollet:
Probably better to keep `^0.2.0 || ^1.0.0`

No reason to bump dependency, until new feature of 1.0.0 are really used.

jrfnl:
Well, it's a dev dependency, so there's no chance of the constraints conflicting with those of projects using this library, which is why I didn't really think that would be necessary.

jesus-khristian (approved):

Synchro:
This needs a tweak to deal with the conflict with the change in #2364

Synchro:
Which I've now done...

PR #2372, opened 2021-06-23
Tests: various improvements

jrfnl:
This is the first PR in a series of PRs to improve the test suite and make it easier to maintain.

### Tests: remove unused test bootstrap file

The PHPUnit config file requires the `vendor/autoload.php` file as the test bootstrap and this `test/bootstrap.php` file is not referenced anywhere in the code base, so this is dead code.

### Tests: apply test method naming conventions

For tests to be picked up by PHPUnit automatically, the method should start with the prefix `test`.
For differently named tests, the `@test` annotation can be used to still mark a method as a test and get PHPUnit to run it.

As the vast majority of tests use the "prefix the method with `test`" convention, this changes the names of the few tests which did not comply with that convention and removes the `@test` annotations.

### Tests: use test skipping where appropriate

In this case, the condition being tested should never be `false`, so could possibly be removed.

All the same, if the condition _would_ result in a `false`, the test would be marked as "risky" as no assertions would be run by it.

This can be avoided by using the condition to set a test skip annotation, instead of wrapping the actual test code in the condition.

### Tests: use strict assertions

PHPUnit contains a variety of assertions and the ones available have been extended hugely over the years.
To have the most reliable tests, the most specific assertion should be used.

Most notably, this changes calls to `assertEquals()` to `assertSame()`, where `assertEquals()` does a loose type comparison `==` and `assertSame()` does a strict type `===` comparison.

The only real exception to this is when comparing two objects, as in that case, the objectID will not be the same, so those should still use `assertEquals()` - or the PHPUnit 9.4.0 `assertObjectEquals()` method for comparing value objects using a callback method in the ValueObject class.

### Tests: use the correct parameter order

For PHPUnit assertions which expect an `$expected` and a `$result` parameter, the parameter order is always `( $expected, $result, ...)`.

While it may not seem important to use the correct parameter order for assertions doing a straight comparison, in actual fact, it is.
The PHPUnit output when the assertions fail expects this order and the failure message will be reversed if the parameters are passed in reversed order which leads to confusion and makes it more difficult to debug a failing test.

### Tests: use static closures

... when the closure doesn't use `$this`.


Synchro:
Thank you so much for all your work on this, really great stuff and much appreciated – and I'm learning a lot too!

jrfnl:
Yvw

PR #2380, opened 2021-06-23
Tests: move tests for the various mail transports to separate file

jrfnl:

Previous PRs in this series: #2372, #2376, #2377, #2378, #2379

---

👉🏻 **_Note: I hope I caught all the relevant tests for this test file. Please let me know if there are any other tests which should be moved to the `MailTransportTest` class._**

### MailTransportTest: various test tweaks

Minor test tweaks:
* Add `@covers` tags ( 👉🏻 **Needs review! - especially the `testMailSend()` method seems to do more than it should**)
* Check if test skipping is necessary at the start of a test method.
* Add "failure message" for each assertion in tests with multiple assertions.
* Tidy up inline comments.


test/PHPMailer/MailTransportTest.php
```diff
@@ -75,20 +82,33 @@ public function testMailSend()
         $this->setAddress('cctestmailsend@example.com', 'cctest', $sType = 'cc');
         $this->setAddress('bcctestmailsend@example.com', 'bcctest', $sType = 'bcc');
         $this->Mail->addReplyTo('replytotestmailsend@example.com', 'replytotest');
-        self::assertContains('testmailsend@example.com', $this->Mail->getToAddresses()[0]);
-        self::assertContains('cctestmailsend@example.com', $this->Mail->getCcAddresses()[0]);
-        self::assertContains('bcctestmailsend@example.com', $this->Mail->getBccAddresses()[0]);
+
+        self::assertContains('testmailsend@example.com', $this->Mail->getToAddresses()[0], 'To address not found');
```
Synchro, line 86: As you say, this whole chunk is testing addressing, not sending through mail(), so should probably move into `testAddressing`

jrfnl: Sorry, didn't get round to changing this yet and the PR has been merged. Will you move these yourself ? Or shall I move them when I address the `testAddressing()` method ?

Synchro: I think I may have underestimated the depths of your ambitions here! If you're going to go into that level of detail, then be my guest, but I'm happy to do it otherwise.

jrfnl: Never underestimate me 😂  It all starts with ambition....

Either way, the idea of this whole exercise is to:
* Make the tests more easily maintainable.
* Get better insight into what code is covered by tests and where additional tests are needed.

To that end, I'm working on:
* Removing the generic `PHPMailerTest` class in favour of individual test classes containing tests for one particular method or a small group of related methods.
* Adding `@covers` tags to all tests.

While doing so and if/when it catches my eye, I also:
* Make sure tests comply with common PHPUnit best practices, like using data providers, correct parameter order, failure messages especially when there are multiple assertions in a test etc.
* Split tests if one test is actually testing multiple things.
* Merge tests and use data providers if they are testing the same thing, but with different inputs.

I suspect I won't be able to finish it all this week, but all the same, made a good start so far I think ?

jrfnl: Oh and that doesn't mean the test suite will be complete after this exercise, more than anything, it will show the gaps in the test suite more clearly and further work is needed after this to address those gaps.

PR #2381, opened 2021-06-24
Tests: move tests related to DKIM to separate file

jrfnl:

Previous PRs in this series: #2372, #2376, #2377, #2378, #2379, #2380

### DKIMTest: stabilize removal of private key file

When an assertion in a test fails, the rest of the code within the test method is no longer executed.
This means that for a test which - like three out of the five DKIM test - creates a file, the test "clean up" via `unlink()` is also no longer executed.
It also means that one test may - unintentionally - interfere with the results of another test due to the file still existing, while it is expected to have been removed.

Aside from that, the `$privatekeyfile` variable used in the various test is the same everywhere, but not consistently used in the tests, as the `'dkim_private.pem'` is hard-coded in multiple places, which decreases maintainability.

This commit fixes both these issues by:
* Declaring a class constant with the target private key file name for use in the test method.
* Implements use of this constant throughout the tests.
* Removes the `unlink()` call from the individual tests, in favour of executing it via the test `tear_down()`, which should still be executed, even when a test has failed.
    Note: as the file also contains two tests which do not create the private key file, but for which the `tear_down()` would also be executed, the `unlink()` call is wrapped in a `file_exists()`.

### DKIMTest: move @group tags

... to the class level and remove them from the individual test functions.

### DKIMTest: add missing @requires tags

Three out of the five tests actually require the `openssl` extension. Only one was so marked.

### DKIMTest: various test tweaks

Minor test tweaks:
* Rename a test to a more specific name to allow for easier test filtering via PHPUnit.
* Add `@covers` tags (Needs review!)
* The `@see` tag is indented for code elements. For external links, the `@link` tag should be used.
* Inline comment punctuation.
* Minor code readability tweaks.

### DKIMTest: add two additional tests

... to cover the "open SSL not available" case.

👉🏻 For these tests to be "hit" in CI, it would be a good idea to have at least one run without the expected extensions. I propose we look at this _after_ finishing the initial test reorganization.


test/PHPMailer/DKIMTest.php
```diff
@@ -79,15 +93,15 @@ public function testDKIMHeaderCanonicalization()
      */
     public function testDKIMOptionalHeaderFieldsCopy()
     {
-        $privatekeyfile = 'dkim_private.pem';
+        $privatekeyfile = self::PRIVATE_KEY_FILE;
```
Synchro, line 110: Any particular reason not to use the constant directly instead of going via a var?

jrfnl: Only for less code churn, but happy to change it.

jrfnl: Done in the applicable commit..

PR #2384, opened 2021-06-24
Tests: move ICal tests to own file

jrfnl:

Previous PRs in this series: #2372, #2376, #2377, #2378, #2379, #2380, #2381, #2382, #2383

### ICalTest: reorganize to use data providers

These three tests were 99% duplicate code. Using a data provider removes the duplication, while the actual tests being run are still 100% the same, including same test count and assertion count.

### ICalTest: add additional tests

... to ensure all valid methods are checked.

Adding these extra tests is a breeze now with the data provider setup.

### ICalTest: various test tweaks

Minor test tweaks:
* Add `@covers` tag. 👈🏻  **_is this correct ?_**
* Minor comment punctuation.


Synchro:
That `@covers` looks correct - this is all about creating the message body

PR #2385, opened 2021-06-24
Tests: move line length detection tests to own file

jrfnl:
This is a next PR in a series of PRs to improve the test suite and make it easier to maintain.
The principle of these changes was proposed to and discussed with the maintainer prior to work being started on it.

I'm splitting this up into several PRs to 1) allow for easier review and 2) allow the repo to benefit from the changes as quickly as possible. (the complete series is still a WIP).

Previous PRs in this series: #2372, #2376, #2377, #2378, #2379, #2380, #2381, #2382, #2383, #2384

### HasLineLongerThanMaxTest: various test tweaks

Minor test tweaks:
* Add `@covers` tag.
* Inline comment punctuation.
* Minor code readability tweaks.

PR #2386, opened 2021-06-25
Tests: move message ID test to own file

jrfnl:

Previous PRs in this series: #2372, #2376, #2377, #2378, #2379, #2380, #2381, #2382, #2383, #2384, #2385

### GetLastMessageIDTest: use the correct parameter order

For PHPUnit assertions which expect an `$expected` and a `$result` parameter, the parameter order is always `( $expected, $result, ...).

While it may not seem important to use the correct parameter order for assertions doing a straight comparison, in actual fact, it is.
The PHPUnit output when the assertions fail expects this order and the failure message will be reversed if the parameters are passed in reversed order which leads to confusion and makes it more difficult to debug a failing test.

### GetLastMessageIDTest: reorganize to use data providers

While initially still only addressing the original test cases, using a data provider here will allow for adding additional test cases more easily.

### GetLastMessageIDTest: add additional tests

Adding additional tests based on the code this is supposed to be testing.

👉🏻  **Note**: I've put the test cases in the "valid"/"invalid" data providers based on the current reality.
Some of the test cases I've added to the "valid" data provider _might_ actually be **_invalid_**. If that's the case, [the regex used on line 2554](https://github.com/PHPMailer/PHPMailer/blob/4ad936e3b5f5bd90ce6e2c08f6ae94a5a3074e22/src/PHPMailer.php#L2554) needs adjusting (or replacing with alternative logic) to account for them (after which the test cases can be moved to the "invalid" data provider).

### GetLastMessageIDTest: add @covers tags

### DIFF
--- a/test/PHPMailer/GetLastMessageIDTest.php
+++ b/test/PHPMailer/GetLastMessageIDTest.php
@@ -17,6 +17,9 @@ use PHPMailer\Test\TestCase;
 
 /**
  * Test setting and retrieving message ID.
+ *
+ * @covers \PHPMailer\PHPMailer\PHPMailer::createHeader
+ * @covers \PHPMailer\PHPMailer\PHPMailer::getLastMessageID
  */
 final class GetLastMessageIDTest extends TestCase
 {
--- a/test/PHPMailer/GetLastMessageIDTest.php
+++ b/test/PHPMailer/GetLastMessageIDTest.php
@@ -22,9 +22,9 @@ final class GetLastMessageIDTest extends TestCase
 {
 
     /**
-     * Test setting and retrieving message ID.
+     * Test setting and retrieving an invalid message ID.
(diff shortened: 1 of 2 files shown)


jrfnl:
> 👉🏻 Note: I've put the test cases in the "valid"/"invalid" data providers based on the current reality.
Some of the test cases I've added to the "valid" data provider might actually be invalid. If that's the case, the regex used on line 2554 needs adjusting (or replacing with alternative logic) to account for them (after which the test cases can be moved to the "invalid" data provider).

@Synchro Should I open a new issue about the above to make sure addressing this will not be forgotten ?

Synchro:
Yes, please do

jrfnl:
Done: https://github.com/PHPMailer/PHPMailer/issues/2388

PR #2387, opened 2021-06-25
Tests: move address splitting tests to own file

jrfnl:

Previous PRs in this series: #2372, #2376, #2377, #2378, #2379, #2380, #2381, #2382, #2383, #2384, #2385, #2386

### Tests/reorganize: move address splitting tests to own file

### ParseAddressesTest: reorganize the testAddressSplitting() test

This commit:
* Adds two new test methods, one for testing address splitting using the PHPMailer native implementation, one for testing address splitting using the IMAP implementation.
    Both these methods use the same data provider.
* Adds a PHPUnit `@requires` tag for the IMAP extension to the IMAP version of the test.
* Adds a type check `assertIsArray()` before doing the content checks on the method return value.
* Changes the tests to do a more detailed `assertSame()` test instead of the previous `assertCount()` and `assert[Not]Empty()` checks.
* Sets up the data provider to have the detailed output arrays for comparison.

The actual test cases in the rewritten test are the same as were previously being tested.

### ParseAddressesTest: replace remaining test by dataprovider entry

The `testImapParsedAddressList_parseAddress_returnsAddressArray()` and the `testImapParsedAddressList_parseAddress_returnsAddressArray_usingImap()` tests were largely duplicates of each other, with the only real difference being the value of the second parameter passed to the `PHPMailer::parseAddresses()` method (`$useimap`).

This commit:
* Removes these two tests in favour of adding the test case to the data provider added in the previous commit.

In effect, the same test case is still being tested just as thoroughly now, just with less (test) code.

### PHPMailer:;parseAddresses(): bug fix

If invalid email addresses are passed to any of the IMAP functions - in this case to `imap_rfc822_parse_adrlist()` - IMAP may generate error notices or warnings.
If nothing is done with these, they will be displayed at PHP shutdown.

The notice will look something like this, for instance:
```
Notice: Unknown: Missing or invalid host name after @ (errflg=3) in Unknown on line 0
```

This bug was exposed by the newly added unit tests. You should be able to see the issue if you run the tests over the previous commit.

As PHPMailer is not interested in invalid addresses, we can just ignore these notices and errors, but we *do* have to _clear_ them to prevent the notices from being thrown.

Refs:
* https://www.php.net/manual/en/function.imap-rfc822-parse-adrlist.php
* https://www.php.net/manual/en/function.imap-errors.php
* ddeboer/imap#308
* https://stackoverflow.com/questions/3378469/how-to-get-rid-of-error-messages-with-phps-imap-fetchstructure

### ParseAddressesTest: reorder the test cases in dataAddressSplitting()

... to a slightly more sane order for humans to follow and see at a glance what is being tested..

### ParseAddressesTest: add @covers tags


jrfnl:
Please note: the bug fix in this PR should probably get a changelog entry.

PR #2398, opened 2021-07-01
TestCase: various tweaks

jrfnl:
Follow up on #2376

### TestCase::setAddress(): add support for ReplyTo

This allows for using the `setAddress()` method in a more consistent manner (where appropriate).

Includes introducing the use of the `setAddress()` function in a few select places.

Note: I do wonder whether this method should ever be used outside of `set_up()` and `tear_down()`, but that is for further discussion and outside the scope of this commit.

### TestCase: add support for initializing PHPMailer with exceptions

The `PHPMailer::__construct()` method has an optional `$exceptions` property to throw exceptions on errors.

Up to now, the `PHPMailer` instance created by the `set_up()` would not pass any parameters, effectively instantiating `PHPMailer` without turning on exceptions.

In a test situation it may be useful for tests to test the behaviour of a method _with_ and _without_ the exceptions option and to test that certain exceptions are thrown and throw the correct message.

With that in mind, I'm introducing a `USE_EXCEPTIONS` class constant to the `TestCase` which can be overloaded in individual test classes and will be used by the `set_up()` method to determine whether it will be instantiated with exceptions or not.

Includes introducing an overload of the class constant in one of the test class for which it would seem appropriate at this time.
Note: this does mean that the `testDKIMSigningMail()` test will show as errored instead of failed if no SMTP connection could be made, but that IMO is the more correct status anyhow.

### TestCase: minor tidying up

... of the inline comments in the `TestCase` file.

### DIFF
--- a/test/PHPMailer/DKIMTest.php
+++ b/test/PHPMailer/DKIMTest.php
@@ -25,6 +25,13 @@ use PHPMailer\Test\TestCase;
 final class DKIMTest extends TestCase
 {
 
+    /**
+     * Whether or not to initialize the PHPMailer object to throw exceptions.
+     *
+     * @var bool|null
+     */
+    const USE_EXCEPTIONS = true;
+
     const PRIVATE_KEY_FILE = 'dkim_private.pem';
 
     /**
@@ -250,7 +257,6 @@ final class DKIMTest extends TestCase
         $this->expectException(Exception::class);
         $this->expectExceptionMessage('Extension missing: openssl');
 
-        $mail = new PHPMailer(true);
-        $mail->DKIM_Sign('foo');
+        $this->Mail->DKIM_Sign('foo');
     }
 }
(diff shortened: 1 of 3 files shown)


test/PHPMailer/DKIMTest.php
```diff
@@ -27,6 +27,13 @@ final class DKIMTest extends TestCase
 
     const PRIVATE_KEY_FILE = 'dkim_private.pem';
 
+    /**
+     * Whether or not to initialize the PHPMailer object to throw exceptions.
+     *
+     * @var bool|null
+     */
+    protected $exceptions = true;
```
Synchro, line 35: It seems very unlikely you'd ever want to change this, and there are unlikely to ever be multiple instances, so would it not be cleaner to make this static?

jrfnl: Good point. What about making it a class constant ? And using it as `static::EXCEPTIONS` ?

Synchro: Sounds good to me.

jrfnl: @Synchro Are you okay with me using "fixup" and force pushing the fixed commit ? That way the change is tidy in the commit history (but it is less easy to see what I changed since you last viewed the PR).

Synchro: Yes

jrfnl: In that case, all done. I've also updated the PR description with the new commit information.

PR #2400, opened 2021-07-02
Tests: move line break normalization tests to own file

jrfnl:

Previous PRs in this series: #2372, #2376, #2377, #2378, #2379, #2380, #2381, #2382, #2383, #2384, #2385, #2386, #2387, #2389, #2392, #2395

### Tests/reorganize: move line break normalization tests to own file

Note: this doesn't move the complete tests from the original test file, just select parts of two test methods.

### NormalizeBreaksTest: reorganize to use data providers

* Maintains (largely) the same test code and test cases.
* Removes code duplication.
* Makes it easier to add additional test cases in the future.

### NormalizeBreaksTest: add additional test case

... to ensure text without line breaks is returned unchanged.

### NormalizeBreaksTest: add extra test

... to verify the behaviour of the `PHPMailer::normalizeBreaks()` method when the `PHPMailer::$LE` property has been customized.

_Note: this test uses the new `TestCase::updateStaticProperty()` method as introduced in #2399._

### NormalizeBreaksTest: add `@covers` tag

### DIFF
--- a/test/PHPMailer/NormalizeBreaksTest.php
+++ b/test/PHPMailer/NormalizeBreaksTest.php
@@ -1,0 +1,56 @@
+<?php
+
+/**
+ * PHPMailer - PHP email transport unit tests.
+ * PHP version 5.5.
+ *
+ * @author    Marcus Bointon <phpmailer@synchromedia.co.uk>
+ * @author    Andy Prevost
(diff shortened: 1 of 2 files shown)


Synchro:
Thanks again!

PR #2410, opened 2021-07-06
Tests: move host validation tests to own file

jrfnl:

Previous PRs in this series: #2372, #2376, #2377, #2378, #2379, #2380, #2381, #2382, #2383, #2384, #2385, #2386, #2387, #2389, #2392, #2395, #2400, #2401, #2404, #2408

### Tests/reorganize: move host validation tests to own file

### IsValidHostTest: reorganize to use data providers

* Split the test into two tests with each a dedicated data provider
* Maintains the same test code and test cases.
* Makes it easier to add additional test cases in the future.

**Note**: the data set description may need some improvement, I've described them based on my best guess of what they were testing.

### IsValidHostTest: minor reordering of data provider

... for easier scanning what's covered reading.

### IsValidHostTest: add additional test cases

... which should be handled correctly based on the code in the method under test.

👉🏻 ~~**_Note: one of the new test case is commented out as it would currently fail, i.e. would be accepted as a valid domain, while AFAICS it is not._**~~
~~This should probably be looked into.~~

With these additional test cases, the method now has 100% code coverage and is actually tested quite fully.

### PHPMailer::isValidHost: minor regex tweak

A `]` is a special character in a regex and should be escaped when the literal is intended. While it was not problematic in this case, it is still recommended to escape it.


Synchro:
I think that commented out case *is* valid, since it would cover `localhost` in particular.

jrfnl:
@Synchro `localhost` is explicitly tested in the `testValidHost()` method.

As far I know, `'a'`, however is not a valid domain. I can uncomment it and move it to the `dataValidHost()` data provider, so at least it's documented that it would pass validation, but I do still wonder whether passing it as valid is desirable.

Synchro:
Good discussion of it here: https://serverfault.com/questions/162038/are-one-letter-host-names-valid

I think they're fine, and they work without problems. Add `127.0.0.1 a` to your hosts file, and ping it:

```
> ping a
PING a (127.0.0.1): 56 data bytes
64 bytes from 127.0.0.1: icmp_seq=0 ttl=64 time=0.091 ms
64 bytes from 127.0.0.1: icmp_seq=1 ttl=64 time=0.105 ms
64 bytes from 127.0.0.1: icmp_seq=2 ttl=64 time=0.135 ms
^C
--- a ping statistics ---
3 packets transmitted, 3 packets received, 0.0% packet loss
round-trip min/avg/max/stddev = 0.091/0.110/0.135/0.018 ms
```

`curl http://a` works fine too.

A separate issue is that you can't use single-character (or any other "dotless" domain) in an email address. But we're not validating email addresses here.

jrfnl:
@Synchro Good point. In that case, let me move the commented out case to the "valid" data provider and uncomment it.

jrfnl:
PR updated, should be good to go once the CI has passed.

PR #2413, opened 2021-07-06
GH Actions: automatically retry postfix install on failure

jrfnl:
The postfix installation step fails regularly, resulting in failed CI builds which have to be restarted, while the failure is not due to anything in the PR.

This commit introduces a new action runner for the postfix install, which will automatically retry the install up to 3 times.

If it works as I expect it to, this should eliminate failed CI builds due to postfix installs erroring out.

Ref: https://github.com/marketplace/actions/retry-step

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -131,10 +131,17 @@ jobs:
         with:
           composer-options: --ignore-platform-reqs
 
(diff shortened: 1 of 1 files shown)


Synchro:
Nice one! This has been annoying!

jrfnl:
Yup. Was starting to annoy me as well, so 🤞🏻 this solves it.

PR #2428, opened 2021-07-08
Localization: allow for more text strings to be translated

jrfnl:
In a number of places in the code base I came across hard-code error messages, which were not set up to be translatable.

For those I found, I've now fixed this.

👉🏻 Note: I've not added the labels used for logging errors via the `PHPMailer:edebug()` method. If so desired, that could be added in a future iteration.

👉🏻 Existing translation files have not been updated. Translators may need to be pinged before the next release to update the translation(s) they maintain.

**[Edit]**
Includes a minor tweak to the PHPCS configuration to allow lines in language files to exceed the "120 chars line max" as using concatenation for text strings in translation files will not work with the current way of loading these files and the new "buggy PHP" message is a long text which would result in nearly all translation files throwing the warning.

### DIFF
--- a/language/phpmailer.lang-nl.php
+++ b/language/phpmailer.lang-nl.php
@@ -7,23 +7,28 @@
  */
 
 $PHPMAILER_LANG['authenticate']         = 'SMTP-fout: authenticatie mislukt.';
+$PHPMAILER_LANG['buggy_php']            = 'PHP versie gededecteerd die onderhavig is aan een bug die kan resulteren in gecorrumpeerde berichten. Om dit te voorkomen, gebruik SMTP voor het verzenden van berichten, zet de mail.add_x_header optie in uw php.ini file uit, gebruik MacOS of Linux, of pas de gebruikte PHP versie aan naar versie 7.0.17+ or 7.1.3+.';
 $PHPMAILER_LANG['connect_host']         = 'SMTP-fout: kon niet verbinden met SMTP-host.';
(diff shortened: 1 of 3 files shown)


Synchro:
There are no translators to ping! All translations have been submitted randomly, and only very rarely have they ever been revised by the same person! This is why I've been hesitant to introduce new translations in the past (because adding one also invites 50 PRs!), and also in light of what I mentioned before about the translations not actually being that useful.

What would be more useful is what was requested in #1579: error codes for every possible problem. They would make translation mapping simpler too. However, it would also be hard to do without a chunky BC break. There is also the issue that we do need to change the format of translation files to something non-executable, like yaml, JSON, po; the current parsing mechanism introduced recently is very much a stop-gap.

I'm happy to merge this anyway as it does improve consistency, but I just though I'd provide some context.

jrfnl:
@Synchro Thanks for the context. When I read through it though, those improvements would all still need a comprehensive "base" language string list and this PR just makes the existing list more complete and the code more consistent regarding localization, so could be seen as a preliminary step.

Re: 50 translation PRs - want me to reduce it by one ? I can adjust the Dutch file in a second commit in this PR if you like.

Synchro:
Sounds good to me!

jrfnl:
Update for the Dutch translation file added + I've added one small extra change to the original commit (PHPCS config tweak).

Synchro:
Thanks again! Good call on the phpcs tweak.

Issue #2266, opened 2021-02-17
function parseAddresses() doesnt decode `name`


stimorol-ice:
# Problem description

Function parseAddresses() doesn't decode names. Please add mb_decode_mimeheader()

# Code to reproduce

I've got to use modified function in wrapper class with added mb_decode_mimeheader() function:

```
public static function parseAddresses($addrstr, $useimap = true)

...
                    if (static::validateAddress($email)) {
                        $addresses[] = [
                            'name' => trim(str_replace(['"', "'"], '', mb_decode_mimeheader($name))),
                            'address' => $email,
                        ];
                    }
...
}
```
# Debug output
In native call:

["To"]=>
    array(1) {
      [0]=>
      array(2) {
        ["name"]=>
        string(68) "=?utf-8?B?0KLQuNC/0L7Qs9GA0LDRhNC40Y8g0L3QsCDQodCy0LjQsdC70L7QstC+?="
        ["address"]=>
        string(24) "print-sviblovo@yandex.ru"
      }

How it's got to be:

["To"]=>
    array(1) {
      [0]=>
      array(2) {
        ["name"]=>
        string(42) "Типография на Свиблово"
        ["address"]=>
        string(24) "print-sviblovo@yandex.ru"
      }

Synchro:
I will need to wrap this in a check for the mbstring extension as it can't work without it, but it's not a *required* extension for PHPMailer otherwise.

PR #2440, opened 2021-07-11
PHPMailer::parseAddresses(): bug fix [1] - extension availability not checked (consistently)

jrfnl:
### PHPMailer::parseAddresses(): bug fix [1] - extension availability not checked

Follow up to #2266

#### Bug fix

In both "arms" (imap vs native implementation) of the `PHPMailer::parseAddresses()` method, the `mb_decode_mimeheader()` function is used to decode a (utf-8) encoded name.

In the IMAP "arm", a check was in place for the Mbstring extension being available before using it. This check was missing from the "native implementation" "arm".

#### Existing Tests

This also means that both currently existing tests have a requirement for the MbString extension being available. This was previously not made explicit in the tests.
Fixed now.

### ParseAddressesTest: add additional test for when Mbstring is not available

What with the previous commit adding a requirement for the `Mbstring` extension to the existing tests, it becomes clear that the "Mbstring extension not available" code path was not covered by the tests.

This commit fixes that by:
* Adding two new test methods which explicitly expect the Mbstring extension to **not** be available.
* Changing the data provider "expected(Imap)" keys.
    The `expected` key in the array will now be an array of arrays.
    A `default` key can be used for when the output across configurations will be the same.
    Any differences across configurations can be provided in separate sub-keys of the `expected` array, using the `native+mbstring`, `imap+mbstring`, `native--mbstring` and/or `imap--mbstring` keys, which match the four test methods which are now in place.
    Additionally, an `native` and/or `imap` key can be used for setting the output expectations for the two native implementation or the two IMAP implementation tests, if the Mbstring extension makes no difference.

### 🆕  PHPMailer::parseAddresses(): replace use of `extension_loaded()`

The `extension_loaded()` PHP function is sometimes disabled by shared hosts as a form of "security by obscurity".
This would lead to a fatal "Function not available" error.

With that in mind, replacing the `extension_loaded()` checks in the `PHPMailer::parseAddresses()` function with a check for the `MB_CASE_UPPER` constant being defined. This _should_ be reliable enough to determine whether the Mbstring extension is available, though has its own drawbacks as when the extension is not available, the constant _could_ be polyfilled by userland code (and sometimes is).

The `MB_CASE_UPPER` constant was chosen as it is one of the few Mbstring constants which is actually available cross-version in PHP 5.5-current.

Refs:
* https://www.php.net/manual/en/mbstring.constants.php
* https://php-legacy-docs.zend.com/manual/php5/en/mbstring.constants

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -1237,7 +1237,7 @@ class PHPMailer
                     $name = trim($name);
                     if (static::validateAddress($email)) {
                         //If this name is encoded, decode it
-                        if (preg_match('/^=\?.*\?=$/', $name)) {
+                        if (extension_loaded('mbstring') && preg_match('/^=\?.*\?=$/', $name)) {
                             $name = mb_decode_mimeheader($name);
                         }
                         $addresses[] = [
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -1205,7 +1205,8 @@ class PHPMailer
                     //Decode the name part if it's present and encoded
                     if (
                         property_exists($address, 'personal') &&
-                        extension_loaded('mbstring') &&
+                        //Check for a Mbstring constant rather than using extension_loaded, which is sometimes disabled
+                        defined('MB_CASE_UPPER') &&
                         preg_match('/^=\?.*\?=$/', $address->personal)
                     ) {
                         $address->personal = mb_decode_mimeheader($address->personal);
@@ -1236,8 +1237,9 @@ class PHPMailer
                     $email = trim(str_replace('>', '', $email));
                     $name = trim($name);
                     if (static::validateAddress($email)) {
+                        //Check for a Mbstring constant rather than using extension_loaded, which is sometimes disabled
                         //If this name is encoded, decode it
-                        if (extension_loaded('mbstring') && preg_match('/^=\?.*\?=$/', $name)) {
+                        if (defined('MB_CASE_UPPER') && preg_match('/^=\?.*\?=$/', $name)) {
                             $name = mb_decode_mimeheader($name);
                         }
                         $addresses[] = [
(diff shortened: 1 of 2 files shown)


src/PHPMailer.php
```diff
@@ -1237,7 +1237,7 @@ public static function parseAddresses($addrstr, $useimap = true)
                     $name = trim($name);
                     if (static::validateAddress($email)) {
                         //If this name is encoded, decode it
-                        if (preg_match('/^=\?.*\?=$/', $name)) {
+                        if (extension_loaded('mbstring') && preg_match('/^=\?.*\?=$/', $name)) {
```
Synchro, line 1240: It's worth avoiding `extension_loaded` because it's sometimes disabled in shared hosting environments. It's safer to check for defined constants, for example using `defined('MB_CASE_UPPER')`. See the same thing is done for detecting OpenSSL.

jrfnl: Agreed. (Though seriously what are they thinking ? Security by obscurity is not security.)
Either way, in the same function, same thing, but in the different "arm" of the logic, it is done in the same way using `extension_loaded`, so for consistency, I'd say that in that case *both* need to be changed.

https://github.com/PHPMailer/PHPMailer/blob/5cbd8f9f62ffe03f120b733c50c7bdc4b655b929/src/PHPMailer.php#L1205-L1212

Note: I did already consider changing to a `function_exists()` originally, but as the second bugfix to be pulled after this one introduces the use of another Mbstring function, `function_exists()` did not seem appropriate (unless used twice as individual functions can also be disabled by hosts via the `disable_functions` ini setting).

Synchro: Oh I know – there is little about shared hosting that's not crappy! Yes, other instances of it should be changed too. Only thing to look out for is in what PHP versions the constants you check for were introduced, as these checks have to work all the way back to 5.5.

jrfnl: > Yes, other instances of it should be changed too.

Will do.

> ... these checks have to work all the way back to 5.5.

Fully aware of that, that is what the upcoming bugfix 2 is about (this function currently does not work on PHP 5.5).

jrfnl: I've added an extra commit to this PR to fix this up.

Issue #2419, opened 2021-07-07
Suggestion for potential improvement for `PHPMailer::getTranslations()`


jrfnl:
Encountered while writing tests for the logic involved in localizing the error messages.

**_Note: I'd be happy to make the changes I'm suggesting here. Just wanted to allow for discussion before sending in a PR._**

The `public` `PHPMailer::getTranslations()` method is defined as such:
https://github.com/PHPMailer/PHPMailer/blob/6dc419aeadf21d99b8c12e8e139383c394355784/src/PHPMailer.php#L2281-L2289

... with the `$language` property defaulting to an empty array.
https://github.com/PHPMailer/PHPMailer/blob/6dc419aeadf21d99b8c12e8e139383c394355784/src/PHPMailer.php#L696

This means that when no call has been made to `PHPMailer::setLanguage()`, the `PHPMailer::getTranslations()` method will return an empty array.

This feels inconsistent with the handling of this in the (`protected`) `PHPMailer::lang()` method, which automatically sets the language strings to the default language if the translations array is empty:
https://github.com/PHPMailer/PHPMailer/blob/6dc419aeadf21d99b8c12e8e139383c394355784/src/PHPMailer.php#L4046-L4050

I'd like to suggest that the `PHPMailer::getTranslations()` method should always return an array with the text strings and if no call has been made to `PHPMailer::setLanguage()` before the call to `PHPMailer::getTranslations()`, this array will contain the default English language strings.

So basically, I'm proposing to change the `PHPMailer::getTranslations()` method to this:
```php
    public function getTranslations()
    {
        if (empty($this->language)) {
            $this->setLanguage(); // Set the default language.
        }

        return $this->language;
    }
```

Opinions ?

Synchro:
Indeed I don't see any advantage in returning that empty array, so your suggestion seems very sensible.

PR #2452, opened 2021-07-13
 PHPMailer::setLanguage(): various fixes + add tests for localization

### LINKED ISSUE #2418: Suggestions for potential improvements for `PHPMailer::setLanguage()`
While writing tests for the logic involved in localizing the error messages, I came across a number of oddities which _may_ benefit from some tweaks to improve things.

**_Note: I'd be happy to make the changes I'm suggesting here. Just wanted to allow for discussion before sending in a PR._**

### Observation 1: Language code is case-sensitive
```php
// This works:
$mail->setLanguage('nl');

// This doesn't:
$mail->setLanguage('NL');
```

In a number of other places in the library, the code is very deliberately case-tolerant.

#### Proposal

It would only be a small tweak to make the `setLanguage()` method case-tolerant for the `$langCode` passed and may improve usability.

### Observation 2: ISO 639-1 multi-level: no fall back to "parent" language when available
```php
// This works:
$mail->setLanguage('nl');

// This doesn't:
$mail->setLanguage('nl_NL'); // "PHP" way of writing the code
$mail->setLanguage('nl-NL'); // Iso way of writing the code
```

For a few languages, there are multiple translation files available for variants of the language. For most languages this is not the case.

#### Proposal

When the `$langCode` is passed as "multi-level" and no translation is found for the specified variant, the localization currently falls back to English, while when a translation _is_ available in the "parent" language, it may be more appropriate to fall back to that translation.

So, when `nl_NL` would be passed, `nl` would be used when `nl_NL` can't be found and `nl` **is** available. The fall back to English would then only happen when neither `nl_NL` nor `nl` is available.

### Observation 3: ISO 639-1 multi-level: underscore versus dash

While for this library, underscore separators for multi-level language files are expected, the documentation talks about ISO 639-1 language codes, which, when multi-level, are expected to use a dash as separator.

#### Proposal

It would only be a small change to support both for the `$langCode` input, but again, may improve usability a little.

### Observation 4: Available language file `sr_latn` can never be used

There is a language file available for `sr_latn` (Serbian, latin script). However, this language code would never pass the regex validation of the `$langCode`, so this file is effectively useless.

#### Proposal

Either remove the translation file or adjust the regex to allow for loading it.

jrfnl:
Fixes #2418 (and could only be pulled no after #2449 was merged as otherwise the tests would fail on PHP 5.5)

### PHPMailer::setLanguage(): make return value consistent

The return value of the method was inconsistent as it could return `true` even when the requested language was not loaded/set.

Previously, the method would:
* Return `false` when a `$langcode` was matched against the regex, but the file couldn't be loaded.
* Return `true` in all other cases, including when a `$langcode` other than `'en'` was passed, but it didn't match the regex, which meant that effectively the requested language was ignored and English was loaded anyway, but the function would still return `true`.

This has now been changed to:
* Return `true` when the requested language was loaded, whether `'en'` or another language.
* Return `false` when the requested language wasn't loaded and the language defaulted to English.

Related to #2418 (https://github.com/PHPMailer/PHPMailer/issues/2418#issuecomment-876265100)

### PHPMailer::setLanguage(): match language codes case-insensitively

For "language code - country code" locale notations, it is common for the "country code" part to be provided in uppercase, like `pt_BR`.

The method currently did not allow for that correctly. The regex check would accept the language code, but then fail to find the file on *nix based systems as the file names are all in lowercase and non-Windows file systems are generally case-sensitive. Which means that in effect, the method, ended up falling back to the default language (English).

The change in this commit makes the handling of the provided `$langcode` case-tolerant.

Note: the language code used in the file names for the language files is still expected to always be lowercase, including for language files in custom paths!

Related to #2418 - observation 1

### PHPMailer::setLanguage(): fix "lang_country" codes do not fall back to "lang"

When the  `$langCode` is passed as "language code - country code" and no translation is found for the specified country variant, the localization would automatically fall back to English, even when there was a viable translation available in the "parent" language, i.e. just based on the _language code_.

This commit changes the logic in the `PHPMailer::setLanguage()` method for when a "lang-country" code is passed.
It will now try and find a "lang-country" file first, if not found, it will try to find a file for just the "lang" and only if that could also not be found, it will fall back to English.

Related to #2418 - observation 2

Includes using named subpatterns in the regex to make the regex self-documenting.

Includes removing a commented out line of code which is superseded anyway.

### PHPMailer::setLanguage(): allow for "lang(_script)?(_country)?" codes

Language codes which were "language-script" based were not accepted by the regex used, which meant that the `sr_latn` script could never be loaded.

After discussion in #2418, it was decided to support "script" in a language code and to support it like so:
```
2-character language code [_] optional 4-character script code [_] optional 2-character country code
```

This combines the annotation forms of the following known standards:
* https://unicode-org.github.io/cldr-staging/charts/37/summary/root.html
* https://docs.oracle.com/cd/E23824_01/html/E26033/glset.html
* http://www.loc.gov/standards/iso639-2/php/code_list.php

This means that all of the below codes will now pass the language code validation:
```
sr
sr_latn
sr_rs
sr_latn_rs
```

But not:
```
sr_rs_latn
```

This commit applies the above change and also adjusts the "language code fall back" logic to take language codes with a script code into account.
Note: if the requested full "language-script-country" code is not available, "language-country" will take precedence over "language-script" for the fallback logic to find the appropriate translation file.

Related to #2418 - observation 4

### Tests: introduce dedicated tests for text localization

So far, the methods related to text localization - `PHPMailer::setLanguage()`, `PHPMailer::getTranslations()` and `PHPMailer::lang()` - did not have dedicated tests.

The test file this commit introduces, tests all aspects of these methods, including the changes introduced in response to #2418 and #2419, as well as documents the current behaviour of the methods for specific, outlier situations.

### PHPMailerTest: remove setLanguage() test

... which wasn't actually testing anything at all.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -2251,19 +2251,32 @@ class PHPMailer
         //Validate $langcode
         $foundlang = true;
         $langcode  = strtolower($langcode);
-        if (!preg_match('/^[a-z]{2}(?:_[a-z]{2})?$/', $langcode) && $langcode !== 'en') {
+        if (!preg_match('/^(?P<lang>[a-z]{2})(?P<country>_[a-z]{2})?$/', $langcode, $matches) && $langcode !== 'en') {
             $foundlang = false;
             $langcode = 'en';
         }
 
-        $lang_file = $lang_path . 'phpmailer.lang-' . $langcode . '.php';
         //There is no English translation file
         if ('en' !== $langcode) {
-            //Make sure language file path is readable
-            if (!static::fileIsAccessible($lang_file)) {
+            $langcodes = [];
+            if (!empty($matches['country'])) {
+                $langcodes[] = $matches['lang'] . $matches['country'];
+            }
+            $langcodes[] = $matches['lang'];
+
+            //Try and find a readable language file for the requested language.
+            $foundFile = false;
+            foreach ($langcodes as $code) {
+                $lang_file = $lang_path . 'phpmailer.lang-' . $code . '.php';
+                if (static::fileIsAccessible($lang_file)) {
+                    $foundFile = true;
+                    break;
+                }
+            }
+
+            if ($foundFile === false) {
                 $foundlang = false;
             } else {
-                //$foundlang = include $lang_file;
                 $lines = file($lang_file);
                 foreach ($lines as $line) {
                     //Translation file lines look like this:
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -2189,6 +2189,8 @@ class PHPMailer
      * The default language is English.
      *
      * @param string $langcode  ISO 639-1 2-character language code (e.g. French is "fr")
+     *                          Optionally, the language code can be enhanced with a 4-character
+     *                          script annotation and/or a 2-character country annotation.
      * @param string $lang_path Path to the language file directory, with trailing separator (slash).D
      *                          Do not set this from user input!
      *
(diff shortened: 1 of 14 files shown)


Synchro:
Great stuff, as always! I knew about named groups in PCRE, but it never occurred to me that they turned up in the `$matches` array!

jrfnl:
@Synchro Well, now you know. 😀

I really like the feature for a couple of reasons:
* Keeping the key reference in `$matches` stable (numeric indexes can change when the complexity of the regex changes)
* The self-documenting part, names are just that much more descriptive when used in `$matches` compared to the numeric indexes.

This regex felt very much like typical place which would really benefit from using them.

PR #2459, opened 2021-07-28
Simplify use of get_oauth_token.php

rx80:
Add a form the user can fill in, avoiding the need to modify clientId and clientSecret in code.


Synchro:
Thanks

PR #2473, opened 2021-08-17
Fixed check for proper email format rather than just an @ character

jsokol:
The email address is not properly validated here.  The code only checks for the existence of an "@", but this functionality can be abused.  This allows adding an "?attach=<filepath>" suffix to the email address which can be used to include files on the operating system (e.g. ~/.ssh/id_rsa).  As a result, sensitive information can be included in these emails and sent to an attacker.  While the applications including PHPMailer should be doing input validation on their side of things, there's also no reason why we shouldn't be doing a more comprehensive check for a proper email address in PHPMailer, since we have the ability to do so.

Quick read about the vulnerability is available here:
https://www.zdnet.com/article/some-email-clients-are-vulnerable-to-attacks-via-mailto-links/

Full original paper reporting this vulnerability is available here:
https://www.nds.ruhr-uni-bochum.de/media/nds/veroeffentlichungen/2020/08/15/mailto-paper.pdf

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -1078,7 +1078,7 @@ class PHPMailer
         $address = trim($address);
         $name = trim(preg_replace('/[\r\n]+/', '', $name)); //Strip breaks and trim
         // Validate the proper email address format
-        if (!filter_var($email, FILTER_VALIDATE_EMAIL)) {
+        if (!filter_var($address, FILTER_VALIDATE_EMAIL)) {
           $error_message = sprintf(
               '%s (%s): %s',
               $this->lang('invalid_address'),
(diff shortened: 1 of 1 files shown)


Synchro:
This is not needed, because all calls to this function eventually go through [`addAnAddress`](https://github.com/PHPMailer/PHPMailer/blob/master/src/PHPMailer.php#L1131), which applies proper validation. PHPMailer already has considerable structure around address validation, and all validation calls go through the [`validateAddress`](https://github.com/PHPMailer/PHPMailer/blob/master/src/PHPMailer.php#L1345) method for consistency. The reason for the simple check at this point is that some addresses cannot be validated until we know what character set is in use (due to IDN), and that is not determined until you actually ask it to send, but one that is that obviously broken can be rejected immediately. PHPMailer doesn't concern itself with mailto URLs, as they are beyond its scope, but the line of defence that we do have is that a mailto URL can never be a valid email address.

If you actually attempt the attack you linked to, you should find that it fails.

jsokol:
Thanks for the quick response, Marcus, and based on what you've said I agree that it's not necessary.  

Thanks for working on PHPMailer!  It's a great platform that simplifies the process of sending emails in PHP.

PR #2501, opened 2021-09-16
5.2 stable : support EAI Mail specific to English and Thai lang

titipong7:
edit validateAddress for validate email in English charactor and Thai charector


Synchro:
Sorry, this isn't the right way to do this. PHPMailer doesn't support SMTPUTF8, so you can't use non-ASCII characters in the local part anyway, and non-ASCII characters in the domain are already handled using IDN.

Secondly, you've done this on the 5.2 branch which is no longer supported.

PR #2513, opened 2021-09-28
PHP 5.4 compatibility

Sebijk:
this merge request will fix the Fatal error: Can't use function return value in write context error, because php 5.4 don't support ini_get in an if condition.

Why i make this patch? CentOS7 ship PHP on the version 5.4 and still support this with security updates until 2024.


Synchro:
If you need PHP 5.4 compatibility, you should be using the legacy 5.2 branch, though it's no longer maintained. I would expect many other things to break in PHP 5.4 as well, and the test suite would need to include 5.4, so there is a lot more to do than just this change. To that end, you might be better off maintaining your own fork or patch set ( I know that's common in packaging systems), since I have no particular desire to support such old versions; 5.5 is bad enough!

PR #2306, opened 2021-04-09
Create phpmailer.lang-mgl.php

wispas:
Translated to Mongolian language

Before submitting your pull request, check whether your code adheres to PHPMailer coding standards (which is mostly [PSR-12](https://www.php-fig.org/psr/psr-12/)) by running [PHP CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer):

    ./vendor/bin/phpcs

Any problems reported can probably be fixed automatically by using its partner tool, PHP code beautifier:

    ./vendor/bin/phpcbf

PR #2516, opened 2021-10-01
++++ phpmailer.lang-mgl.php

codehero7386:


Synchro:
If this is Mongolian, please use the ISO 639-1 code for it, which is `mn`, not `mgl`.

Synchro:
Just noticed that this is a direct copy of #2306, sharing exactly the same problem!

Issue #2488, opened 2021-09-02
Clean up after sending the POP3 QUIT command.


wpkc:
Please check these things before submitting your issue:

- [x] *Read the error message* you're seeing - it often tells you what is wrong, and may contain useful links & instructions
- [x] Make sure you're using the latest version of PHPMailer
- [x] Check that your problem is not dealt with in [the troubleshooting guide](https://github.com/PHPMailer/PHPMailer/wiki/Troubleshooting), especially if you're having problems connecting to Gmail or GoDaddy
- [x] Include sufficient code to reproduce your problem
- [x] If you're having an SMTP issue, include the debug output generated with `SMTPDebug = 2` set
- [x] If you have a question about how to use PHPMailer (rather than reporting a bug in it), tag a question on Stack Overflow with `phpmailer`, but [**search first**](http://stackoverflow.com/questions/tagged/phpmailer)!

# Problem description
The POP3::disconnect() method in version 6.5.1 does not clean up after sending the QUIT command and closing the socket.

# Code to reproduce
This code...
```
    public function disconnect()
    {
        $this->sendString('QUIT');
        //The QUIT command may cause the daemon to exit, which will kill our connection
        //So ignore errors here
        try {
            @fclose($this->pop_conn);
        } catch (Exception $e) {
            //Do nothing
        }
    }
```
Should be something like this...
```
   public function disconnect() {
      $this->sendString('QUIT');

      // RFC 1939 shows POP3 server sending a +OK response to the QUIT command.
      // Try to get it.  Ignore any failures here.
      try {
            $this->getResponse();
      } catch (Exception $e) {
            //Do nothing
      }

      //The QUIT command may cause the daemon to exit, which will kill our connection
      //So ignore errors here
      try {
            @fclose($this->pop_conn);
      } catch (Exception $e) {
            //Do nothing
      }

      // Clean up attributes.
      $this->connected = false;
      $this->pop_conn  = false;
   }
```
# Debug output
n/a

Synchro:
Could you wrap this and #2487 into a PR?

yuxiang03:
Are you sure?

PR #2519, opened 2021-10-04
POP3 cleanup

### LINKED ISSUE #2487: POP3::login() should immediately return false if not connected.
Please check these things before submitting your issue:

- [x] *Read the error message* you're seeing - it often tells you what is wrong, and may contain useful links & instructions
- [x] Make sure you're using the latest version of PHPMailer
- [x] Check that your problem is not dealt with in [the troubleshooting guide](https://github.com/PHPMailer/PHPMailer/wiki/Troubleshooting), especially if you're having problems connecting to Gmail or GoDaddy
- [x] Include sufficient code to reproduce your problem
- [x] If you're having an SMTP issue, include the debug output generated with `SMTPDebug = 2` set
- [x] If you have a question about how to use PHPMailer (rather than reporting a bug in it), tag a question on Stack Overflow with `phpmailer`, but [**search first**](http://stackoverflow.com/questions/tagged/phpmailer)!

# Problem description
POP3::login() should immediately return false if not connected.  The login method will _eventually_ fail and return false.  But it really should return false immediately after the call to POP3::setError().

# Code to reproduce
Current code in POP3.php version 6.5.1
```
    public function login($username = '', $password = '')
    {
        if (!$this->connected) {
            $this->setError('Not connected to POP3 server');
        }
```

Should be...
```
   public function login($username = '', $password = '')
    {
        if (!$this->connected) {
            $this->setError('Not connected to POP3 server');
            return false;
        }
```
# Debug output
n/a

miken32:
This PR resolves #2487 and #2488 by adding the suggested code. If there's a response from the server to the `QUIT` command it will be captured, and the `connected` property will properly be `false` after a disconnect.

### DIFF
--- a/src/POP3.php
+++ b/src/POP3.php
@@ -308,6 +308,7 @@ class POP3
     {
         if (!$this->connected) {
             $this->setError('Not connected to POP3 server');
+            return false;
         }
         if (empty($username)) {
             $username = $this->username;
--- a/src/POP3.php
+++ b/src/POP3.php
@@ -337,6 +337,15 @@ class POP3
     public function disconnect()
     {
         $this->sendString('QUIT');
+        
+        // RFC 1939 shows POP3 server sending a +OK response to the QUIT command.
+        // Try to get it.  Ignore any failures here.
+        try {
+            $this->getResponse();
+        } catch (Exception $e) {
+            //Do nothing
+        }
+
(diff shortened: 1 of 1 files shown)


Synchro:
Thanks, it's easy when someone else writes the code for you!

miken32:
All part of getting that T-shirt lol

Synchro:
I suspected as much!

Issue #1405, opened 2018-04-04
Improve feedback on TLS errors


Synchro:
By far the most common problem with PHPMailer is those running into issues with TLS, particularly the problem caused by gmail updating its certs to one signed with a relatively new CA that is not included in many older OS CA cert bundles. The errors generated when this happens are not especially clear, and though PHPMailer always provides a link to the troubleshooting guide which describes these issues and fixes for them in detail, that seems to be too difficult for many. It would be better to improve the error reporting at the point they occur, so for example if an SMTPS connection to port 465 fails, that it retrieves the underlying error and shows that, rather than a generic "connect failed". The messages generated via exceptions are slightly better in this area, but the regular error handling could do with improving to give better feedback.

madbananaman:
Would setting SMTPAutoTLS = false as default be a viable option?

Or potentially echoing a message for users as feedback to include:
$mail->SMTPAutoTLS = false;

Synchro:
Not really. It would defeat the entire point of the setting. The idea is to work as well as possible with well-configured servers (which is most of them). If you have a defective server, it’s up to you to deal with it, not to degrade everyone else.

That’s not to say that error messages can’t be improved, as per my original comment.

Letsgetitons:
How much is the deposit

PR #2526, opened 2021-10-11
Provide additional error details after connection failures

miken32:
This should go some way to resolving #1405. TLS connection errors are already being gathered by the SMTP class but these details are not being included in the exception thrown [when `$this->smtp->startTls()` fails](https://github.com/PHPMailer/PHPMailer/blob/master/src/PHPMailer.php#L2146). This PR adds those details.

Unless I'm missing something, it seems like an exception is never thrown when `$this->smtp->connect()` returns false and exceptions are enabled. I don't know if that's a deliberate decision, but this PR also throws an exception with these details in that case.

Now TLS failure exceptions will have a message like
```
SMTP Error: Could not connect to SMTP host. Connection failed. stream_socket_enable_crypto(): SSL operation failed with code 1. OpenSSL Error messages:
error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed
```
Or
```
SMTP Error: Could not connect to SMTP host. Connection failed. stream_socket_enable_crypto(): Peer certificate CN=`smtp.example.com' did not match expected CN=`mail.example.com'
```

This is a bit verbose, but parsing all possible error messages (esp. from all versions of PHP going back almost 10 years) into something readable seems like an unrealistic job. Certainly a subset could be translated into something more user-friendly, perhaps within the `SMTP::errorHandler()` method.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -2143,7 +2143,8 @@ class PHPMailer
                     }
                     if ($tls) {
                         if (!$this->smtp->startTLS()) {
-                            throw new Exception($this->lang('connect_host'));
+                            $message = $this->getSmtpErrorMessage('connect_host');
+                            throw new Exception($message);
                         }
                         //We must resend EHLO after TLS negotiation
                         $this->smtp->hello($hello);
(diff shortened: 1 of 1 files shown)


9564519:
This software was used to send a phishing

PR #2528, opened 2021-10-12
fr_FR update

Machou:
I corrected the spaces with this one: U+202F

I don't understand the problem, can you please explain?


language/phpmailer.lang-fr.php
```diff
@@ -9,24 +9,30 @@
...
-$PHPMAILER_LANG['variable_set']         = 'Impossible d\'initialiser ou de réinitialiser une variable : ';
-$PHPMAILER_LANG['extension_missing']    = 'Extension manquante : ';
+$PHPMAILER_LANG['recipients_failed']    = 'Erreur SMTP : les destinataires suivants ont échoué : ';
+$PHPMAILER_LANG['signing']              = 'Erreur de signature : ';
+$PHPMAILER_LANG['smtp_code']            = 'Code SMTP : ';
+$PHPMAILER_LANG['smtp_code_ex']         = 'Informations supplémentaires SMTP : ';
+$PHPMAILER_LANG['smtp_connect_failed']  = 'La fonction SMPT connect() a échouée.';
+$PHPMAILER_LANG['smtp_detail']          = 'Détails : ';
+$PHPMAILER_LANG['smtp_error']           = 'Erreur du serveur SMTP : ';
+$PHPMAILER_LANG['variable_set']         = 'Impossible d’initialiser ou de réinitialiser une variable : ';
```
williamdes, line 37: I do not like the silent change done from an ASCII character to a non ASCII one.

cc @Synchro

Synchro: I'm not bothered by that – other characters in the same sentence are not ASCII either.

williamdes: Okay, If you have more information to share about using an ASCII `'` or not that would be awesome. For now I try to stick to `'` as it has less chances to break stuff. Was just concerned that this change done silently would break other workflows

Thanks for your kind response

Synchro: In PHP (and most languages), `'` and `"` have special meanings, which is why the original translation needed the `'` to be escaped. The curly apostrophe has no such meaning (it's treated like any other letter), so it doesn't need escaping, and doesn't represent any particular risk. There are other substitutions which might cause problems via "homoglyph" attacks (perhaps the "fullwidth apostrophe": `＇` Unicode: U+FF07, UTF-8: EF BC 87), but I don't think this is one of them.

As of PHP 6.5.0, the format of these translation files effectively changed – while they are formatted as PHP, they are actually only ever read as plain text files using a very simple parser. This was done to avoid security issues - PHP files of any description can contain executable code, and that's not something we want to have lurking in translation files in case an attacker manages to upload one and trick an app into running it. At some point I'll come up with a better format, preferably one that does not require external libraries to read reliably (e.g. YAML, INI), and convert all the translations to use that directly.

PR #2564, opened 2021-11-19
Fix #2563: Provide the right flags to IDN methods for email addresses conversion

### LINKED ISSUE #2563: Provide the right flags to IDN methods for email addresses conversion
PHP Mailer converts the email addresses domain to A-label as it does not implement SMTPUTF8 flag but this conversion is not fully IDNA2008 compliant.
The following flags should be provided to idn_to_ascii convertion: ` IDNA_DEFAULT | IDNA_USE_STD3_RULES | IDNA_CHECK_BIDI | IDNA_CHECK_CONTEXTJ | IDNA_NONTRANSITIONAL_TO_ASCII` when using `INTL_IDNA_VARIANT_UTS46` variant.
[This line](https://github.com/PHPMailer/PHPMailer/blob/e84b657160c16fbf3b582cae3d3ccb06d49120c1/src/PHPMailer.php#L1454) should be updated.
A test case would be to provide an email address with domain  fußball.test that should be converted to xn--fuball-cta.test instead of fussball.test.

See [Unicode IDNA compatibility processing](https://www.unicode.org/reports/tr46/)

j-bernard:


Synchro:
Thanks!

PR #2578, opened 2021-12-08
Add smtp_transaction_id pattern for Mailjet

Fonata:
Responses from the Mailjet servers look like this:  
`250 OK queued as a08b7f37-ed61-423a-ba75-38ed2fd4b54c`


Synchro:
Thanks

PR #2582, opened 2021-12-16
Do not call `call_user_func` on `error_log`

### LINKED ISSUE #2581: Debug to `error_log` with `SMTP::DEBUG_SERVER` raises PHP warning
# Problem description
Debugging to the`error_log` when `SMTPDebug` is set to `SMTP::DEBUG_SERVER` (`2`) ends with warning:

```Warning: error_log(): TCP/IP option not available!```

This is because the `2` value is passed to the `error_log()` which is the [`message_type`](https://www.php.net/manual/en/function.error-log.php#refsect1-function.error-log-parameters) where 

>  2 | No longer an option.

See the code:
https://github.com/PHPMailer/PHPMailer/blob/a127e25118a653a60742f909416b5d8fe399dc30/src/PHPMailer.php#L904-L908

# Code to reproduce
Just the Simple Example plus code below:

```php
$mail->SMTPDebug = SMTP::DEBUG_SERVER;
$mail->Debugoutput = 'error_log';
```

# Debug output
```
Warning: error_log(): TCP/IP option not available!
```

This could be easily done by removing the `error_log` from the `in_array()` check, as it would be handled by the switch below.

```php
if (is_callable($this->Debugoutput) && !in_array($this->Debugoutput, ['html', 'echo'])) {
    call_user_func($this->Debugoutput, $str, $this->SMTPDebug);

    return;
}
```

vokiel:
Do not call `call_user_function` when `Debugoutput` is set to `error_log`  
as it raises PHP warning when `SMTPDebug` is set to `SMTP::DEBUG_SERVER`.

Closes #2581

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -901,7 +901,7 @@ class PHPMailer
             return;
         }
         //Avoid clash with built-in function names
-        if (is_callable($this->Debugoutput) && !in_array($this->Debugoutput, ['error_log', 'html', 'echo'])) {
+        if (is_callable($this->Debugoutput) && !in_array($this->Debugoutput, ['html', 'echo'])) {
             call_user_func($this->Debugoutput, $str, $this->SMTPDebug);
 
             return;


vokiel:
Mistake, not needed.

PR #2583, opened 2021-12-18
Composer: update version constraints

jrfnl:
### Composer: update version constraints

Follow up on c1a63391084a9cb55bba53cb341049428aa3a1c0 which enabled PHP 8.1 support, let's make sure that the version constraints of (dev) dependencies are such that PHP 8.1 compatible versions will be installed.

Refs:
* https://github.com/php-parallel-lint/PHP-Parallel-Lint/releases
* https://github.com/squizlabs/PHP_CodeSniffer/releases

### CS: minor cleanup

PHPCS 3.6.2 added a sniff for a PSR-12 rule which was previously not strictly checked: "No blank line after the opening brace of a class".

This fixes the newly flagged issues.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,10 +35,10 @@
         "dealerdirect/phpcodesniffer-composer-installer": "^0.7.0",
         "doctrine/annotations": "^1.2",
         "php-parallel-lint/php-console-highlighter": "^0.5.0",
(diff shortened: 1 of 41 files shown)


Synchro:
Thanks Juliette!

jrfnl:
Your welcome, next one coming up.

PR #2588, opened 2021-12-22
Update phpmailer.lang-fr.php

nogaruki:
The word "bogue" is less used than bug. 
Both of them are correct, but "bug" is more used.


Synchro:
Merci!

PR #2590, opened 2021-12-24
GH Actions: version update for `ramsey/composer-install`

jrfnl:
The action used to install Composer packages and handle the caching has released a new major (and some follow-up patch releases), which means, the action reference needs to be updated to benefit from it.

Refs:
* https://github.com/ramsey/composer-install/releases/tag/2.0.0
* https://github.com/ramsey/composer-install/releases/tag/2.0.1
* https://github.com/ramsey/composer-install/releases/tag/2.0.2


Synchro:
Thanks again, now go and do something Christmasy!

Issue #1702, opened 2019-04-02
No support for AMP4Email


hiburn8:
PHPMailer doesnt have any way to control the MIME structure of the email, so its not possible to add arbitrary MIME sections like 'text/x-amp-html'. While i dont really like Google's attempt to tighten their grip on the content of people's emails, I think PHPMailer should add support for AMP-emails as AMP4email as PHPMailer is basically the driver for a lot of the mailing-list industry, and AMP is the cool new thing.

NiklasBr:
For precisely this reason I think it should *not* be added :) If we don't like that Google tightens its grip, why should we help them?

It's easy enough to extend the base PHPMailer class and add your own custom/proprietary mail mime constant:
```php
class AMPMailer extends PHPMailer
{
...
        const CONTENT_TYPE_X_AMP = 'text/x-amp-html';
...

**Comment by Synchro** [comment]
Indeed. AMP on the web is dying out and I've no particular desire to follow Google's lead, after all they have inflicted upon us with gmail. Pull requests welcome of course.

In the longer term, I'd really like to get rid of all the preset MIME structure stuff as it's a right mess, and that would make implementing things like this easier. I'd really like to implement support for SMTPUTF8 as I do not know of any implementations of that for PHP at all.

**Comment by tiwarishikesh** [comment]
Create a  public $AMP_Body class PHPMailer

```
public function createBody(){
...
          case '*':
         $body .= static::$LE;
                if(!empty($this->AMP_Body)){
                    $body .= '--'.$this->boundary[1].static::$LE;
                    $body .= 'Content-Type: '.static::CONTENT_TYPE_AMP_HTML.'; charset='.static::CHARSET_UTF8.static::$LE;
                    $body .= $this->encodeString($this->AMP_Body, $bodyEncoding);
                    $body .= static::$LE;
                }
                $body .= $this->getBoundary($this->boundary[1], $bodyCharSet, 
                static::CONTENT_TYPE_TEXT_HTML, $bodyEncoding);
                $body .= $this->encodeString($this->Body, $bodyEncoding);

...
}
```

jab1000:
[...]

Issue #2329, opened 2021-04-28
Amp4Email support - Google's 2 cents to next-gen emails


slavino:
https://github.com/PHPMailer/PHPMailer/blob/0acc57db14c444fe25f479583da74ac675c0c161/src/PHPMailer.php#L945

Details here: https://github.com/ikkez/f3-mailer/issues/4

https://amp.dev/documentation/guides-and-tutorials/contribute/integrate-amphtml-email/?referrer=ampproject.org

Synchro:
Not going to happen. AMP for both web and email is a cynical attempt by google to push everyone into using their proprietary protocols, and in the many years it's been available it has proved only that it takes traffic and attention *away* from those who use it, and as a result it's becoming much less popular. If you really want it, you can implement it yourself by setting content types manually and/or subclassing, I don't think you need any underlying changes in PHPMailer.

slavino:
You must be right - it's just like Google's cynical OS called Android... Who on Earth would ever use it?

Synchro:
As I said, PHPMailer doesn't do anything to stop you using it, it's just not going to go out of its way to help you use it. AMP is simply terrible technology. If Google really wanted to do something of benefit to the world of email, they could start by making gmail good.

PR #2607, opened 2022-01-17
Support for AMP mails

martinvitak:
Pull request for issue #1702. We use this feature for sending AMP mails.


Synchro:
Generally speaking, [like many others[(https://duckduckgo.com/?q=why+is+google+amp+bad&t=osx&ia=web), I'm quite against AMP for both web and email. It's a failing standard that has done little more than give Google yet more opportunities to abuse their position – see #2329 for a previous take. As such, I don't really want it in PHPMailer, though of course you can put the code you have (which looks quite tidy, though it will not support messages with attachments, and lacks tests) into a child class to retain compatibility with upstream releases.

PR #2641, opened 2022-03-05
GH Actions: version update for various predefined actions

jrfnl:
A number of predefined actions have had major release, which warrant an update to the workflow(s).

These updates don't actually contain any changed functionality, they are mostly just a change of the Node version used by the action itself (from Node 14 to Node 16).

Refs:
* https://github.com/actions/checkout/releases


Synchro:
Thank you!

jrfnl:
You're welcome ;-)

PR #2663, opened 2022-04-05
GH Actions: version update for codecov action runner

jrfnl:
Yet another predefined action has had a major release.

This is, again, mostly just a change of the Node version used by the action itself (from Node 12 to Node 16).

Refs:
* https://github.com/codecov/codecov-action/releases

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -163,7 +163,7 @@ jobs:
 
       - name: Send coverage report to Codecov
         if: ${{ success() && matrix.coverage == true }}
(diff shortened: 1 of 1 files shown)


jrfnl:
> Thank you again Juliette! Would this kind of update be spotted by dependabot or similar? It would be nice to automate these things.

Yes, I've recently started looking into this and was running some tests still. I can send you a PR with a config if you like.

Synchro:
That would be great, yes please!

jrfnl:
Coming up and I saw there was another suggested change I'd not applied here yet either, so will pull that too.

PR #2665, opened 2022-04-07
Composer script: show deprecations when linting

jrfnl:
While rare, there are some deprecations which PHP can show when a file is being linted.
By default these are ignored by PHP-Parallel-Lint.

Apparently though, there is an option to show them (wasn't documented until recently), so let's turn that option on.


jrfnl:
Oops... looks like I forgot to push the second commit. Will open a separate PR for that now.

PR #2688, opened 2022-05-31
Either validate hostnames correctly or not at all.  Also: cannot do Codecov if fork

### LINKED ISSUE #2684: isValidHost: don't try to validate hostnames
https://github.com/PHPMailer/PHPMailer/blob/bf99c202a92daa6d847bc346d554a4727fd802a5/src/PHPMailer.php#L4093 is simply a bad practice. You're trying to validate hostnames with a regex that doesn't conform to all past or future RFCs. It's good to validate against the empty string, and to classify ip4, ipv6, etc, but the regex here doesn't conform to valid hostnames used outside of DNS. And there are plenty of reasons to use hostnames that are not used with DNS, for instance, docker hosts, internal or unroutable hosts, etc.

otheus:
Let the host's DNS resolution system determine if the host can be mapped or not.  For instance, UTF-8 extended  characters will be mapped into RFC-compliant host/domain names, (see RFC 5890  to 5894)... `öbb.at -> xn-bb-eka.at`. So unless we intend to handle all possible cases here, just pass it on the the OS.  Besides, RFC 952 et al only specify what hostnames are valid for hosts in the DNS system. On other networks and, those restrictions are not necessarily valid, such as kubernetes clusters where pseudo-hostnames refer to ephemeral containers. Fixes #2684


otheus:
"Much as you might like this library to be purist and abstract, it's not even vaguely close to that, and doesn't aim to be."
Sorry if I made myself unclear: YOUR LIBRARY BREAKS COMMON USE-CASES IN THE REAL WORLD.  I didn't spend all that time trying to build this project because of an _abstract case_. I did it because I need it to work in the real world and in a use case that is increasingly common. 

Is there a compromise here? A configuration variable perhaps? 

And FYI, the product that this library is being used in simply ignores the exceptions, so all your bound checking has the result that the end user _never knows_ the email didn't work. A bounce message would be far superior to that. Or do you have a clever way of forcing app developers to handle exceptions?

PR #2702, opened 2022-06-16
regular callable call

MathiasReker:
Callables must be called without using call_user_func* when possible.


jrfnl:
Actually, this is not necessarily a good pattern for callables you do not control.

The pattern you are using was one of the reasons this recent [PHP RFC](https://wiki.php.net/rfc/partially-supported-callables-expand-deprecation-notices) was needed.

PR #2703, opened 2022-06-16
visibility required

MathiasReker:
Visibility MUST be declared on all properties and methods; abstract and final MUST be declared before the visibility; static MUST be declared after the visibility.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -31,34 +31,34 @@ namespace PHPMailer\PHPMailer;
  */
 class PHPMailer
 {
-    const CHARSET_ASCII = 'us-ascii';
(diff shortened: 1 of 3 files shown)


Synchro:
Judging by your string of PRs, you seem unaware that this has to run on PHP 5.5.

MathiasReker:
I am sorry about that. I have overlooked this.

Synchro:
The CI runs will show you where the problems are, should you want to persevere, but this will unfortunately apply to a lot of your PRs!

MathiasReker:
> Judging by your string of PRs, you seem unaware that this has to run on PHP 5.5.

Yes, you are right! I overlooked this.

PR #2704, opened 2022-06-16
list syntax

MathiasReker:
List (array destructuring) assignment should be declared using the configured syntax.


MathiasReker:
not compatible with PHP 5.5, I think

jrfnl:
Correct, short lists were only [introduced in PHP 7.1](https://wiki.php.net/rfc/short_list_syntax).

PR #2705, opened 2022-06-16
modernize strpos

MathiasReker:
Replace strpos() calls with str_starts_with() or str_contains() if possible.


Synchro:
Nope, not compatible with old PHP.

PR #2707, opened 2022-06-16
simplified if return

MathiasReker:
Simplify if control structures that return the boolean result of their condition.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -4105,12 +4105,12 @@ class PHPMailer
             //Is it a valid IPv4 address?
             return filter_var($host, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4) !== false;
         }
-        if (filter_var('http://' . $host, FILTER_VALIDATE_URL) !== false) {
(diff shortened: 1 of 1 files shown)


src/PHPMailer.php
```diff
@@ -4105,12 +4105,12 @@ public static function isValidHost($host)
             //Is it a valid IPv4 address?
             return filter_var($host, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4) !== false;
         }
-        if (filter_var('http://' . $host, FILTER_VALIDATE_URL) !== false) {
+        return (bool) (filter_var('http://' . $host, FILTER_VALIDATE_URL) !== false)
```
jrfnl, line 4108:
```suggestion
        return (filter_var('http://' . $host, FILTER_VALIDATE_URL) !== false);
```

If you are returning the outcome of the comparison, it will already be boolean, so no need for the type cast.

(though you do need a `;`  at the end as this is now a parse error).

jrfnl (changes requested):

Synchro (commented):
Cleanup

PR #2708, opened 2022-06-16
use arrow functions

MathiasReker:
Anonymous functions with one-liner return statement must use arrow functions.


Synchro:
Sorry, no arrow functions in old PHP!

PR #2709, opened 2022-06-16
void return

MathiasReker:
Add void return type to functions with missing or empty return statements, but priority is given to @return annotations.


MathiasReker:
not compatible with PHP 5.5

PR #2711, opened 2022-06-16
explicit string variable

MathiasReker:
Converts implicit variables into explicit ones in double-quoted strings or heredoc syntax.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -1815,7 +1815,7 @@ class PHPMailer
 
         if (
             escapeshellcmd($string) !== $string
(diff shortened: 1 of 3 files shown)


Synchro:
Isn't this syntax deprecated in PHP 8.2?

MathiasReker:
I don't think so. It's a part of the Symfony CS: https://cs.symfony.com/doc/rules/string_notation/explicit_string_variable.html

IMO its better practice.

Do you have a link to any documentation about this? :)

MathiasReker:
Nice catch! Thanks

PR #2712, opened 2022-06-16
integer literal case

MathiasReker:
Integer literals must be in correct case.


Synchro:
What coding standard is this from?

MathiasReker:
> What coding standard is this from?

https://cs.symfony.com/doc/rules/casing/integer_literal_case.html

jrfnl:
PHP supports this in a case-insensitive manner, so this is an unnecessary change.

Synchro:
There are quite a few things in the Symfony standard that I don't like, and this project isn't using it. It makes no functional difference (as Juliette says), so I don't see this as adding value.

PR #2714, opened 2022-06-18
ISO 639-1 Language code and content do not match

ZhangChengLin:
ISO_639-1 https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes

The content of this file is Chinese Simplified, not the Chamorro language corresponding to ISO_639-1, and the author's site address involves online gambling.

Later, it is hoped that a more comprehensive and authoritative language code standard can be adopted.


ZhangChengLin:
@Synchro

Maikuolan:
If it's just a problem regarding the language code used, and not a problem regarding the content of the translation, wouldn't it be better to just rename it to the correct language code instead of deleting it entirely?

ZhangChengLin:
> If it's just a problem regarding the language code used, and not a problem regarding the content of the translation, wouldn't it be better to just rename it to the correct language code instead of deleting it entirely?

Translations of Chinese already exist, including Chinese Simplified (zh_cn) and Chinese Traditional (en).

Maikuolan:
> Translations of Chinese already exist, including Chinese Simplified (zh_cn) and Chinese Traditional (en).

Ah, I didn't notice before. Makes sense then. Thank you for clarifying. :-)

Synchro:
Thanks for spotting this

PR #2723, opened 2022-06-29
Verification is required

mortalys:


mortalys (commented):
I got an error, accidental pass an object and notice that string was not getting validated

Synchro:
Tests are just code too.

PR #2727, opened 2022-07-07
Automatic text/csv detection

Cleptomania:
Per [RFC 7111](https://www.rfc-editor.org/rfc/rfc7111) the `text/csv` MIME type should be used for CSV files.

This just adds detection for the `csv` file extension to use this in the automatic type detection.


Synchro:
Nice catch, thanks

PR #2731, opened 2022-07-18
small code-styling

rsgrinko:
replace quotes to ', replace \n to PHP_EOL const, replace foreach to implode()

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -547,7 +547,7 @@ class PHPMailer
      * Value can be any php callable: http://www.php.net/is_callable
      *
      * Parameters:
-     *   bool $result        result of the send action
+     *   bool    $result        result of the send action
      *   array   $to            email addresses of the recipients
      *   array   $cc            cc email addresses
      *   array   $bcc           bcc email addresses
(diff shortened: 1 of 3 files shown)


src/PHPMailer.php
```diff
@@ -866,14 +866,14 @@ private function mailPassthru($to, $subject, $body, $header, $params)
         //Calling mail() with null params breaks
         $this->edebug('Sending with mail()');
         $this->edebug('Sendmail path: ' . ini_get('sendmail_path'));
-        $this->edebug("Envelope sender: {$this->Sender}");
-        $this->edebug("To: {$to}");
-        $this->edebug("Subject: {$subject}");
-        $this->edebug("Headers: {$header}");
+        $this->edebug('Envelope sender: {' . $this->Sender . '}');
```
Synchro, line 869: This is not a correct substitution. `{` has special meaning when used as in the original code. It's also doing nothing for readability.

Synchro (commented):
There's just too much wrong in here to be worth reviewing or cleaning up. I recommend starting again – but only a couple of changes are possibly worth keeping so I don't know that I would bother!

src/PHPMailer.php
rsgrinko: I'll correct it now

rsgrinko: fixed

Synchro:
I suggest you finish off cleaning up and then squash and force-push so that we're not left with a messy commit history. That's why I suggested starting again as it's probably quicker!

PR #2745, opened 2022-08-10
Update PHPDoc for PHP 7.2+ `mail()`

schulerj89:
As of PHP 7.2 the `$additional_headers` param in `mail()` also takes an array. Updated the PHPDoc and the debugging.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -847,11 +847,11 @@ class PHPMailer
      * claims to be sendmail), don't pass params (not a perfect fix,
      * but it will do).
      *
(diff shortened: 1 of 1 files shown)


Synchro:
I don't think this is appropriate. The `$headers` param submitted to this function is _always_ a a string, regardless of the PHP version used, and PHPMailer's only supports that as a consistent interface. Allowing an array here would imply that all upstream callers also need to be able to handle the headers as an array, and that's just unnecessary. It's also a private method so is not accessible to any external callers anyway.

Issue #2753, opened 2022-08-21
Invalid DKIM-signature with only BCC-addresses


sverpoorten:
Hi! 

Today I upgraded from version 6.0.1 to 6.6.3 of PHPMailer. This corrected some previous issues, but there is a new one.
The case:
- Using PHP mail() by applying `isMail()`
- Applying DKIM-signature
- Multiple BCC-addressees
- No To or CC-addressees

**This results an invalid DKIM-signature.**

What happens is the following:
1. In `preSend()` line 1607 the To-addressee `undisclosed-recipients:;` is added to the headers in `$this->mailHeader`
2. This is used to construct the DKIM-signature (in particular for the `z`-field)
3. Next in `mailSend()` the variable `$to` will be empty since there are no To-addressees and the email will be sent as such

This results in the following example headers (with invalid DKIM-signature):
> Received: by mailserver (Postfix, from userid 33) id C53F3520B8D; Sat, 20 Aug 2022 14:00:55 +0000 (UTC)
> **To:** 
> Subject: Testmail
> [...]
> DKIM-Signature: v=1; [...] h=Date:From:Reply-To:Message-ID:X-Mailer:MIME-Version:Content-Type:To:Subject; z=Date:Sat,=2020=20Aug=202022=2016:00:55=20+0200 |[...]**To:undisclosed-recipients:=3B** |Subject:Testmail;[...]

I was able to correct the behaviour by adding the following line after line 1894 in `mailSend()`:
`if($to == "") $to = "undisclosed-recipients:;";`

This results in the following example headers (with valid DKIM-signature):
> Received: by mailserver (Postfix, from userid 33) id CE9AE520036; Sun, 21 Aug 2022 08:45:47 +0000 (UTC)
> **To: undisclosed-recipients:;**
> Subject: Testmail
> [...]
> DKIM-Signature: v=1; [...] h=Date:From:Reply-To:Message-ID:X-Mailer:MIME-Version:Content-Type:To:Subject; z=Date:Sun,=2021=20Aug=202022=2010:45:47=20+0200 |[...]|**To:undisclosed-recipients:=3B** |Subject:Testmail;[...]

Synchro:
Good catch. Could you please wrap that up in a PR?

sverpoorten:
Review-ready :) See PR #2754

PR #2754, opened 2022-08-21
Bugfix invalid DKIM-signature with zero To-addresses

sverpoorten:
Resolves issue #2753 

When using PHP's `mail()` function, and when there are no To-addressees (e.g. only BCC), then the DKIM-signature was not correct. There was a discrepancy between how `mail()` was called (empty `$to`) versus the DKIM-signature (`'undisclosed-recipients:;'`).


Synchro:
Thanks

Issue #2626, opened 2022-02-08
Mail will not send, no error message is displayed


trebr:
`<?php
use PHPMailer\PHPMailer\PHPMailer;
use PHPMailer\PHPMailer\Exception;

require 'PHPMailer-master/src/Exception.php';
require 'PHPMailer-master/src/PHPMailer.php';
require 'PHPMailer-master/src/SMTP.php';

// I installed using the zip file, so I am not using autoloader

$mail = new PHPMailer;
$mail->isSMTP(); 
$mail->Host = "smtp.office365.com";
$mail->SMTPDebug = 3; 
$mail->Port = 587; 
$mail->SMTPSecure = 'tls'; 
$mail->SMTPAuth = true;
$mail->Username = "[ my Exchange email address ]";
$mail->Password = "[ my Exchange password ]";
$mail->setFrom("[  my Exchange email address ]", "[ my name ]");
$mail->addAddress("[ recipient email ]", " [ recipient name ]");
$mail->Subject = 'PHPMailer SMTP test';
$mail->msgHTML("test body");
$mail->AltBody = 'HTML messaging not supported';
if(!$mail->send()){
    echo "Mailer Error: " . $mail->ErrorInfo;
}else{
    echo "Message sent!";
}
?>`

The only output is:

> 2022-02-08 14:36:21 Connection: opening to smtp.office365.com:587, timeout=300, options=array()

There is no other output whatsover.

Other notes:

- I am using PHPMailer version 6.5.3
- I checked my outgoing SMTP connection from my local server at https://testconnectivity.microsoft.com/tests/exchange and no errors were reported.
- The exchange account I am sending from does not have multifactor authentication enabled, so I do not need to use an app password.
- I checked the login details for my email address.
- I was using this code without issue until last Friday.

Synchro:
This looks like a straightforward SMTP block by your hosting provider, who can change their policies at any time. It's not possible to use those Microsoft tools to troubleshoot your *outbound* connection; you can test using the advice in the PHPMailer troubleshooting guide though, using telnet and openssl.

trebr:
> 

I should have made it clear that I was raising two issues:
1. a potential issue with the SMTP (which you've addressed, so thank you for that)
2. the fact no errors were reported

If I set
    $mail->SMTPDebug
to either 1 or 2 there is no output at all. If I set it to 3, I only get the output quoted in my issue: the email does not send yet no error is reported.

[...]

Issue #2570, opened 2021-11-29
$this->smtp->error info is overwritten after SMTP connect fail, useful error msg discarded


neekfenwick:
# Problem description

PHPMailer 6.5.3.

When I accidentally mistype the host or port number, the useful error message that is created gets discarded from the ErrorInfo data before PHPMailer returns from Send().

The sequence of events is:

- Create PHPMailer instance with Host = 'mistaken-smtp.gmail.com', everything else normal and try to send an email.
- smtpConnect() calls $this->smtp->connect
- in class.smtp.php the connect() function sets a custom errorHandler: set_error_handler(array($this, 'errorHandler'));
- The call to fsockopen() fails because of the mistyped hostname
- errorHandler is invoked, calls $this->setError($notice, $errno, $errmsg) with a useful errMsg "stream_socket_client(): php_network_getaddresses: getaddrinfo failed: Name or service not known"
- Because the connection failed, getSMTPConnection() then calls setError itself (line 414), overwriting the error object set by the errorHandler but still has some useful information about what caused the connection failure, and returns false.
- Because $this->smtp->connect returned false, PHPMailer.smtpConnect skips the workhorse body of the function and skips to line 1727 where it calls $this->smtp->close() and will return false.
- The call to $this->smtp->close causes the smtp class to complete wipe out its error object with $this->setError(''); .. it's here that any useful error messages get completely lost.
- Because smtpConnect() returned false, PHPMailer.smtpSend throws a phpmailerException on line 1558 which is caught in postSend() on line 1352 and PHPMailer.setError is called with the exception's message, which by now is a rather unuseful "SMTP connect() failed. https://github.com/PHPMailer/PHPMailer/wiki/Troubleshooting"
- PHPMailer.setError does try to inspect $this->smtp->getError() on line 3273 instead of using the plain $msg passed in, but because the close() earlier wiped out the error details in the smtp object, PHPMailer.setError simply sets ErrorInfo to the unhelpful error string from the exception.

# Code to reproduce

```php
<?php
require('vendor/autoload.php');
use PHPMailer\PHPMailer\PHPMailer;
use PHPMailer\PHPMailer\Exception;

$mail = new PHPMailer();
$mailerErrors = [];
$mail->Debugoutput = function ($str, $level) use (&$mailerErrors) {
    $mailerErrors[] = [ 'str' => $str, 'level' => $level ];
};
$mail->SMTPDebug = 3;

$mail->IsSMTP();
[...]

PR #2787, opened 2022-10-03
fix bug error message supressed on close()

dazsim:
fixes error message getting suppressed when the connection is closed.

https://github.com/PHPMailer/PHPMailer/issues/2626
https://github.com/PHPMailer/PHPMailer/issues/2570


Synchro:
Thanks

Issue #2788, opened 2022-10-04
Documentation improvement: How to use Microsoft OAuth with PHPMailer 


Synchro:
Now that Microsoft have discontinued their support for simple username/password authentication, many people are trying to switch to use XOAUTH2 instead. Unfortunately this appears to be a complete nightmare, as discussed in #2747, #2116, #1959, #1901, etc. A wiki article detailing how to set it up would be much appreciated by many of PHPMailer's users.

greew:
Here we go!

https://github.com/PHPMailer/PHPMailer/wiki/Microsoft-Azure-and-XOAUTH2-setup-guide

I'll have a Pull Request in a few minutes with additions for get_oauth_token.php and an azure example file :)

Synchro:
Thanks for all your effort on this. Hopefully Microsoft will make it easier and more reliable at some point, but until then...

wayzz1992:
Anyone else having problems with GuzzleHttp :
Fatal error: Uncaught GuzzleHttp\Exception\RequestException: cURL error 60: SSL certificate problem: unable to get local issuer certificate (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for https://login.microsoftonline.com/tenant-id/oauth2/v2.0/token

I am trying to get the RefreshToken on my Local machine before testing on a dev environment. Using same code as the example.
$token = $provider->getAccessToken(
        'authorization_code',
        [
            'code' => $_GET['code']
        ]
    );

PR #2793, opened 2022-10-07
Added Azure XOAUTH2 help and examples

greew:
I hope it's quite self explanatory :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,6 +48,7 @@
     },
     "suggest": {
         "ext-mbstring": "Needed to send email in multibyte encoding charset or decode encoded addresses",
+        "greew/oauth2-azure-provider": "Needed for Microsoft Azure XOAUTH2 authentication",
         "hayageek/oauth2-yahoo": "Needed for Yahoo XOAUTH2 authentication",
         "league/oauth2-google": "Needed for Google XOAUTH2 authentication",
         "psr/log": "For optional PSR-3 debug logging",
--- a/examples/azure_xoauth2.phps
+++ b/examples/azure_xoauth2.phps
@@ -1,0 +1,126 @@
+<?php
+
+/**
+ * This example shows how to send via Microsoft Outlook's servers using XOAUTH2 authentication
+ * using the league/oauth2-client to provide the OAuth2 token.
+ * To use a different OAuth2 library create a wrapper class that implements OAuthTokenProvider and
+ * pass that wrapper class to PHPMailer::setOAuth().
+ */
+
+//Import PHPMailer classes into the global namespace
+use PHPMailer\PHPMailer\PHPMailer;
+use PHPMailer\PHPMailer\SMTP;
+use PHPMailer\PHPMailer\OAuth;
+//Alias the League Google OAuth2 provider class
+use Greew\OAuth2\Client\Provider\Azure;
+
+//SMTP needs accurate times, and the PHP time zone MUST be set
+//This should be done in your php.ini, but this is how to do it if you don't have access to that
+date_default_timezone_set('Etc/UTC');
+
+//Load dependencies from composer
+//If this causes an error, run 'composer install'
+require '../vendor/autoload.php';
+
+//Create a new PHPMailer instance
+$mail = new PHPMailer();
+
+//Tell PHPMailer to use SMTP
+$mail->isSMTP();
+
+//Enable SMTP debugging
+//SMTP::DEBUG_OFF = off (for production use)
+//SMTP::DEBUG_CLIENT = client messages
+//SMTP::DEBUG_SERVER = client and server messages
+$mail->SMTPDebug = SMTP::DEBUG_SERVER;
+
+//Set the hostname of the mail server
+$mail->Host = 'smtp.office365.com';
+
+//Set the SMTP port number:
+// - 465 for SMTP with implicit TLS, a.k.a. RFC8314 SMTPS or
+// - 587 for SMTP+STARTTLS
+$mail->Port = 587;
+
+//Set the encryption mechanism to use:
+// - SMTPS (implicit TLS on port 465) or
+// - STARTTLS (explicit TLS on port 587)
+$mail->SMTPSecure = PHPMailer::ENCRYPTION_STARTTLS;
+
+//Whether to use SMTP authentication
+$mail->SMTPAuth = true;
+
+//Set AuthType to use XOAUTH2
+$mail->AuthType = 'XOAUTH2';
+
+//Start Option 1: Use league/oauth2-client as OAuth2 token provider
+//Fill in authentication details here
+//Either the microsoft account owner, or the user that gave consent
+$email = 'someone@somemicrosoftaccount.com';
+$clientId = 'RANDOMCHARS-----duv1n2TS';
+$clientSecret = 'RANDOMCHARS-----lGyjPcRtvP';
+$tenantId = 'RANDOMCHARS-----HSFTAOIlagss';
+
+//Obtained by configuring and running get_oauth_token.php
+//after setting up an app in Google Developer Console.
+$refreshToken = 'RANDOMCHARS-----DWxgOvPT003r-yFUV49TQYag7_Aod7y0';
+
+//Create a new OAuth2 provider instance
+$provider = new Azure(
+    [
+        'clientId' => $clientId,
+        'clientSecret' => $clientSecret,
+        'tenantId' => $tenantId,
+    ]
+);
+
+//Pass the OAuth provider instance to PHPMailer
+$mail->setOAuth(
+    new OAuth(
+        [
+            'provider' => $provider,
+            'clientId' => $clientId,
+            'clientSecret' => $clientSecret,
+            'refreshToken' => $refreshToken,
+            'userName' => $email,
+        ]
+    )
+);
+//End Option 1
+
+//Option 2: Another OAuth library as OAuth2 token provider
+//Set up the other oauth library as per its documentation
+//Then create the wrapper class that implementations OAuthTokenProvider
+$oauthTokenProvider = new MyOAuthTokenProvider(/* Email, ClientId, ClientSecret, etc. */);
+
+//Pass the implementation of OAuthTokenProvider to PHPMailer
+$mail->setOAuth($oauthTokenProvider);
+//End Option 2
+
+//Set who the message is to be sent from
+//For Outlook, this generally needs to be the same as the user you logged in as
+$mail->setFrom($email, 'First Last');
+
+//Set who the message is to be sent to
+$mail->addAddress('someone@someserver.com', 'John Doe');
+
+//Set the subject line
+$mail->Subject = 'PHPMailer Outlook XOAUTH2 SMTP test';
+
+//Read an HTML message body from an external file, convert referenced images to embedded,
+//convert HTML into a basic plain-text alternative body
+$mail->CharSet = PHPMailer::CHARSET_UTF8;
+$mail->msgHTML(file_get_contents('contentsutf8.html'), __DIR__);
+
+//Replace the plain text body with one created manually
+$mail->AltBody = 'This is a plain-text message body';
+
+//Attach an image file
+$mail->addAttachment('images/phpmailer_mini.png');
+
+//send the message, check for errors
+if (!$mail->send()) {
+    echo 'Mailer Error: ' . $mail->ErrorInfo;
+} else {
+    echo 'Message sent!';
+}
--- a/get_oauth_token.php
+++ b/get_oauth_token.php
@@ -44,6 +44,8 @@ use League\OAuth2\Client\Provider\Google;
 use Hayageek\OAuth2\Client\Provider\Yahoo;
 //@see https://github.com/stevenmaguire/oauth2-microsoft
 use Stevenmaguire\OAuth2\Client\Provider\Microsoft;
+//@see https://github.com/greew/oauth2-azure-provider
+use Greew\OAuth2\Client\Provider\Azure;
 
 if (!isset($_GET['code']) && !isset($_POST['provider'])) {
     ?>
@@ -57,11 +59,14 @@ if (!isset($_GET['code']) && !isset($_POST['provider'])) {
     <label for="providerYahoo">Yahoo</label><br>
     <input type="radio" name="provider" value="Microsoft" id="providerMicrosoft">
     <label for="providerMicrosoft">Microsoft</label><br>
+    <input type="radio" name="provider" value="Azure" id="providerAzure">
+    <label for="providerAzure">Azure</label><br>
     <h1>Enter id and secret</h1>
     <p>These details are obtained by setting up an app in your provider's developer console.
     </p>
     <p>ClientId: <input type="text" name="clientId"><p>
     <p>ClientSecret: <input type="text" name="clientSecret"></p>
+    <p>TenantID (only relevant for Azure): <input type="text" name="tenantId"></p>
     <input type="submit" value="Continue">
 </form>
 </body>
@@ -77,18 +82,22 @@ session_start();
 $providerName = '';
 $clientId = '';
 $clientSecret = '';
+$tenantId = '';
 
 if (array_key_exists('provider', $_POST)) {
     $providerName = $_POST['provider'];
     $clientId = $_POST['clientId'];
     $clientSecret = $_POST['clientSecret'];
+    $tenantId = $_POST['tenantId'];
     $_SESSION['provider'] = $providerName;
     $_SESSION['clientId'] = $clientId;
     $_SESSION['clientSecret'] = $clientSecret;
+    $_SESSION['tenantId'] = $tenantId;
 } elseif (array_key_exists('provider', $_SESSION)) {
     $providerName = $_SESSION['provider'];
     $clientId = $_SESSION['clientId'];
     $clientSecret = $_SESSION['clientSecret'];
+    $tenantId = $_SESSION['tenantId'];
 }
 
 //If you don't want to use the built-in form, set your client id and secret here
@@ -127,6 +136,17 @@ switch ($providerName) {
             'scope' => [
                 'wl.imap',
                 'wl.offline_access'
+            ]
+        ];
+        break;
+    case 'Azure':
+        $params['tenantId'] = $tenantId;
+
+        $provider = new Azure($params);
+        $options = [
+            'scope' => [
+                'https://outlook.office.com/SMTP.Send',
+                'offline_access'
             ]
         ];
         break;


greew:
Fixes #2788

Synchro:
Bravo, thank you very much. I'm sure that many will appreciate your considerable efforts here. I've pinged a few people who were running into Azure problems to see if they will give it a test before I merge it.

greew:
I forgot to add the example file - it has been committed now as well :)

Synchro:
I didn't receive any feedback, however, you've done a great job here so I'm merging it anyway – if anyone has issues to report, I'm sure they will turn up eventually!

greew:
Well, Packagist says 25 installs overall, and I may have the first 5-10 during development.. so at least 15 people have started using it, and if neither you nor I have heard anything yet, let's hope it works well 😀

adrianalumne:
Hello, I have been behind this topic of sending via **PHPMailer**, **SMTP** and **Microsoft OAuth** for a while without success.

I recently found your guide on how to correctly configure **Azure Active Directory** for sending emails and after carefully following your guide and testing the code I ran into several errors and it never works.

If I follow all the steps and keep this scope _'https://outlook.office.com/SMTP.Send_' it doesn't give me any type of error, just this:

```
Auth method requested: XOAUTH2

Auth methods available on the server: LOGIN,XOAUTH2

CLIENT -> SERVER: QUIT
```

So I continued to investigate and found that I could use this scope: '_https://graph.microsoft.com/SMTP.Send_' but it returns the following error:

`AUTH command failed: 535 5.7.3 Authentication unsuccessful`

What is this error due to? Is it a misconfiguration of Azure? Can outlook and office365 accounts send SMTP through Azure? How are those scopes different? I have also seen some scopes similar to this 'wl.smtp_send'. What does wl mean? Are they outdated?

I know there are a lot of questions but we have been looking for a solution for several months and we have only found your guide. Thanks greetings!

PS: The PHP code **is similar** to the one you use in your guide

Synchro:
I suspect you may only have `SMTPDebug = 1` set, so you're only seeing client-side messages. Set it to 3 and see if you get some more clues.

decomplexity:
What permissions have you set in AAD against Graph resource? You need **SMTP.Send**.
And your client scope should indeed be **https://outlook.office.com/SMTP.Send** (preceded by **offline_access** to get a refresh token)

greew:
@adrianalumne Both @Synchro and @decomplexity have already given some great pointers.

Give it a go - if it still fails, I'll be happy to see if I can help you further.

adrianalumne:
Hello again! Thanks for your help @greew , regarding @Synchro's comment, we have changed `SMTPDebug = 4` to `SMTPDebug = 3` and this is the log it returns:

image

I don't really see **any error messages** and everything seems to be working fine. 

It is true that by putting the scope mentioned by @decomplexity (`https://outlook.office.com/SMTP.Send`) the error that some messages indicated above **no** **longer appears** (`AUTH command failed: 535 5.7.3 Authentication unsuccessful`)

These are the permissions set in AAD:

image

Actually in the PHP code we only request these:
```
use Greew\OAuth2\Client\Provider\Azure;

$params = [
    'clientId' => $clientId,
    'clientSecret' => $clientSecret,
    'redirectUri' => $redirectURL,
    'tenantId'=> $tenantId,
    'accessType' => 'offline',
];
$provider = new Azure($params);
$options = [
    'scope' => [
        'offline_access',
        'https://outlook.office.com/SMTP.Send'
    ]
];
```

Any idea what to try? For more information, we are using:
- CakePHP 2.10.24
- PHP 8.0
- PHPMailer 6.6.5

On the same server we have managed to get **SMTP sending using Google's OAuth to work**, so it is understood that the server is **correctly configured** for sending emails. 
Thank you all!

greew:
Hi @adrianalumne 

Have you managed to get a refresh_token successfully from Microsoft with before sending the mail?

adrianalumne:
Hi @greew 
Yes, we get an alphanumeric refresh_token and it starts like this:
0.AXoA4HdsoNQi_EOoT...

decomplexity:
Trivial point to check, but are you sure that this is a refresh token and not an authorization code? They both start in a similar way.

adrianalumne:
> Trivial point to check, but are you sure that this is a refresh token and not an authorization code? They both start in a similar way.

```
//Try to get an access token (using the authorization code grant)
$token = $provider->getAccessToken(
    'authorization_code',
    [
        'code' => $_GET['code']
    ]
);
//Use this to interact with an API on the users behalf
//Use this to get a new access token if the old one expires
$refreshToken = $token->getRefreshToken();
debug($refreshToken);
//0.AXoA4HdsoNQi_EOoT...
```

I guess that is correct, initially I retrieved the AccessToken and later the RefreshToken. It is right?

decomplexity:
Looks OK. Have a look at the Signin log in AAD :
select Enterprise Applications => <your app> =>Sign-in logs => User sign-ins (non-interactive) => [open an Aggregate if there is is one] =>select the latest one of two signins =>note the resource ID => [on the list top line select the "..." more => select Additional details => note the Oauth Scope Info.
Please let us know the two noted items. There might also be an error shown instead of Success

adrianalumne:
Here are the requested data

`00000002-0000-0ff1-ce00-000000000000`

`Oauth Scope Info:	["SMTP.Send"]
`

decomplexity:
Hmm.... those are both correct. 
Could I then suggest that you display and copy the access token you are actually using. Then decode with e.g. jwt.io and check that the salient fields, especially 'aud' and 'scp' are the same as the ones in AAD signin.
 
To be sure it is the one really being used, you could add the  following highlighted three lines temporarily to PHPMailer module OAuth.php:
   protected function getToken()
	{
	  **$accesstoken = $this->provider->getAccessToken(**
           $this->getGrant(),
           ['refresh_token' => $this->oauthRefreshToken]
           );
	     **var_dump($accesstoken);
	     return $accesstoken;**
        }
	
and then execute a PHPMailer send.

adrianalumne:
We have modified the file but **without success**, it seems that it does not reach that point. We have made a **var_dump of $mail** and it returns us the **correct data** and in the immediately following line is the **$mail->send()** but it seems that it is **not executed**.

Anyway, I wanted to ask you, on what occasion do you use the accessToken? we use the code that we receive through **$_GET['code']**, then with **$token = $provider->getAccessToken()** we obtain an object that, among other data, contains a refreshToken that is obtained with **$token->getRefreshToken()** and accessToken which is obtained with **$token->getToken()** but we do not use the latter and in your examples it is not mentioned either. 

Perhaps we are mixing name and variables, everything has a similar name

decomplexity:
You are correct. The refresh token is obtained in the way you suggest and plugged into PHPMailer via:
$mail->setOAuth(
    new OAuth(
        ['refreshToken' => $refreshToken,
           etc...        ]
    )
); 
From memory, theLeague's oauth2-client uses the refresh token to obtain the access token, which is then requested by PHPMailer's OAuth.php. PHPMailer.php itself initiates the authentication and sending using this access token.

adrianalumne:
Hi again,
We have been debugging the PHPMailer code to see at what point the connection fails and we have found the following.

In the src/Provider/AbstractProvider.php file, within the getParsedResponse() function we have retrieved the data from the $parsed variable (line 635) and it returns the following:
```
array(
	'error' => 'invalid_grant',
	'error_description' => 'AADSTS65001: The user or administrator has not consented to use the application with ID 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx' named 'My PHPMailer app'. Send an interactive authorization request for this user and resource.
	Trace ID: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
	Correlation ID: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
	Timestamp: 2022-10-25 14:40:52Z',
	'error_codes' => array(
		(int) 0 => (int) 65001
	),
	'timestamp' => '2022-10-25 14:40:52Z',
	'trace_id' => 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx',
	'correlation_id' => 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx',
	'suberror' => 'consent_required'
)
```

What is this error due to? Is it a DAA thing? How could we solve it?

We are still stuck and unable to move forward, thanks for your help

greew:
@decomplexity Great point!

In the walkthrough I made in the wiki article, I used the same user for both user principal and for oauth2 authentication. 

@adrianalumne Let us know, if it helps using the "Grant Admin Consent" and I will update the wiki article with the updated data :)

adrianalumne:
Hello! We have tried it, we have clicked on the button and the 'Status' column has been updated but nothing has changed, it is still showing the same error.

image

In the following path: `/vendor/league/oauth2-client/src/Provider/AbstractProvider.php (line 636)` we have added a debug of the response and the following appears:

```
object(GuzzleHttp\Psr7\Response) {
	[private] reasonPhrase => 'Bad Request'
	[private] statusCode => (int) 400
	[private] headers => array(
		'Cache-Control' => array(
			(int) 0 => 'no-store, no-cache'
		),
		'Pragma' => array(
			(int) 0 => 'no-cache'
		),
		'Content-Type' => array(
			(int) 0 => 'application/json; charset=utf-8'
		),
		'Expires' => array(
			(int) 0 => '-1'
		),
		'Strict-Transport-Security' => array(
			(int) 0 => 'max-age=31536000; includeSubDomains'
		),
		'X-Content-Type-Options' => array(
			(int) 0 => 'nosniff'
		),
		'P3P' => array(
			(int) 0 => 'CP="DSP CUR OTPi IND OTRi ONL FIN"'
		),
		'x-ms-request-id' => array(
			(int) 0 => 'XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX'
		),
		'x-ms-ests-server' => array(
			(int) 0 => '2.1.14006.8 - WEULR2 ProdSlices'
		),
		'X-XSS-Protection' => array(
			(int) 0 => '0'
		),
		'Set-Cookie' => array(
			(int) 0 => 'fpc=XXXXXXXXXXXXXXXXXXXXXX-XXXXXXXXXXXXXXXXXXXX; expires=Fri, 25-Nov-2022 07:45:50 GMT; path=/; secure; HttpOnly; SameSite=None',
			(int) 1 => 'x-ms-gateway-slice=estsfd; path=/; secure; samesite=none; httponly',
			(int) 2 => 'stsservicecookie=estsfd; path=/; secure; samesite=none; httponly'
		),
		'Date' => array(
			(int) 0 => 'Wed, 26 Oct 2022 07:45:50 GMT'
		),
		'Content-Length' => array(
			(int) 0 => '604'
		)
	)
	[private] headerNames => array(
		'cache-control' => 'Cache-Control',
		'pragma' => 'Pragma',
		'content-type' => 'Content-Type',
		'expires' => 'Expires',
		'strict-transport-security' => 'Strict-Transport-Security',
		'x-content-type-options' => 'X-Content-Type-Options',
		'p3p' => 'P3P',
		'x-ms-request-id' => 'x-ms-request-id',
		'x-ms-ests-server' => 'x-ms-ests-server',
		'x-xss-protection' => 'X-XSS-Protection',
		'set-cookie' => 'Set-Cookie',
		'date' => 'Date',
		'content-length' => 'Content-Length'
	)
	[private] protocol => '1.1'
	[private] stream => object(GuzzleHttp\Psr7\Stream) {
		[private] stream => resource
		[private] size => (int) 604
		[private] seekable => true
		[private] readable => true
		[private] writable => true
		[private] uri => 'php://temp'
		[private] customMetadata => array()
	}
}
```

The first two lines catch my attention, it seems that an error has occurred when making the request. Any ideas?

After this debug is the one that happened to you yesterday, it seems that the one from yesterday is a consequence of this.

PS: I have consulted and it seems that the account we are using is the same one that created the Azure Active Directory, so I understand that we are administrators

decomplexity:
After granting admin consent, have you regenerated the refresh token (whose settings in turn generate the access token)?
It is just a thought, because I cannot see why you should get a 'bad request' from an access token with insufficient privileges. 
One other thing I can think of worth trying is to swap the provider for TheNetworg/oauth2-azure (Jan Hajek's) provider and see if you get the same error.

JosephAlonzo:
hello guys, i try to use azure provider as well, but i get this message 

"AADSTS65001: The user or administrator has not consented to use the application with ID '8a89ecdc-67ac-4b79-a52b-7b601b1a0929' named 'demo9'. Send an interactive authorization request for this user and resource.Trace ID: 30109632-ec93-4b41-8cc9-2466b71e5500Correlation ID: 05cef712-8481-4a27-8f70-b56461df06afTimestamp: 2022-10-26 11:50:41Z"

any idea ?

Thanks 

1.- i give it permissions
image

2.- i get the token
image

3.- i get this message when i try to send an email
image

greew:
@JosephAlonzo 

Are you using the same user to both get the access token and also send mail from?

JosephAlonzo:
yes i'm using the same 

```
$mail->setOAuth(
    new OAuth(
        [
            'provider' => $provider,
            'clientId' => $clientId,
            'clientSecret' => $clientSecret,
            'refreshToken' => $refreshToken,
            'userName' => $email, <------ **I'm using the same email that i use for get token**
        ]
    )
);
```

JosephAlonzo:
image

i got this after introduce my email

JosephAlonzo:
If i change the scope like this:

image

 i get this message (when i try to send the email) instead of the error "AADSTS65001":
 
image

i don't know if it is better or not

decomplexity:
MSFT OAuth scope parameters are by default space (and not comma) separated and each parameter is not enclosed in quotes. Just enclose the entire scope parameter list in quotes, i.e. treat as one string.
Also, in the example you give above, I cannot think why you would receive TWO (failed) attempts at authentication.

greew:
I can see that when I select the API permission scopes, the `User.read` scope is already selected, but it is not requested in the example, and my setup works anyway.

I've just tried setting up a new App registration in Azure, and the first time I tried getting a refresh token, I got an error that it was a wrong client secret, but when I retried 10 seconds later, the same data was accepted. I did try only a few seconds after having created the client secret, so maybe the data hadn't propagated all the way through Azure datacenters yet.

greew:
Another question:

When you register the application, which of the account types do you choose?

image

Personally, I have only tried with the first one (single tenant) as it's the only one, I have permission to use.

greew:
I have just tried setting up an Application without the "User.read" scope, and this works as well (even though Azure warns me, that it is an essential scope and the application will fail, if I remove it).

JosephAlonzo:
Hello i just made a new application with this configuration to validate with you, but i have the same error 
"AADSTS65001: The user or administrator has not consented to use the application with ID..."

**i have this configurations in azure**

2022-10-27_09h50_01
2022-10-27_09h53_14

**i'm using these scopes (I'm using the same scopes that i have in my azure config)**

image

**but i validate my token_access and i saw the scp value not contain the offline scope, idk if that is normal**

image

Thank you very much for the help.

decomplexity:
Unless you are really using implicit grant rather than authorization_code grant, you shouldn't have the 'Access tokens used for implicit flows' ticked (and if you are using implicit grant, it is deprecated as insecure!)
And the absence of 'offline_access' in scp is expected as it is the one openid parameter that is not strictly a permission as it does not give the app itself any new access rights. But why MSFT didn't leave it scp is known only to them.

JosephAlonzo:
Hello @decomplexity thanks i made a new one using this config, and i'm getting the same

image

JosephAlonzo:
yes @greew i was also set the profile, openid, email and User.read scopes in the Azure Portal 

with this configuration i don't have the error de "AADSTS65001: The user or administrator has not consented to use the application with ID..." but it's still not working, i'm getting this output

image

decomplexity:
> I have just tried setting up an Application without the "User.read" scope, and this works as well (even though Azure warns me, that it is an essential scope and the application will fail, if I remove it).

From memory, I believe you can first authenticate such a registered app  with no permissions whatever set in AAD, provided you are logged on as user principal and  give 'user consent' in the dialogue that follows.  I guess MSFT never thought of what happens when you run a 'demon' application that uses authorization_code grant but - like e.g. PHPMailer - doesn't need end-user logon: the single refresh token is always for the user principal user so doesn't need User.Read (or Profile) - we know who the user is! 
In passing, I am still engaged in a dialogue with MSFT's Graph Development Support team in India, because I have some evidence that scopes in a scope list without a URI prefix do not always default to a URI of https://graph.microsoft.com; this leads to all sorts of side-effects, especially that the order of scopes matters...
I am waiting for MSFT to shoot me down!

adrianalumne:
Hello again, we have been reading all these messages and we have continued working to find a solution, but we are still the same, stuck in the same problem as @JosephAlonzo.

We have been trying the solutions that you proposed to him, but nothing, maybe he and we have the same problem?

I still think it may be a misconfiguration of the Azure account, some kind of subscription to send emails through SMTP or a failure to create the microsoft account. I am referring to the account itself, not the Azure portal, because it is rare to have administrator permissions missing, but that account is the administrator of the Azure portal. Can you think of any way to solve it?

decomplexity:
Worth checking the following:
AAD portal (headed <tenant name> Overview) => Properties => (at bottom of panel) Access management for Azure resources.
Your logon name should set to Yes below “can manage access to all Azure subscriptions and management groups in this tenant”

adrianalumne:
Reviewed! It is the same email, it is correct so that is not the fault.

Now that it is mentioned, do we have to create groups or roles? we have not managed absolutely none of that, apart from that, is any extra configuration needed outside of AAD?

I mean in [home page](https://portal.azure.com/#home). Following your guide doesn't mention any extra steps outside of AAD, but just in case we're missing something...

decomplexity:
There is no need to create groups or roles for authorization_code grant flows under  Delegated Permissions.

JosephAlonzo:
i try with another azure account and i don't have anymore the error of permission, 
maybe something with azure config?

joe-oems:
The 3rd link was broken on the wiki page, I went and fixed it myself thinking Github would stop me but it didn't. Maybe there should be some permissions? or is this an acceptable and trustful standard I just haven't seen before

Synchro:
The wiki has flipped in and out of being publicly writeable over the years, but for the most part they have been ok being open. The wiki is fully versioned, so changes can always be reversed. Leaving it open means that not everything is left to maintainers (i.e. me!)

dataconnectnl:
Got a question. I've got this working. However something isnt clear in my mind. If I generate a secret token it's only the token for the account which i'm logged in with to generate the token. So for info@ it's a different token then support@. I know it's the case because I already tested it and with both I can mail with their own tokens. But why not one token for all mail accounts that are granted access (owners) in de APP in Azure. Something I missed? Or is it just how it works? 

Hope that microsoft is thinking of replacing that stupid popup credentials screen for code solution...

witpok:
@adrianalumne could you please respond if you have managed to get it working? I am setting everything exactly as specified on the wiki page, yet no success with sending emails. 535 5.7.3 Authentication unsuccessful
Application created by the same user as sender, global admin. User sign-ins (not-interactive) list correctly each attempt to send email, status success, additional details: 
Root Key Type Unknown
Oauth Scope Info ["SMTP.Send"]
All permissions set as described, user.read, offline_access, SMTP.send, mail.send, even mail.send.shared, all delegated.
If anyone has working example and is able to send emails via smtp and ms accounts, I would be grateful to have a look. Thanks

decomplexity:
A scope of "SMTP.Send" will force your resource API to be Graph. And Graph does not support SMTP.Send itself.
Try a scope of:
```
"offline_access https://outlook.office.com/SMTP.Send"
```
If this sounds bizarre, it is! See the explanation in "_Microsoft OAuth2 SMTP issues_" in the PHPMailer Wiki
And don't feel tempted to add any other permissions to the Scope; it can be done, but you need to understand some undocumented MSFT rules, one of which is that the order of scope permissions matters

decomplexity:
And what operands are you using in your global PHPMailer code when instantiating OAuth in (e.g.) 
```           $this->mail->setOAuth(
               new OAuth(
                   [ ?? 
                   ]
```

witpok:
`$smtp->setOAuth(
   new OAuth(
      [
             'provider' => $provider,
             'clientId' => $clientId,
             'clientSecret' => $clientSecret,
             'refreshToken' => $refreshToken,
             'userName' =>$username
       ]
   )
);
`

decomplexity:
Try:

- specifing your app to AAD as multitenant. If you are specifying the tenant ID, double-check the tenant value 
  - sign in to AAD
  - select Applications
  - select App registrations
  - select your app
  - look in the central area near the top and copy Directory (tenant) ID  

- changing your provider from Greew to TheNetworg – this is the one we use for all our sites

- double checking that the token you are copying from the initial authorization-flow run to plug into PHPMailer is actually a refresh token and not an access token or an authorization code. Access tokens can be decoded with jwt.io or jwt.ms

witpok:
Here is my full script for email sending, removed some specific internal parts that are not needed here.
```<?php
chdir("phpmailer/vendor");
require 'autoload.php';

use PHPMailer\PHPMailer\Exception;
use PHPMailer\PHPMailer\OAuth;
use PHPMailer\PHPMailer\PHPMailer;
use PHPMailer\PHPMailer\SMTP;

use Greew\OAuth2\Client\Provider\Azure;
//use TheNetworg\OAuth2\Client\Provider\Azure;
//use Stevenmaguire\OAuth2\Client\Provider\Microsoft; <-- no matter which one used, I can connect via provider (so clientId, tenantId and secret are correct) but SMTP authentication unsuccessful error still appears

date_default_timezone_set('Etc/UTC');

$host = "smtp.office365.com";

$smtp = new PHPMailer();
$smtp->isSMTP();
$smtp->Host = $host;             
$smtp->Port = 587;  
$smtp->SMTPSecure = PHPMailer::ENCRYPTION_STARTTLS;
$smtp->SMTPAuth = true;
$smtp->AuthType = 'XOAUTH2'; 
$username = 'username@microsoft.com'; //both creator of the app, global administrator and email sender, to simplify the use case

$clientId = 'client id';
$clientSecret = 'client secret';
$tenantId = 'tenant id';

$redirectUri = 'https://mydomain/phpmailer/get_oauth_token.php';

$refreshToken = '0.ATsAJUw1LGBtTUgf2acU....'; // received under $redirectUri  with the same $clientId, $clientSecret, $tenantId  as above

$provider = new Azure(
     [
     'clientId' => $clientId ,
     'clientSecret' => $clientSecret,
     'tenantId' => $tenantId
     ]
);
$provider->scope = 'offline_access https://outlook.office.com/SMTP.Send'; // scope or scopes, no difference. In Azure.php in greew library there are defaults as below:
/* $defaultScopes = [
        'https://outlook.office.com/SMTP.Send',
        'offline_access'
    ];
*/


//$provider->urlAPI = 'https://outlook.office.com'; // tried with this one, no changes

$smtp->setOAuth(
    new OAuth(
        [
        'provider' => $provider,
        'clientId' => $clientId,
        'clientSecret' => $clientSecret,
        'refreshToken' => $refreshToken,
        'userName' =>$username
        ]
    )
);
$smtp->SMTPDebug = 3;

$smtp->addAddress('recipients email address');
$smtp->isHTML(true);
$smtp->Subject = 'test subject';
$smtp->Body = 'email body';

$smtp->setFrom($username);

$mail = $smtp->send();
echo $mail->ErrorInfo;
echo 'sending... ';
?>
```

witpok:
I have just changed from single-tenant to multi-tenant:
obraz
Tenant ID is correct from Directory (tenant) ID
obraz
Changed to `use TheNetworg\OAuth2\Client\Provider\Azure;`
Debugged refreshToken on jwt.io, can't be debugged.

I am constantly trying again, as maybe some changes need time to propagate. I will try with creating new refresh token after change tu multi-tenant.

decomplexity:
Other things to check:

1.  Whether SMTP AUTH is ‘allowed’ in Admin Center / Active Users / [select user] / Mail tab / Manage email apps

2.  Your ISP is not blocking external SMTP endpoints such as smtp.office365.com. GoDaddy and other ISPs who use GoDaddy cpanel services block such SMTP access: they insist you go via their localhost. If you are blocked in this way, it should be apparent from the PHPMailer diagnostics in $this->mail->SMTPDebug = SMTP::DEBUG_LOWLEVEL is set (I believe you have this already set).  

3.  If you haven’t looked already, it is worth accessing the AAD sign-in logs to see if there is anything of interest there. After you have signed in once, the MSFT authorization-code server should think that it has already passed you an authorization code when it calls the redirectURI (the code is suffixed to the URI).
Try:
- sign in to AAD
- select Enterprise Applications
- select your app
- select Sign-in Logs (near the bottom of the vertical list)
- select User-sign-ins (non-interactive)
- select Refresh
- select the Request ID ‘Aggregate’ item for today’s date and for Resource ‘Office 365 Exchange Online’, 'Graph' or other request you wish to check
- select the event for the time of your last failed event (or any other event in which you are interested)
This may have some useful information.

Then also:
- select the ‘…’ on the RHS of the title
- see the Oauth Scope Info field in the Additional Details tab


4. To check if the access token itself causes the problem, you could try inserting the token you got from the initial authorization flow run  directly into PHPMailer’s Oauth.php module. Since access tokens have a short life, it is best to do this fairly quickly.
Something like: 

a. Get the initial access token from 
```
$token = $provider->getAccessToken('authorization_code', [ 'code' => $_GET['code']]
```
b. plug it (temporarily!) into Oauth.php:
```
protected function getToken()
    {
        return <your initial access token>;
   }
```
It is worth noting that whereas the authentication code (in the initial authorization flow run) is allowed to contain clashing scopes, multiple resources and so on, the token endpoint imposes more stringent criteria when generating an access token.  Exactly why is known only to MSFT!
The implication is that you could have a successful initial authorization flow run that generates a refresh token that is more restrictive; but exactly how restrictive it is only becomes apparent later when it is exchanged for an access token (which Oauth2.php fails to authenticate)

witpok:
1. In https://portal.office.com/Adminportal this user has Authenticated SMTP checked (every other option as well). This user did not previously had a mailbox created, but now it has and I have successfully sent an email from the mailbox on https://outlook.office365.com/. No changes with phpmailer unfortunately, even with new refreshtoken.
2. debug was set to level 3, changed to yours and there is more information, such as Auth method requested: XOAUTH2, methods available on the server: LOGIN,XOAUTH2, then some call with CLIENT -> SERVER: AUTH XOAUTH2 [some long token here], and then SMTP INBOUND: "535 5.7.3 Authentication unsuccessful, then SERVER -> CLIENT: 535 5.7.3 Authentication unsuccessful, then SMTP ERROR: AUTH command failed: 535 5.7.3 Authentication unsuccessful, SMTP Error: Could not authenticate, Quit, Service closing transmission channel, closed. SMTP connect() failed. I don't think ISP is blocking anything, as we are still using the old way smtp to send emails from ms accounts and everything still works, even if MS was to shut it down in January.
3. Yes, I have checked Sign-in logs, everyt attempt has status Success, in Additional details `Oauth Scope Info
["Mail.Send","Mail.Send.All","Mail.Send.Shared","SMTP.Send","User.Read"]`
In authentication details and basic information everything looks fine to me
obraz
4. I will do the tests, decode the access token and comment below. Thank you for all the help, I really appreciate it.

witpok:
4. - success! after I have copied the access token received from get_oauth_token.php into phpmailer/src/Oauth.php getToken, I got Authentication successful and I have received the email on recipients inbox. Now, what do we do with this to make it work?

witpok:
I have compared two access tokens and to my surprise - with both of them authentication now passes through, emails are sent. Either the first correct token pasted into OAuth.php made the path open for following tokens or some previous changes made in the administration panels finally propagated. I'm baffled, can't explain it. The only differences in both tokens are times (iat, nbf, exp), and fields aio and uti. Scope is the same, everything else is the same. Aud is "https://outlook.office.com", scope is Mail.Send Mail.Send.All Mail.Send.Shared SMTP.Send User.Read.
I will repeat tests tomorrow, when tokens expire to be sure if it is really working. 
If I were now send emails as other accounts, do I need to add my main account in https://admin.exchange.microsoft.com ->Manage Mailboxes->mailbox to be sent from->Delegation->Send as for each mailbox or set it up somewhere else? Right now I have 5.2.252 SendAsDenied if I try to change setFrom address, and Authentication unsuccessful if I try to change $username.

decomplexity:
Since you have an ‘aud’ of https://outlook.office.com/ in both tokens, that is half the battle of getting SMTP authentication to work! As I mentioned earlier in this thread, your client scope normally only needs to be:
```
"offline_access https://outlook.office.com/SMTP.Send"
```
Re the setup for allowing other users (within the tenant presumably) to ‘send as’ the base mailbox, two points: 
- the base mailbox becomes a shared mailbox, and shared mailboxes run disabled from login, which might not be what you want if the base mailbox is an admin one! 
- do the other users want to send as themselves (‘Send On Behalf Of’ – which is set using a cmdlet) or simply use the shared mailbox as a common address to send from (Send As) ?

Without experimenting, I really can’t say whether adding delegates to a shared mailbox in Send As mode and then hoping to allow each to use PHPMailer to send via SMTP will work even in principle. Logically it shouldn’t. 
But anyway, there should be no salient difference to the email recipient in sending via SMTP from ANY user (whether in the tenant or not – as in a website Contact form for example) in the usual way PHPMailer is used as a central daemon application, as the one central shared email address is used by this and by Send As. To distinguish each different sender you should be able to use different email name fields as in **\<John Doe\>** Mydept@Deer.com, where there can be umpteen different John Does but only one Mydept@Deer.com

witpok:
After leaving it for the night, this morning "Send as" delegation started to work without a touch, so information for everyone - be patient, changes on MS side can take much more time than described in the panel (5 minutes). Thank you for all the help, I hope this will be of use for everyone else fighting with this.

marcusBamboo:
Hello,
I'm currently trying to get oauth2 for office365 to work with PHP Mailer and have simmilar problems as stated in this thread.

I'm using the 'TheNetworg' Provider to get a RefreshToken and am trying to use it to send mails, too. But I keep running into the "SMTP ERROR: AUTH command failed: 535 5.7.3 Authentication unsuccessful" problem.

I tried putting the accessToken into the OAuth class as described above, but it didn't change anything. On the initial permission grant request, I noticed that it didn't ask me to grand Mail perssions. So I looked into the Sign-In logs and saw that the scopes arent the ones I specified in the provider. Any Idea why this would happen?

image

At some point I added these permissions to the AAD App, but that didn't change anything:
image

Provider Factory Function:
```
	public function getProvider($mandant, $mailProfile = null) {
		
		if (empty($mailProfile)) {
			/** @var MailProfile $mailProfile */
			$mailProfile = $this->em->getRepository("BSLibEntityBundle:StammDat")->getMailProfile($mandant);
		}
		
		if(isset($_SERVER['HTTPS'])){
                        $protocol = ($_SERVER['HTTPS'] && $_SERVER['HTTPS'] != "off") ? "https" : "http";
		} else{
			$protocol = 'http';
		}
		$baseUrl = $protocol . "://" . $_SERVER['HTTP_HOST'];
	
		$provider = new Provider\Azure([
		     'clientId'           => $mailProfile->getUser(), // Azure App (user) ID
		     'clientSecret'    => $mailProfile->getPassword(), // Azure App Secret
		     'tenantId' 	    => $mailProfile->getTenantId(), // Azure App Directory (tenant) ID
		     'redirectUri'      => $baseUrl . $this->router->generate("Mail_OAuth_Get_Token", ['mandant' => $mandant]),
		     'scopes'            => ['offline_access', 'https://outlook.office.com/SMTP.Send'],
		     'defaultEndPointVersion' => '2.0',
	        ]);
		
		// Set to use v2 API, skip the line or set the value to Azure::ENDPOINT_VERSION_1_0 if willing to use v1 API
		$provider->defaultEndPointVersion = Provider\Azure::ENDPOINT_VERSION_2_0;
		
		//$provider->scope = "offline_access https://outlook.office.com/SMTP.Send";
		//$provider->urlAPI = "https://outlook.office.com";
		
		return $provider;
	}
```

Mailer:
```
$mail = new PHPMailer(true);
		
//Server settings
$mail->SMTPDebug = SMTP::DEBUG_LOWLEVEL ; 
$mail->isSMTP();
$mail->Host             = $mailProfile->getServer();              // smtp.office365.com
$mail->SMTPAuth    = true;
$mail->Username     = $mailProfile->getUser();                // AAD App (client) id
$mail->Password      = $mailProfile->getPassword();        // AAD App secret
$mail->SMTPSecure = $mailProfile->getEncryption();      // tls
$mail->Port               = $mailProfile->getPort();                // 587
$mail->Timeout        = $mailProfile->getTimeout();          // 60

if ($mailProfile->getAuthMethod() == 'oauth') {
	$mail->AuthType = 'XOAUTH2';
	
	$mail->setOAuth(new OAuth(
		[
			'provider' => $this->oAuthService->getProvider($item->getMandant(), $mailProfile),
			'clientId' => $mailProfile->getUser(),
			'clientSecret' => $mailProfile->getPassword(),
			//'accessToken' => $mailProfile->getOAuthAccessToken(),
			'refreshToken' => $mailProfile->getOAuthRefreshToken(),
			'userName' => $senderMail,
		]
	));
}
```

If anyone has some insight, it would be much appreciated.

decomplexity:
Yes - you are right  to be suspicious of the scopes shown in the Signin log!
The scopes presented to the authorization server do not include SMTP.Send, and subsequent authentication will fail.

From a very brief look at your settings, try:

- in new Provider\Azure([
remove the 'tenantId' and 'scopes' operands and their settings

in the subsequent few lines:
- uncomment the //$provider->scope ...    line

- replace the //$provider->urlAPI = https://outlook.office.com;
by
$provider->urlAPI = https://graph.microsoft.com/;

- add:
$provider->API_VERSION = '1.0';

marcusBamboo:
Hey, thanks. I got it to work, I found out that I had an old script loaded where I was setting the scope to the old params.

matt-t-tinkers:
Is there a route to achieving something similar but without the deprecated SMTP Auth?

br4nnigan:
@greew 

can you help me out maybe? I did everything as described in the tutorial, except that I don't know how to get "the user that gave consent" userName parameter (`$provider->getResourceOwner($token)` yields a "Access token is empty" error). But even if I hard code that to the e-mail I gave consent with, I still get the same error:

	AADSTS9002313: Invalid request. Request is malformed or invalid. Trace ID: 3f2d3ced-7966-4d9d-b33a-28196f6b3302 Correlation ID: 9c436b3c-a277-4197-96c9-9aced7d81d88 Timestamp: 2024-04-08 16:10:13Z2024-04-08 16:10:13 SERVER -> CLIENT: 220 FR5P281CA0031.outlook.office365.com Microsoft ESMTP MAIL Service ready at Mon, 8 Apr 2024 16:10:12 +0000
	\n2024-04-08 16:10:13 CLIENT -> SERVER: EHLO mydomain.org
	\n2024-04-08 16:10:13 SERVER -> CLIENT: 250-FR5P281CA0031.outlook.office365.com Hello [85.13.166.xxx]250-SIZE 157286400250-PIPELINING250-DSN250-ENHANCEDSTATUSCODES250-STARTTLS250-8BITMIME250-BINARYMIME250-CHUNKING250 SMTPUTF8
	\n2024-04-08 16:10:13 CLIENT -> SERVER: STARTTLS
	\n2024-04-08 16:10:13 SERVER -> CLIENT: 220 2.0.0 SMTP server ready
	\n2024-04-08 16:10:13 CLIENT -> SERVER: EHLO mydomain.org
	\n2024-04-08 16:10:13 SERVER -> CLIENT: 250-FR5P281CA0031.outlook.office365.com Hello [85.13.166.xxx]250-SIZE 157286400250-PIPELINING250-DSN250-ENHANCEDSTATUSCODES250-AUTH LOGIN XOAUTH2250-8BITMIME250-BINARYMIME250-CHUNKING250 SMTPUTF8

If I use SSL instead of TLS the smtp connect times out.

decomplexity:
Probably something wrong with your refresh token. 
Have a look at  the AAD sign-in logs. 

- sign in to AAD
- select Enterprise Applications
- select your app
- select Sign-in Logs (near the bottom of the vertical list)
- select User-sign-ins (non-interactive)
- select Refresh
- select the Request ID ‘Aggregate’ item for today’s date and for Resource ‘Office 365 Exchange Online’, or 'Graph' 
- select the event for the time of your last failed event 
This may have some useful information.

Then also:

- select the ‘…’ on the RHS of the title
- see the OAuth Scope Info field in the Additional Details tab.
- 

Let us have the _Resource_ shown in the basic tab and the _Scope_ shown in the Additional Details tab

br4nnigan:
Thank you, pointed me in the right direction! the error was on my side, the token was truncated in the DB because of the size of the field...

next I only had to enable SMTP auth as described [here](https://learn.microsoft.com/en-us/exchange/clients-and-mobile-in-exchange-online/authenticated-client-smtp-submission), which might be useful

Shanksum:
Hopping on here, because after my first implementation was successful, the Azure refresh token has now expired after 90 days. Is it possible to configure the code, such that a new refresh token is pulled automatically without admin input @greew? 

If I manually request a refresh token from `get_oauth_token.php`, I am redirected to my Office365 login where I have to accept the new token generation/access. I fear that therefore I cannot automatically generate a new token after 90 days... Would be a pain if I had to come back every three months to keep my mail system running. Appreciate any ideas or links to further documentation, since your example file has a hardcoded refresh token!

decomplexity:
To avoid having to login, you need to capture a new refresh token when you obtain an access token. 
You can either do this every time, or keep a record of when a refresh token is coming up to its lifetime and then capture a new refresh token. It is easiest - but not perhaps quite as efficient - to do this each time.

Firstly, make sure that your 'scope' claim has requested 'offline_access'

You then need to make a one-line addition to TheLeague oauth2-client/src/Token/AccessToken.php:
The existing code from around line 107 (exact line is version dependent) reads:
```
        if (!empty($options['refresh_token'])) {
            $this->refreshToken = $options['refresh_token'];
```
Then before the }, add the line:
```
  	        $_SESSION[__NAMESPACE__ . "\\updatedRefreshToken"] =  $this->refreshToken; 
```
This saves the refresh token as a SESSION variable (you can use any name: the use of NAMESPACE in the key is simply to try to avoid any clash)

Later, you then just need to read this SESSION variable, save the token in single record file or database, and read it next time around.

Shanksum:
Ah thanks for the quick reply! I had problems with the namespace SESSION variable approach and ended up using a fixed name instead. I also had to add a `session_start()` because I wasn't working with sessions before.

For now, it seems to be working and I successfully generated a new refresh token that way. I went with saving the creation date and triggering a refresh every 60 days, so fingers crossed that in 60 or 90 days the system will continue to work :)

timramakers:
> Since you have an ‘aud’ of https://outlook.office.com/ in both tokens, that is half the battle of getting SMTP authentication to work! As I mentioned earlier in this thread, your client scope normally only needs to be:
> 
> ```
> "offline_access https://outlook.office.com/SMTP.Send"
> ```
> 
> Re the setup for allowing other users (within the tenant presumably) to ‘send as’ the base mailbox, two points:
> 
> * the base mailbox becomes a shared mailbox, and shared mailboxes run disabled from login, which might not be what you want if the base mailbox is an admin one!
> * do the other users want to send as themselves (‘Send On Behalf Of’ – which is set using a cmdlet) or simply use the shared mailbox as a common address to send from (Send As) ?
> 
> Without experimenting, I really can’t say whether adding delegates to a shared mailbox in Send As mode and then hoping to allow each to use PHPMailer to send via SMTP will work even in principle. Logically it shouldn’t. But anyway, there should be no salient difference to the email recipient in sending via SMTP from ANY user (whether in the tenant or not – as in a website Contact form for example) in the usual way PHPMailer is used as a central daemon application, as the one central shared email address is used by this and by Send As. To distinguish each different sender you should be able to use different email name fields as in **<John Doe>** [Mydept@Deer.com](mailto:Mydept@Deer.com), where there can be umpteen different John Does but only one [Mydept@Deer.com](mailto:Mydept@Deer.com)

Hopping onto this; my access token comes back perfectly but the `aud` is consistently `https://graph.microsoft.com` instead of `https://outlook.office.com/` when I decode it on jwt.ms. I can't seem to figure out what causes this, I've followed all the guides as well as this whole thread to a T but to no avail. Any help would be greatly appreciated!

PR #2796, opened 2022-10-10
GH Actions: use semver branch/tag references instead of commit references

jrfnl:
Most predefined action runners offer a long-running branch or a tag which gets moved every release to allow staying on the "latest" version of a certain major of their action, without having to update the workflow scripts on every release of the action runner.

This works well for action runners which follow semver.

I've reviewed the existing workflows and tweaked the versions used whenever possible to make optimal use of this.

* For the `ossf/scorecard-action` action runner, I have not been able to find a workable tag/branch to fix this on.
    Note: I have remove the "# v1.1.1" comment though as it doesn't get updated by Dependabot and was sorely out of date (you are currently on version `2.0.4`).

With the changes in this PR, Dependabot should become less noisy and will only send in PRs to update the action runner versions when a new major release has been tagged. (save for the one exceptions mentioned above)

I would recommend watching the following repos for new releases:
* https://github.com/actions/checkout
* https://github.com/actions/upload-artifact
* https://github.com/shivammathur/setup-php
* https://github.com/ramsey/composer-install
* https://github.com/nick-fields/retry
* https://github.com/codecov/codecov-action
* https://github.com/JamesIves/github-pages-deploy-action
* https://github.com/ossf/scorecard-action
* https://github.com/github/codeql-action

This will ensure you will get an email with the changelogs for those action runners on all releases, so you can still monitor for changes in the action runners you need to be aware of.

image

Closes #2794
Closes #2795

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -20,7 +20,7 @@ jobs:
       - name: Build Docs
         uses: ./.github/actions/build-docs
       - name: Publish Docs to gh-pages
-        uses: JamesIves/github-pages-deploy-action@v4.4.0
+        uses: JamesIves/github-pages-deploy-action@v4
         with:
           branch: gh-pages
           folder: docs
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -25,12 +25,12 @@ jobs:
     
     steps:
       - name: "Checkout code"
-        uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846 # v3.0.0
+        uses: actions/checkout@v3
         with:
           persist-credentials: false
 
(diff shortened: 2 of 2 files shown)


Synchro:
Thank you!

Synchro:
See you in 2 weeks, and don't forget my elephpant!

jrfnl:
> See you in 2 weeks, and don't forget my elephpant!

I have a feeling I'm missing something... ?

Synchro:
Demian said he gave you a Red Laravel elephpant to give to me in Munich at IPC?

jrfnl:
There must be some confusion. He didn't give me an elephant to pass on and I won't be at IPC.
Would be lovely to see you again though. Let's see if we can make it happen in another way.

PR #2798, opened 2022-10-12
Add line ending \n after POP3 quit command

czirkoszoltan:
This is required by some POP3 servers which will wait until a newline char which is delimiting all commands. Without it, the login works, but it waits for the full timeout.


Synchro:
Thanks

PR #2800, opened 2022-10-14
GH Actions: fix use of deprecated `set-output`

jrfnl:
GitHub has deprecated the use of `set-output` (and `set-state`) in favour of new environment files.

This commit updates the test workflows to avoid `set-output`.

Refs:
* https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
* https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -105,20 +105,11 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v3
 
-      - name: Set coverage variable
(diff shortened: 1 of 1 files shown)


Synchro:
What would we do without you?

jrfnl:
> What would we do without you?

I don't know ? Fix it yourself once things actually break ?

PR #2804, opened 2022-10-23
GH Actions: harden the workflow against PHPCS ruleset errors

jrfnl:
If there is a ruleset error, the `cs2pr` action doesn't receive an `xml` report and exits with a `0` error code, even though the PHPCS run failed (though not on CS errors, but on a ruleset error).

This changes the GH Actions workflow to allow for that situation and still fail the build in that case.


Synchro:
Thanks Juliette

PR #2814, opened 2022-10-26
Avoid POP3 client error message when TCP connection failed

czirkoszoltan:
The disconnect() method throws a TypeError when the TCP connection cannot be created. Error and trace:

fgets(): Argument #1 ($stream) must be of type resource, bool given

phpmailer/phpmailer/src/POP3.php(372): fgets()
phpmailer/phpmailer/src/POP3.php(345): PHPMailer\PHPMailer\POP3->getResponse() phpmailer/phpmailer/src/POP3.php(230): PHPMailer\PHPMailer\POP3->disconnect() PHPMailer\PHPMailer\POP3->authorise()

Reproduce with:

include __DIR__ . "/POP3.php";
\PHPMailer\PHPMailer\POP3::popBeforeSmtp('doesnotexist', 110);

### DIFF
--- a/src/POP3.php
+++ b/src/POP3.php
@@ -337,6 +337,11 @@ class POP3
      */
     public function disconnect()
     {
+        // If could not connect at all, no need to disconnect
+        if ($this->pop_conn === false) {
+            return;
+        }
+
         $this->sendString('QUIT' . static::LE);
 
         // RFC 1939 shows POP3 server sending a +OK response to the QUIT command.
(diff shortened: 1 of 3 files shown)


czirkoszoltan:
Hi @Synchro!

This one fixes another bug, when the TCP connection was unsuccessful, POP3::disconnect threw a TypeError.

The test `codecov/patch` cannot pass currently, because the `POP3` class is tested in `PopBeforeSmtpTest`, which is in the group named `pop3`, but that is excluded in phpunit.xml.dist line 23. I did not want to change that, probably there is a reason why it is set like that, and not overridden elsewhere.

In the meantime, the second commit fixes the test itself, because it had a bug. Now one can run it manually with `phpunit --group pop3`.

Synchro:
Thank you. I sincerely hope that nobody really needs the POP-before-SMTP auth mechanism any more. I last used it in about 1991, but I'm sure there must be some unfortunate people stuck with it!

czirkoszoltan:
Unfortunately it's still used...

PR #2830, opened 2022-11-22
Don't let RSET errors overwrite the original error

dotdash:
On keep-alive connections, we issue an RSET command in case of errors. But that RSET command may fail, which causes the error information from the original error to be replaced by the error information from the RSET error, which isn't helpful in diagnosing the actual cause. To avoid this problem, we need to extract the error information first, and only issue the RSET command afterwards.


Synchro:
Good catch, thanks

PR #2835, opened 2022-11-28
Fixed SMTP

lwcorp:


Synchro:
If you're using `require_once` in a modern PHP app, you're probably doing something much bigger wrong.

lwcorp:
I was just trying to help the project, sorry if I offended you.
Is there a source for this statement?
In any, I re-submitted this time keeping `require` intact.

Synchro:
Not offended, just pointing out a problematic approach that I wouldn't want to encourage in beginners in particular. No particular reference, but it's implicit in the way that it works. If you're using `require_once`, it means you don't have a single, consistent method for loading classes, leading you to risk situations where you might accidentally run this code more than once. If you use a proper loading mechanism, such as composer provides, you never need to do that. That particularly applies to structured situations where you're loading package classes, but for simple stand-alone scripts (like the readme example), there's no point in the added overhead of `require_once`, because the code in the script is only ever going to be run once during any single request, so it will never be called a second time. So `require_once` is a workaround and a symptom of poor design elsewhere in your code, and should not be encouraged.

Regarding the addition of `use PHPMailer\PHPMailer\SMTP;`, it looks like you didn't read the comment that you edited! Just because you load a class (directly or indirectly) doesn't mean you need a `use` statement too; that *only* applies if you are referencing that class *explicitly* by name *in that script*. In this case, the SMTP class needs to be *loaded* in order to send via SMTP, however, you never need to actually refer to it, hence there is no need for a `use` statement. The one way that it is used in some PHPMailer examples is when it's used to refer to SMTP class constants, such as `SMTP::DEBUG_SERVER`. It's not needed otherwise.

Overall, this looks more like an attempt to get a hacktoberfest-style credit than an actual improvement. Sorry to rain on your parade!

lwcorp:
I assume you don't mean to, but please understand you sound very aggressive (e.g. the last line), you and I are just volunteers, all is done in good faith, there's no reason to keep "attacking" (again, I assume you don't mean to) anything or anyone. The README didn't elaborate like your good explanation here about  `use` only needed for SMTP class constants. I'd edit it to include this short explanation, but it doesn't seem welcome.

Synchro:
Sorry if it came across that way – I get quite a lot of PRs that are essentially just people changing random things in order to say they had a PR accepted (especially in a relatively popular package like this one)!  There are so many ways that packages can be used, it's really not sustainable to explain every aspect of using PHP in every one of them – the PHP docs and many tutorials and articles do that. So while I rejected your PR, I wanted you to understand the reasons why I did so, and hopefully you now have a better understanding of how autoloading and namespacing works.

PR #2839, opened 2022-12-02
changing strpos to function_exists, because of the performance

jokerrs:


jrfnl (commented):
Sorry, but this code is like this for a reason.

Only a check with `strpos()` will give consistent results cross-version, which makes every other way of checking this invalid.

This can be confirmed by running the following script on multiple PHP versions:

```php
echo 'function_exists: ', var_export(function_exists('set_time_limit'), true), PHP_EOL;
echo 'is_callable: ', var_export(is_callable('set_time_limit'), true), PHP_EOL;
echo 'strpos/str_contains: ', var_export(strpos( ini_get( 'disable_functions' ), 'set_time_limit' ), true), PHP_EOL;
```
Do make sure to set `disable_functions = 'set_time_limit'` in the php.ini file prior to running the script as it can not be changed at run-time (which is also the reason why I'm not posting an 3v4l link).

Synchro:
I doubt there is much to be concerned about with respect to performance here, but the reason this somewhat obtuse approach is used is that it's quite common for shared hosting providers to disable the `function_exists` function (don't ask me why!), so it's a workaround. It's similar to the reason I check for the existence of constants to see whether `mbstring` is available, because `extension_loaded` often has the same problem.

jokerrs:
Thank you both for the response!

@jrfnl Thank you for the explanation of that approach! now when I think about that, it makes sense!

@Synchro I really didn't know that is a case, with disabling ```function_exists```. Thank you for your insight!

PR #2849, opened 2022-12-20
Fixes the, admittedly rare, case when the display name is '0'.

campbell-m:


jrfnl (commented):
Could you please add a test to safeguard the fix ?

campbell-m:
I think that's beyond my capabilities, at least not without a substantial investment of time!

campbell-m:
I've now added a test.  Hope I've done it correctly.

test/PHPMailer/AddrFormatTest.php
```diff
@@ -0,0 +1,70 @@
...
+ * @author    Marcus Bointon <phpmailer@synchromedia.co.uk>
+ * @author    Andy Prevost
+ * @copyright 2012 - 2020 Marcus Bointon
+ * @copyright 2004 - 2009 Andy Prevost
+ * @license   http://www.gnu.org/copyleft/lesser.html GNU Lesser General Public License
+ */
+
+namespace PHPMailer\Test\PHPMailer;
+
+use PHPMailer\Test\PreSendTestCase;
```
jrfnl, line 16:
```suggestion
use PHPMailer\Test\TestCase;
```

test/PHPMailer/AddrFormatTest.php
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Test address formatting.
+ *
+ * @covers \PHPMailer\PHPMailer\PHPMailer::addrFormat
+ */
+final class AddrFormatTest extends PreSendTestCase
```
jrfnl, line 23:
```suggestion
final class AddrFormatTest extends TestCase
```

jrfnl (commented):
Thank you for adding those tests @campbell-m ! Much appreciated.

Looks good to me. I've verified that the particular test related to your fix fails without the fix and passes with it. ✅ 

I've just left one small nitpick as an inline comment, other than that, all good. I can think of some more test cases to add, but that's outside the scope of this PR and can be done in a future iteration on the tests.

test/PHPMailer/AddrFormatTest.php
campbell-m: I must admit I'm unfamiliar with the whole process!  I started off by using `TestCase`, but found that I don't have access to an instance of PHPMailer in `TestCase` (`addrFormat()` is not a static method).  Is it acceptable to create a new instance of PHPMailer in the test?

campbell-m: Sorry - I'm talking rubbish.  It's OK now.

campbell-m: I've now changed the code, as you'll see.  I didn't commit your suggestion because I also had to change the `use` statement.

jrfnl: > I must admit I'm unfamiliar with the whole process!

No worries, you're doing great!

> I started off by using `TestCase`, but found that I don't have access to an instance of PHPMailer in `TestCase` (`addrFormat()` is not a static method). Is it acceptable to create a new instance of PHPMailer in the test?

Could it be that the `TestCase` you started off with was the PHPUnit `TestCase`, not the PHPMailer one ?

The PHPMailer base `TestCase` contains the `$Mail` property which you are using in the test, so switching to the PHPMailer base `TestCase` should work absolutely fine and you don't need to create a new instance of PHPMailer in the test as the base `TestCase` already takes care of that in the `setUp()` method.

_(well, the method is really called `set_up()` as the test suite uses the [PHPUnit Polyfills](https://github.com/Yoast/PHPUnit-Polyfills) to make the test suite cross-version compatible with all PHPUnit versions needed to run the tests against all supported PHP versions, but that's not really important)._

jrfnl: Only just now seeing your follow-up comments (weird, they didn't show when I was typing my message). Turns out you'd figured out already. 👍🏻

Synchro:
@campbell-m, thank you for spotting the issue and persevering with the tests, @jrfnl thanks for shepherding!

PR #2852, opened 2023-01-03
XCLIENT implementation

mta59066:
XCLIENT allows a mail server to receive certain parameters (such as ip address, sasl login, etc.) from a trusted source. https://www.postfix.org/XCLIENT_README.html

This pull request adds XCLIENT capability to PHPMailer. This way PHPMailer can connect to an SMTP server and if trusted can pass these values to the server, for example an already authenticated user will be able to send mail without the need to pass the password but will still be treated as smtp authenticated by the smtp server.

Sample code:
```php
$mail = new PHPMailer();
$mail->isSMTP();
$mail->SMTPAuth = false;
$mail->SMTPXClient = [
    'LOGIN' => 'yourname@example.com',
    'ADDR' => '10.10.10.10',
];
```

I would appreciate any comments.


Synchro:
This is interesting, but I'm not too keen on adding more public properties; Better to have them protected and provide get/set methods so that the format can be enforced. Could you also add tests to cover the functionality please?

mta59066:
Yes @Synchro that makes sense, I was trying to follow the current way of doing things, but I'll add those in. 

It would be great to setup the actions to lunch postfix and actually do the integration tests, it is pretty alien to me but I'll try to figure that out.

mta59066:
@Synchro I changed it to be in a protected variable and added a set method, also added unit tests.

mta59066:
Sample code now becomes
```php
$mail = new PHPMailer();
$mail->isSMTP();
$mail->SMTPAuth = false;
$mail->setSMTPXclientAttribute('LOGIN', 'yourname@example.com');
$mail->setSMTPXclientAttribute('ADDR', '10.10.10.10');
```

Synchro:
That's much better, thanks. Could you add a `getSMTPXclientAttributes` which simply returns the array?
Since the validation happens in two places, it's currently repeating the list of permitted values, so I suggest moving that into a constant in the SMTP class, and reference it from the PHPMailer class.

I don't think we need a whole separate example to demonstrate this – adding a few lines to the SMTP example would be enough, since it's an SMTP feature.

mta59066:
@Synchro I made the suggested changes. I added the valid attributes as a constant in SMTP class but php 5.5 does not allow arrays in constants, any suggestions on this? Would not want to use a public variable instead. I assume PHPMailer continues to be available for php 5.5?

Synchro:
Hm. That's a good point; we can't break BC for this. Annoying! I guess it will have to be a `public static` property – though I don't remember if that's permitted either!

jrfnl:
> Hm. That's a good point; we can't break BC for this. Annoying! I guess it will have to be a `public static` property – though I don't remember if that's permitted in 5.5 either!

@Synchro Properties have been able to take arrays since a long time and visibility + `static` is fine for properties (since 5.0), so using a property will work for PHP 5.5.

I'm just wondering whether it should be a `public` property or if a more restrictive visibility would be more appropriate ? Making something less visible is a BC-break, making something more visible is not, so unless there is good reason not to, I'd suggest starting with `private static`. It can always be changed to `protected` or `public` later if needs be.

Synchro:
The reason I suggested a public constant is because we wanted it to be accessible from outside the SMTP class, so that the list of valid properties could be defined in one place and used from two.

mta59066:
@Synchro I updated it with a `public static` variable instead of a `const`

mta59066:
@Synchro were you able to check the latest changes? I would like to use this in a project and would rather use the mainline instead of a fork. Let me know if there is anything else I can add to it?

Synchro:
Thanks - sorry this took so long!

PR #2863, opened 2023-01-19
added SMTP Transaction ID regex for ZoneMTA mailer

detain:
Simple change, just adding the regex so it can match the transaction id.


Synchro:
Thanks

PR #2864, opened 2023-01-19
Update for my previous ZoneMTA transaction id regex fixing the regex …

detain:
…as the id does not have parenthesis surrounding it.

sorry about that.


Synchro:
I was wondering about tests for this, however, it's a non-critical function and the place it's used isn't readily accessible from test contexts, making it a bit hard to test. A better approach would have been to pass the response from `getLastReply()` to a method that sees if it matches anything, as then it could have been more easily tested with fixed strings without needing an SMTP conversation, but hindsight and all... It could still be done by breaking it out into a protected method and tweaking visibility in the test, but that's a much bigger PR than this.

jrfnl:
@Synchro Sounds like a nice follow-up PR ;-) (no rush 😝)

PR #2869, opened 2023-02-04
Translating some hindi language keys.

jayantisuthar:
most key's can be verified easily be google translator still some words are not correctly placed by google translator so added fix for that as well


.gitignore
```diff
@@ -5,6 +5,8 @@ test/testbootstrap.php
 build/
 vendor/
 *.pem
```
Synchro, line 7: Please remove this change. The place for these IDE-specific ignores is in your global `.gitignore`, not in every package, and it has nothing to do with the subject of your PR.

jayantisuthar:
also updated some fixes i found

Synchro:
Thanks. Hindi looks very pretty, but I don't understand a word of it, so I'm glad you do!

PR #2877, opened 2023-03-01
Add support to send mails using AWS SES

christiandavilakoobin:


christiandavilakoobin:
Hi @jrfnl. I'm not sure how to test this without an AWS account and without AWS SDK libraries installed. Also, I don't know how to warn anyone who wants to use this feature that they must have AWS SDK v3 installed.

christiandavilakoobin:
Hi @Synchro. Using getSentMIMEMessage() was my first aproach, but it requieres some extra adjustments to use PHPMailer with this, and this way the "bcc" headers are ignored because "preSend" doesn't include them on MIME. Finally I had to made those changes to be more friendly to people who are already using PHPMailer. Now you're only to mark as "$mail->useSes()" and it works same way than all others.

Synchro:
BCC addresses should never be in headers? Even then, you could easily insert an extra header via your own wrapper after fetching the built message. I still think there is too much going on in here for a single proprietary option and it would be better implemented in a subclass, where you would not even have to call `useSes`.

PR #2879, opened 2023-03-06
Composer: update some CS related dependencies

jrfnl:
The Composer PHPCS plugin has released its 1.0.0 version. :tada:

As minors below a 1.0 are considered as majors by Composer, this needs an update to the `composer.json` file to get access to the new version. And as PHPCS has also released a new version, may as well update that as well.

Refs:
* https://github.com/PHPCSStandards/composer-installer/releases/tag/v1.0.0
* https://github.com/squizlabs/PHP_CodeSniffer/releases/tag/3.7.2

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,13 +37,13 @@
         "ext-hash": "*"
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


Synchro:
Thanks again!

PR #2886, opened 2023-03-27
Prevent PHP8+ Error 8192 - strlen(): Passing null to parameter #1 ($s…

xaviertnc:
…tring) of type string is deprecated


jrfnl (commented):
IMO this fix is incorrect.

The `$basedir` parameter expects a `string`. The _calling function_ passing a non-string value for `$basedir` is _doing it wrong_ and should be fixed.

This method should not be changed.

PR #2904, opened 2023-05-08
PHPMailer.php Supplemental Chinese Simplified translation。

yaoyingkeji:
Supplemental Chinese Simplified translation。
补充简体中文翻译

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -2271,6 +2271,7 @@ class PHPMailer
             'rs' => 'sr',
             'tg' => 'tl',
             'am' => 'hy',
+            'zh-CN' => 'zh_cn',
         ];
 
         if (array_key_exists($langcode, $renamed_langcodes)) {


Synchro:
I'm not clear why you added this mapping. This translation file already has always had the correct name, so it doesn't need a mapping from an old name.

yaoyingkeji:
> 
Because many programs are confusing and the names are not uniform, resulting in the invalidity of "zh-CN"; As a result, the localization program fails.
If not merged, it can be used as a reference for the revision here.
```
因很多程序混乱，名称不统一，导致  “zh-CN”失效；因此本地化失效。
如果不合并，可以作为记录此处修订参考。
```

Synchro:
OK, but that's not what that array is for; It's only there to provide backwards compatibility with language files that were renamed in the past, so that the renaming would not break existing code. It's not to make it easier to match random additions.

PR #2918, opened 2023-06-20
remove double LANG entries

Colinax:
these PHPMAILER_LANG entries already exist in the respective file


Synchro:
Thanks, good spot

PR #2929, opened 2023-07-12
Readme: Update `PHPMailerLangTest` references to `TranslationCompletenessTest`

Ayesh:
In #2377, the `PHPMailerLangTest` test was renamed to `TranslationCompletenessTest` along some optimizations. However, the link in `README.md` file still links to the old file name.

This updates the `README.md` file to use the new file name and URL.

PR #2930, opened 2023-07-12
Add Sinhalese translations

Ayesh:
Adds [Sinhalese](https://en.wikipedia.org/wiki/Sinhala_language) (ISO 639-1 `SI`) translations.


Synchro:
Thanks, and thanks for #2929 too!

PR #2947, opened 2023-08-23
fix(SMTP): set SMTP Timelimit same as Timeout

andrebolina:
In case of having problems in the SMTP server after the connection, the Timeout set wasn't being applied. This PR set the Timelimit value with the same value as Timeout to fix it.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -2114,6 +2114,7 @@ class PHPMailer
         }
 
         $this->smtp->setTimeout($this->Timeout);
+        $this->smtp->setTimelimit($this->Timeout);
         $this->smtp->setDebugLevel($this->SMTPDebug);
         $this->smtp->setDebugOutput($this->Debugoutput);
         $this->smtp->setVerp($this->do_verp);
--- a/src/SMTP.php
+++ b/src/SMTP.php
@@ -1400,6 +1400,26 @@ class SMTP
     }
 
     /**
+     * Set SMTP timelimit.
+     *
+     * @param int $timelimit The timelimit duration in seconds
+     */
+    public function setTimelimit($timelimit = 0)
+    {
+        $this->Timelimit = $timelimit;
+    }
+
+    /**
+     * Get SMTP timelimit.
+     *
+     * @return int
+     */
+    public function getTimelimit()
+    {
+        return $this->Timelimit;
+    }
+
+    /**
      * Reports an error number and string.
      *
      * @param int    $errno   The error number returned by PHP
--- a/src/SMTP.php
+++ b/src/SMTP.php
@@ -1384,7 +1384,7 @@ class SMTP
      *
      * @param int $timeout The timeout duration in seconds
      */
-    public function setTimeout($timeout = 0)
+    public function setTimeout($timeout)
     {
         $this->Timeout = $timeout;
     }
@@ -1404,7 +1404,7 @@ class SMTP
      *
      * @param int $timelimit The timelimit duration in seconds
      */
-    public function setTimelimit($timelimit = 0)
+    public function setTimelimit($timelimit)
     {
         $this->Timelimit = $timelimit;
     }


Synchro:
I'm not clear what problem this solves – the timeouts for initial connections vs individual commands are intentionally independent, as the RFC says. You can already get at both properties as they are both public. Can you provide more background information?

Also your timelimit parameter should not default to 0 – that will cause almost everything to fail. Use the RFC value as the default value:

    setTimelimit($timelimit = 300)

andrebolina:
Thank you for the quick review @Synchro. In the past month, 2 times, I faced issues with my host where the connection with the SMTP server was working, but something was blocking the sending process from finishing. Then, it made the sending mail script take all 300 seconds, even if I set the timeout (because timeout only applies to the connection step). Since I'm using the PHPMailer class, I can't set the SMTP Timelimit directly, so the workaround I found yesterday (when the last server occurrence happened) was applying the same timeout value, which makes sense to me, and it worked.

Regarding the default value for the set method, I agree. I've just copied from the existing setTimeout, but now I have updated both to make the value required. Since the class already has a default value for both, having this optional param doesn't make sense to me.

Synchro:
> Since I'm using the PHPMailer class, I can't set the SMTP Timelimit directly

Yes you can:

    $mail->getSMTPInstance()->Timelimit = 123;

andrebolina:
Interesting. But it still looks misleading to me that only the Timeout var exists in the PHPMailer class and that it only refers to the connection step. The first time I faced the issue with my server, I thought the timeout not being applied was a bug. I needed to open the PHPMailer and SMTP classes to understand the whole scenario. I indeed would like to improve these aspects, some options came to my mind:
A) At least renaming it to ConnectionTimeout in PHPMailer to make explicit the purpose and effect here;
B) Doing A, plus creating a CommandsTimeout var in PHPMailer and using it in the setTimelimit implemented initially by this PR.

WDYT?

Synchro:
I'm not really much in favour of encouraging changing these values since they are stipulated by RFCs. That a shorter timeout would have helped your script return sooner is really missing the point that this is a *feature* of SMTP. It certainly means that SMTP isn't suitable for use in interactive HTTP page processing, something I've said many times (but it doesn't stop people trying). The right way to deal with it is to hand off responsibility to a local mail server that can do all the waiting around for you. PHPMailer is not an MTA! Also, if your mail server has problems, then the solution is to fix the mail server rather than breaking RFCs in the client.

andrebolina:
When I say script, I don't mean interactive HTTP page processing. And sure, whenever we have server problems, the final solution is always to fix the server side, but it is helpful to have a way to handle our services in the meantime.

I'm closing the PR, and thanks for letting me know about `$mail->getSMTPInstance()->Timelimit = 123;`. It will solve my scenario. Although I'm closing it, I'd like to reinforce that my main objective here would be to create a pattern and avoid contradictions. Because of some points in your last comment, questions came to my mind. If we should **never** break RFCs (**I agree we shouldn't by default**): 1) Why is not Timeout in the PHPMailer class a const? 2) Why aren't the Timeout and Timelimit in the SMTP class consts? 3) Why do you have a setTimeout method to easily change the RFC value but not a setTimelimit? 4) Why can I set SMTP Timeout by changing PHPMailer Timeout value but can't do the same for Timelimit? _And the question that I now know the answer, but people can still face if the problem is after connection: why I'm setting  `$mail->Timeout = X;` but it's not being applied, and `$mail->send()` still takes 300 seconds?_

Thank you for all the interaction @Synchro and for providing so helpful lib for so long!

Synchro:
I agree with most of your points - but changing these things will result in BC breaks. Some future major revision might perhaps take different approaches, but we are stuck with past decisions for now (these properties were added by Apache in 2012). One good reason for allowing timeouts to be changed is to allow them to be _increased_ rather than _decreased_ – the numbers in RFCs are minimums. That could of course be enforced in setters, but it's not any kind of priority, and you can always just set public properties. There's also the fact that the RFC specify that some individual SMTP commands have *different* timeouts, especially `DATA`.

Anyway, thanks for your efforts, and I'm glad you find PHPMailer useful.

PR #2950, opened 2023-09-05
Add Bengali translations

Manish4586:
Adds [Bengali](https://en.wikipedia.org/wiki/Bengali_language) (ISO 639-1 'BE') translations.

### DIFF
--- a/language/phpmailer.lang-as.php
+++ b/language/phpmailer.lang-as.php
@@ -1,0 +1,27 @@
+<?php
+
+/**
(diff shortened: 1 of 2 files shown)


Synchro:
Thank you for these. Unfortunately I think you started from an incomplete language file, so there are some translations missing. You can see the full list [here](https://github.com/PHPMailer/PHPMailer/blob/master/src/PHPMailer.php#L2281). If you could add those missing ones, I will merge this.

Issue #2959, opened 2023-09-29
PHP Deprecated: trim(): Passing null to parameter #1 ($string) of type string is deprecated


peti123x:
# Problem description
Using the latest PHP8.2 image and the latest version of phpmailer (6.8.1), I'm getting a deprecation warning pointing to https://github.com/PHPMailer/PHPMailer/blob/master/src/PHPMailer.php#L1574 (the `trim()` function gets a null value).

# Code to reproduce
Not sure what in the system is actually making the deprecation warning emit, but nevertheless the fact that the warning appears means the trimmed value can be `null`.

# Debug output

Synchro:
This is a little mysterious because all of `From`, `Sender`, and `ConfirmReadingTo` are initialised to empty strings, and nowhere are any of them set to `null`. However, they are public properties that can be set to `null` by calling code, which I assume is what's happening for you. I'm not too concerned about that, but I'll accept a PR if you want to fix it.

Synchro:
This is fixed in the main branch and will be in the next release

PR #2962, opened 2023-10-02
fix: null issue in trim method

HarshitSingh96:
fixes : #2959

Passing null to parameter #1 ($string) of type string is deprecated

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -1571,7 +1571,10 @@ class PHPMailer
 
             //Validate From, Sender, and ConfirmReadingTo addresses
             foreach (['From', 'Sender', 'ConfirmReadingTo'] as $address_kind) {
-                $this->{$address_kind} = trim($this->{$address_kind} ?? '');
+                if($this->{$address_kind} === null) {
+                    $this->{$address_kind} = '';
+                }
+                $this->{$address_kind} = trim($this->{$address_kind});
                 if (empty($this->{$address_kind})) {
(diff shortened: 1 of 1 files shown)


src/PHPMailer.php
```diff
@@ -1571,7 +1571,7 @@ public function preSend()
 
             //Validate From, Sender, and ConfirmReadingTo addresses
             foreach (['From', 'Sender', 'ConfirmReadingTo'] as $address_kind) {
-                $this->{$address_kind} = trim($this->{$address_kind});
+                $this->{$address_kind} = trim($this->{$address_kind} ?? '');
```
Synchro, line 1574: Unfortunately null coalesce was only introduced in PHP 7.0, and we have BC back to 5.5, so this syntax can't be used. Since these properties are not meant to be nullable, I think a better approach is to stop them being null, like this:

```php
if ($this->{$address_kind} === null) {
    $this->{$address_kind} = '';
}
```
That will solve the trim issue and also fix any unexpected nulls that crop up elsewhere.

HarshitSingh96: Yup, thanks for reviewing...
Committed the fix...

HarshitSingh96:
Cool, thanks!

HarshitSingh96:
@Synchro can you merge and close this issue?

Synchro:
Thanks

peti123x:
Thanks guys!

PR #2981, opened 2023-11-22
Add methods to update or clear custom headers by name.

frankforte:
When sending a personalized email blast, it is sometimes useful to just update the "to" address and the body (rather than re-creating the entire email object). It is also useful to update the custom "List-Unsubscribe" header since it can contain a URL that is specific to each recipient. Currently, PHPMailer only has a method to clear ALL custom headers. However, we might want to keep some custom headers and only clear a specific one.

This PR adds the ability to clear or update a specific custom header by name.

Example one: clear a custom header:

    $mail->clearCustomHeader('List-Unsubscribe');

Example two: update a custom header:

    $mail->replaceCustomHeader('List-Unsubscribe', '<https://yourdomain/unsubscribe/hash>');

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -4059,6 +4059,30 @@ class PHPMailer
     }
 
     /**
+     * Clear a specific custom header.
+     */
+    public function clearCustomHeader($name)
+    {
+        foreach ($this->CustomHeader as $k => $pair) {
+            if ($pair[0] == $name) {
+                unset($this->CustomHeader[$k]);
+            }
+        }
+    }
+
+    /**
+     * Replace a custom header.
+     */
+    public function replaceCustomHeader($name, $value)
+    {
+        foreach ($this->CustomHeader as $k => $pair) {
+            if ($pair[0] == $name) {
+                $this->CustomHeader[$k] = [$name, $value];
+            }
+        }
+    }
+
+    /**
      * Add an error message to the error container.
      *
      * @param string $msg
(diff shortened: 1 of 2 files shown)


frankforte:
PR re-submitted with formatting fixes.

Synchro:
Thanks, nice addition. It would be good to show how to use this in the mailing list example.

Synchro:
One minor thought – the loops don't break after finding a match, so if you have multiple headers with the same name (which is allowed), it will remove/replace all of them.

frankforte:
Thanks for the feedback, Marcus. Let me know which solution you prefer:

1) Add a third, optimal argument $replace to *addCustomHeader* so it works
a bit like the PHP *header()* function (in which case replaceCustomHeader
becomes redundant.  We would have to make it backward compatible, so
replace would unfortunately *not *be the default. (developers that assume
it works like *header()* might get annoyed)

2) Alternatively, we can stick with the *replaceCustomHeader* method.

In either case, we should to continue through the loop and unset any
remaining headers with the same name, so it works like *header()* for
instance:

header('x-test: 1', false);
header('x-test: 2', false);
header('x-test: 3');
print_r(headers_list());

// output
// the first two headers are replaced by a single header
Array (
    [0] => X-Powered-By: PHP/8.1.6
    [1] => x-test: 3
)




Regards,

Frank

Synchro:
I think stick with the `replaceCustomHeader` approach; it's lower impact from a BC perspective. I don't see that replacing repeated custom headers is a big problem – this is quite a niche use case, and if you have trouble, you can always clear headers and start again anyway.

PR #2992, opened 2023-12-06
Fixed incorrect past participle agreement in phpmailer.lang-fr.php

### LINKED ISSUE #2993: Incorrect past participle agreement in phpmailer.lang-fr.php
Please check these things before submitting your issue:

- [x] *Read the error message* you're seeing - it often tells you what is wrong, and may contain useful links & instructions
- [x] Make sure you're using the latest version of PHPMailer
- [x] Check that your problem is not dealt with in [the troubleshooting guide](https://github.com/PHPMailer/PHPMailer/wiki/Troubleshooting), especially if you're having problems connecting to Gmail or GoDaddy
- [x] Include sufficient code to reproduce your problem
- [x] If you're having an SMTP issue, include the debug output generated with `SMTPDebug = 2` set
- [x] If you have a question about how to use PHPMailer (rather than reporting a bug in it), tag a question on Stack Overflow with `phpmailer`, but [**search first**](https://stackoverflow.com/questions/tagged/phpmailer)!

# Problem description

There's a grammatical error in the french localization file. It's for the error I encounter most often and it's been bugging me for some time.

As is, the message reads "La fonction SMTP connect() a échouée". "échouée" should be "échoué", because, when using the "avoir" auxiliary, the past participle agrees with the direct object, and never the subject. In the absence of a direct object (as is the case here), it shouldn't agree with anything, and the base form should be used.

[Here](https://french.kwiziq.com/revision/grammar/special-cases-when-the-past-participle-agrees-in-number-and-gender-when-used-with-avoir-in-le-passe-compose-conversational-pastl) is an english-language source for my claim.

Blue3957:
Fixes #2993 

There's a grammatical error in the french localization file. It's for the error I encounter most often and it's been bugging me for some time.

As is, the message reads "La fonction SMTP connect() a échouée". "échouée" should be "échoué", because, when using the "avoir" auxiliary, the past participle agrees with the direct object, and never the subject. In the absence of a direct object (as is the case here), it shouldn't agree with anything, and the base form should be used.

[Here](https://french.kwiziq.com/revision/grammar/special-cases-when-the-past-participle-agrees-in-number-and-gender-when-used-with-avoir-in-le-passe-compose-conversational-pastl) is an english-language source for my claim.


Synchro:
Merci @Blue3957!

PR #2998, opened 2024-01-01
Update PHPMailerAutoload.php

SoumitraDas1:
I have got an error:
Fatal error: __autoload() is no longer supported, use spl_autoload_register() instead in D:\xampp\htdocs\test\PHPMailer\PHPMailerAutoload.php on line 45

Then I edit line no 45:
function spl_autoload_register($classname)

Now my error is fixed. 
This reason for this error is
1. SPL autoloading was introduced in PHP 5.1.2
2.  __autoload() is no longer supported So just change the function name

### DIFF
--- a/PHPMailerAutoload.php
+++ b/PHPMailerAutoload.php
@@ -42,7 +42,7 @@ if (version_compare(PHP_VERSION, '5.1.2', '>=')) {
      * Fall back to traditional autoload for old PHP versions
      * @param string $classname The name of the class to load
      */
-    function __autoload($classname)
+    function spl_autoload_register($classname)
     {
         PHPMailerAutoload($classname);
     }


jrfnl:
Hmm... have you read the code around the code you changed ? This is a fallback for old PHP, which can probably be removed, but should definitely **_not_** be changed to `spl_autoload_register()`.

I wonder what is going on with your codebase that this line would get hit as it should only be executed if `PHP_VERSION` < 5.1.2...

Synchro:
This is also in a branch that is no longer supported, so I'm not accepting PRs for it. This issue will only occur if you're running on a newer version of PHP anyway, in which case you should be running a newer version of PHPMailer that doesn't have this issue (the `PHPMailerAutoload.php` file no longer exists).

PR #3000, opened 2024-01-03
Add support for RFC6530, using it only when required.

arnt:
This adds the ability to send email to addresses like grå@grå.org, but preserves phpmailer's old behaviour for all addresses that worked before (such as info@grå.org).

There's one new validator, which is like the W3c validator, but extended to cover EAI. It's sort of based on Firefox' extension of the same validator.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -661,6 +661,14 @@ class PHPMailer
     protected $ReplyToQueue = [];
 
     /**
+     * Whether the need for SMTPUTF8 has been detected. Set by
+     * preSend() if necessary.
+     *
+     * @var bool
+     */
+    public $UseSMTPUTF8 = false;
+
+    /**
      * The array of attachments.
      *
      * @var array
@@ -1363,6 +1371,7 @@ class PHPMailer
      * * `pcre` Use old PCRE implementation;
      * * `php` Use PHP built-in FILTER_VALIDATE_EMAIL;
      * * `html5` Use the pattern given by the HTML5 spec for 'email' type form input elements.
+     * * `eai` Use a pattern similar to the HTML5 spec for 'email' and to firefox, extended to support EAI (RFC6530).
      * * `noregex` Don't use a regex: super fast, really dumb.
      * Alternatively you may pass in a callable to inject your own validator, for example:
      *
@@ -1432,6 +1441,17 @@ class PHPMailer
                 return (bool) preg_match(
                     '/^[a-zA-Z0-9.!#$%&\'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}' .
                     '[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/sD',
+                    $address
+                );
+            case 'eai':
+                /*
+                 * This is the pattern used in the HTML5 spec for validation of 'email' type form input elements, modified to accept unicode email addresses. This is also more lenient than Firefox' html5 spec, in order to make the regex faster.
+                 *
+                 * @see https://html.spec.whatwg.org/#e-mail-state-(type=email)
+                 */
+                return (bool) preg_match(
+                    '/^[a-zA-Z0-9.!#$%&\'*+\/=?^_`{|}~\x80-\xff-]+@[a-zA-Z0-9\x80-\xff](?:[a-zA-Z0-9\x80-\xff-]{0,61}' .
+                    '[a-zA-Z0-9\x80-\xff])?(?:\.[a-zA-Z0-9\x80-\xff](?:[a-zA-Z0-9\x80-\xff-]{0,61}[a-zA-Z0-9\x80-\xff])?)*$/sD',
                     $address
                 );
             case 'php':
@@ -1567,9 +1587,24 @@ class PHPMailer
             $this->error_count = 0; //Reset errors
             $this->mailHeader = '';
 
+            //The code below tries to support full use of unicode,
+            //while remaining compatible with legacy SMTP servers to
+            //the greatest degree possible: If the message uses
+            //unicode in the localparts of any addresses, it is sent
+            //using SMTPUTF8. If not, it it sent using
+            //pynycode-encoded domains and plain SMTP.
+            if (static::CHARSET_UTF8 === strtolower($this->CharSet) &&
+                ($this->anyAddressHasUnicodeLocalpart($this->RecipientsQueue) ||
+                 $this->anyAddressHasUnicodeLocalpart(array_keys($this->all_recipients)) ||
+                 $this->anyAddressHasUnicodeLocalpart($this->ReplyToQueue) ||
+                 $this->addressHasUnicodeLocalpart($this->From))) {
+                $this->UseSMTPUTF8 = true;
+            }
             //Dequeue recipient and Reply-To addresses with IDN
             foreach (array_merge($this->RecipientsQueue, $this->ReplyToQueue) as $params) {
-                $params[1] = $this->punyencodeAddress($params[1]);
+                if (!$this->UseSMTPUTF8) {
+                    $params[1] = $this->punyencodeAddress($params[1]);
+                }
                 call_user_func_array([$this, 'addAnAddress'], $params);
             }
             if (count($this->to) + count($this->cc) + count($this->bcc) < 1) {
@@ -2163,6 +2198,7 @@ class PHPMailer
         $this->smtp->setDebugLevel($this->SMTPDebug);
         $this->smtp->setDebugOutput($this->Debugoutput);
         $this->smtp->setVerp($this->do_verp);
+        $this->smtp->setSMTPUTF8($this->UseSMTPUTF8);
         if ($this->Host === null) {
             $this->Host = 'localhost';
         }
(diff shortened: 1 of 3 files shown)


src/PHPMailer.php
```diff
@@ -660,6 +660,14 @@ class PHPMailer
      */
     protected $ReplyToQueue = [];
 
+    /**
+     * Whether the need for SMTPUTF8 has been detected. Set by
+     * preSend() if necessary.
+     *
+     * @var bool
+     */
+    public $UseSMTPUTF8 = false;
```
Synchro, line 669: One key problem that has stopped me pursuing SMTPUTF8 support in PHPMailer is that you can't tell ahead of time if the server supports it, so while you might provide UTF8 addresses that are valid during message construction, if the server doesn't support it it's not going to work. [RFC6531](https://www.rfc-editor.org/rfc/rfc6531#section-3.2) says the downgrade method is up to the MSA (i.e. PHPMailer), but how it should be handled is not defined, and it's not a simple problem to solve. Unfortunately the only reliable approach is to avoid using it altogether, which is where I've been sitting!

src/PHPMailer.php
```diff
@@ -4271,6 +4307,35 @@ public static function isValidHost($host)
...
     }
 
+    /**
+     * Check whether the supplied address uses unicode in the localpart.
+     *
+     * @return bool
+     */
+    public function addressHasUnicodeLocalpart($address)
+    {
+        return (bool) preg_match( '/[\x80-\xFF].*@/', $address);
```
Synchro, line 4317: Is this a reliable pattern? An address can contain multiple @ symbols if they are quoted, so this might match an address like `"user@host@näme"@example.com"` more by luck than design. I can't think of a non-matching address right now though.

src/PHPMailer.php
```diff
@@ -4271,6 +4307,35 @@ public static function isValidHost($host)
         return filter_var('http://' . $host, FILTER_VALIDATE_URL) !== false;
     }
 
+    /**
+     * Check whether the supplied address uses unicode in the localpart.
+     *
+     * @return bool
+     */
+    public function addressHasUnicodeLocalpart($address)
```
Synchro, line 4315: I'd prefer these new methods to be `protected` - it's not necessary to make them part of the public API as they are only intended for internal use.

src/SMTP.php
```diff
@@ -905,7 +914,15 @@ protected function parseHelloFields($type)
      * $from. Returns true if successful or false otherwise. If True
      * the mail transaction is started and then one or more recipient
      * commands may be called followed by a data command.
-     * Implements RFC 821: MAIL <SP> FROM:<reverse-path> <CRLF>.
+     * Implements RFC 821: MAIL <SP> FROM:<reverse-path> <CRLF> and
+     * two extensions, namely XVERP and SMTPUTF8.
+     *
+     * The server's EHLO response is not checked. If use of either
```
Synchro, line 920: I mentioned this earlier. You can check if the mail server supports an extension by looking in the `SMTP::$server_caps` array after EHLO has happened, like this:

    array_key_exists('SMTPUTF8', $this->server_caps)

test/PHPMailer/PHPMailerTest.php
```diff
@@ -1191,6 +1191,46 @@ public function testDuplicateIDNRemoved()
...
+    public function testSmtpUTF8()
+    {
+        //No reason to use SMTPUTF8
+        $this->Mail->isSMTP();
+        $this->Mail->addAddress('foo@example.com', '');
+        $this->Mail->preSend();
+
+        //Using a punycoded domain is enough
+        self::assertFalse($this->Mail->UseSMTPUTF8);
+        $this->Mail->addAddress('foo@motörhead.example', '');
```
Synchro, line 1215: I like to avoid characters like ö for testing UTF-8 support because they are also valid in ISO8859 charsets. The classic example is `Iñtërnâtiônàlizætiøn`, which is quite useless as all of its chars are also in 8859-1! Maybe something radically non-ISO, like 😱?

Synchro (commented):
Thanks for this – SMTPUTF8 support is something that's been wanted for a while, but I've been avoiding as I have not wanted to deal with the support issues it will inevitably cause when connecting to servers that don't support it, and also because I've wanted to avoid a half-hearted solution that only supports UTF-8 in addresses, not in all headers (which is what Symfony Mailer does).

However, I'm willing to persevere with your approach, especially since you wrote tests too :)

test/PHPMailer/PHPMailerTest.php
jrfnl: Let me know if you need some more test data (or have a look here for some samples you can build on: https://github.com/PHPCSStandards/PHPCSUtils/blob/develop/Tests/Utils/Orthography/FirstCharTest.php)

Talking about the tests - those could maybe do with some expanding / these methods getting their own test file ? With a `@covers` tag too ?

src/PHPMailer.php
```diff
@@ -1434,6 +1443,17 @@ public static function validateAddress($address, $patternselect = null)
                     '[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/sD',
                     $address
                 );
+            case 'eai':
+                /*
+                 * This is the pattern used in the HTML5 spec for validation of 'email' type form input elements, modified to accept unicode email addresses. This is also more lenient than Firefox' html5 spec, in order to make the regex faster.
```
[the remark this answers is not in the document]
arnt: Oh, did you? In that case I wouldn't mind having a chat about how it might be updated to cover EAI, if you have a spare half hour sometime. That github issue isn't going anywhere.

src/PHPMailer.php
arnt: That part of 6531 is IMO not quite as well written as it could be.

When you walk through the various cases, the only one where a downgrade is both possible and better than an available alternative is MSA-time downgrade of the sender address. That's not a terribly relevant case for phpmailer (like for for mikel-mail, smtplib, javamail, or Sendgrid, Mailchimp, etc).

IMO, the kind of software that typically calls phpmailer is best served by using noreply@asciidomain as sender address and trusting the caller to supply the actually desired recipient address. Some phpmailer users might be seen as MSAs that could respond to errors by picking a different sender address, but that decision is IMO solidly outside phpmailer's scope.

Perhaps I should comment about how others do it... Ruby, Python and Java all have libraries that raise exceptions when something doesn't work. If the server refuses to play because we don't have the right password, they throw an exception. If the server won't relay because of a misconfiguration, another exception. The javamail/net-smtp/smtplib user is expected to catch the exceptions. I grepped around github to see what callers responded to those exceptions and didn't find any that distinguished between missing SMTPUTF8 and other exceptions.

So if it's bulk mail being generated from a database query, missing SMTPUTF8 will just add the relevant address to the list of undeliverable addresses. If it's a contact form, missing SMTPUTF8 will show up as the same error as if the server rejects the message (e.g., Postfix can reject unknown domains at injection time).

src/PHPMailer.php
arnt: I believe it to be correct for all valid addresses. `.*` is greedy, so `[\x80-\xff]` will match any non-ASCII byte before the last `@` in the address.

If you feed it garbage (such as a grå@grå.org encoded as cp865 instead of utf8), it'll return true even if the localpart doesn't contain unicode (it contains cp865). GIGO; cp865 isn't valid in email addresses. I don't think the PHP language/runtime gives us a really great option in this case.

src/PHPMailer.php
arnt: Will do. Do you prefer force-push or another commit?

Synchro: I'm fine with force push for PRs.

test/PHPMailer/PHPMailerTest.php
arnt: What, you're not a Motörhead fan? I'll change to Spın̈al Tap! The n̈ is in unicode, but not in any 8859 variant.

I'd rather not use emoji or anything else that's blocked in public TLDs, bad examples leads to FAQs for me at work. (Emoji are banned in general, because second-level domains should be distinguishable and many emoji pairs look confusingly similar.)

src/SMTP.php
arnt: The check is easy, the next line is the question. I considered throwing an exception, but decided against it for the moment.
 - I didn't want to have SMTPUTF8 behave differently from XVERP in that function
 - Changing XVERP would change existing functionality (that noone should depend on, but...)
 - Callers can disable exceptions anyway

I could of course call setError(), like how sendCommand() would do when the `MAIL FROM` fails. Shall I?

arnt:
Hi,

I force-pushed with most things fixed. Remaining issues:
1. The code doesn't check whether SMTPUTF8 is supported before using it, to be similar to XVERT and because it doesn't make a difference (it only saves one roundtrip before throwing an exception).
2. There's no attempt at downgrading. There can't very well be, phpmailer isn't given enough information by its caller.

I could change both of these at once… I could add a callback with which phpmailer would ask its caller to downgrade. The default value would be a function that returns `<>` if it's the sender and throws an exception if it's a recipient address. Of course I'd add unit tests to make sure ascii@idn works. Does that sound good do you? Do you think you might merge that?

arnt:
The more I think about it, the better I like that approach.

PrOgramm3r1:
Unsubscribe

mschandrau:
l,

kmk142789 (approved):

arnt:
I'll close this pull request since the same code was merged in #3104 (with signed commits and a little extra code).

PR #3020, opened 2024-02-19
Fix cannot access protected method

Yurunsoft:


Synchro:
Thanks. It would have been helpful to have a failing test showing why this was needed.

PR #3027, opened 2024-03-02
Add urdu language support

saqibsra:


saqibsra:
Add urdu language file

PR #3032, opened 2024-03-12
[smtp_code_ex] bad encoding

nenge123:
add `mb_convert_encoding` convert


Synchro:
Why is this needed?

Separately, I avoid using `function_exists` because it's often blocked in shared hosting PHP configs; check for a defined constant instead.

joangava (approved):

joangava (approved):

joangava (approved):

Issue #3031, opened 2024-03-08
Add the possibility of changing @phpmailer.0 string in asHtml method when composing image cid.


Matslom:
Hello,

There is a hard coded string in https://github.com/PHPMailer/PHPMailer/blob/master/src/PHPMailer.php#L4418 and https://github.com/PHPMailer/PHPMailer/blob/master/src/PHPMailer.php#L4452 which leaks server/app details. For security reasons, we should be able to change this to a custom one, like other headers.

Currently, only way to do that, is to override the entire msgHTML method. Is there a chance to introduce a more flexible approach in the library?

Synchro:
How do you see this as a risk? What threat model do you foresee as applying to it? Neither of these uses are headers. Also bear in mind that `msgHTML` is purely a convenience function that you don't need to use at all. How would you change the ID generation, bearing in mind that even a random string would be readily fingerprintable?

Synchro:
I fixed some small issues in the PR, it's now merged, will be in the next release.

PR #3036, opened 2024-03-15
Add cid default domain

mariuszkrzaczkowski:
In reference to issue [3031](https://github.com/PHPMailer/PHPMailer/issues/3031), I have prepared a patch


Synchro:
This is simple, but it opens the door for people setting it to an empty string, something too long, containing spaces, or other invalid value, which would result in invalid cids. I'd suggest either making it protected and using a setter method, or validating the value before using it, and if it's invalid, reverting to a fixed (but valid) string. Also, this needs test coverage.

Synchro:
Another thought: this issue only occurs inside `msgHTML()`, so it would seem sensible to add it as an optional param to that method. That way we don't need a new property (we already have too many), and can inline the validation for it, so don't need a new method either.

mariuszkrzaczkowski:
I've added some corrections, check them out

Synchro:
While that's much neater, the default of `a` for the domain is as identifiable as the original `phpmailer.0` default so it's not solving much from a security perspective (though obviously it is if you populate the parameter). Also I recall that there was something that required a `.` within the domain part, so that it would parse as a domain, but didn't need to actually exist, hence the current `.0`.

mariuszkrzaczkowski:
To make it work automatically, maybe let's set $cid_domain to the sender's domain from $this->From ?

mariuszkrzaczkowski:
maybe something like that?
```php
    public function msgHTML($message, $basedir = '', $advanced = false)
    {
        $cid_domain = '@phpmailer.0';
        if (filter_var($this->From, FILTER_VALIDATE_EMAIL)) {
            //prepend with a character to create valid RFC822 string in order to validate
            $cid_domain = substr( $this->From, strpos( $this->From, '@') + 1);
        }

        preg_match_all('/(?<!-)(src|background)=["\'](.*)["\']/Ui', $message, $images);
```

mariuszkrzaczkowski:
Any comment on this?

Synchro:
Apologies for the late response. That's a much better idea, and I'm looking at doing that.

joangava (approved):

joangava (approved):

Shinogreat (approved):

kmk142789 (approved):

mariuszkrzaczkowski:
there is acceptance, so it will be merged?

Synchro:
No, because the implementation that's in here still has problems, the suggested (better) alternative is not in the PR, and it's not covered by tests. You could fork the PR if you want to improve it.

Separately, splitting the domain from the location part using `strpos( $this->From, '@')` will not work as an address can contain multiple `@` characters; `strrpos( $this->From, '@')` (starting from the right-hand side) is safer.

mariuszkrzaczkowski:
multiple `@` characters will not pass filter_var

mariuszkrzaczkowski:
I corrected PR

Synchro:
> multiple `@` characters will not pass filter_var

They need to be quoted, like `"a@b"@example.com`, which passes filter_var. You can see that using `strpos` to split that would extract `b"@example.com` as the domain, which is wrong.

mariuszkrzaczkowski:
corrected PR

jrfnl:
Hmm.. if I look at the diff for this PR, it looks like some merge-backs or rebases went wrong... ?

Synchro:
> Hmm.. if I look at the diff for this PR, it looks like some merge-backs or rebases went wrong... ?

No, I think it's just that there have been a lot of changes in master since the fork, and I just merged them in.

jrfnl:
> > Hmm.. if I look at the diff for this PR, it looks like some merge-backs or rebases went wrong... ?
> 
> No, I think it's just that there have been a lot of changes in master since the fork, and I just merged them in.

But merging those in, should remove them from the diff in this PR, Not add them to it (as they are already in `master`).

Synchro:
Don't know? It ends up with the same set of changes either way? All I did was merge master into this branch.

jrfnl:
Thanks @mariuszkrzaczkowski for fixing that diff!

mariuszkrzaczkowski:
no problem, it's great that you managed to solve the reported problem

Issue #3045, opened 2024-04-05
edebug() should probably also get ride of "\r\n" when using LoggerInterface


axi:
# Problem description
Using LoggerInterface, I don't see why we should have the SMTP::LE chars passed to the debug method

``` 
    protected function edebug($str)
    {
       ...
       if ($this->Debugoutput instanceof \Psr\Log\LoggerInterface) {
            $this->Debugoutput->debug($str);

            return;
        }
``` 
In my case, with Monolog\Formatter\LineFormatter and a $allowInlineLineBreaks = true, I was seeing logs with an empty line between each log line

```  
[2024-03-14T15:44:05.083183+00:00] logger.DEBUG: CLIENT -> SERVER: 
  
[2024-03-14T15:44:05.083250+00:00] logger.DEBUG: CLIENT -> SERVER: .
  
[2024-03-14T15:44:05.247371+00:00] logger.DEBUG: CLIENT -> SERVER: QUIT
```  

Setting $allowInlineLineBreaks to false solved it but I don't thing this is usefull.

Using same preg_replace as 'html' and 'echo' cases would give this:
``` 
       if ($this->Debugoutput instanceof \Psr\Log\LoggerInterface) {
            $this->Debugoutput->debug(preg_replace('/[\r\n]+/', '', $str));

            return;
        }
``` 

Should I propose a PR for that ?

Synchro:
Yes, that seems reasonable, go ahead with a PR.

PR #3047, opened 2024-04-08
Remove potential ending line break when using edebug() with LoggerInterface

axi:
This PR is for #3045.

### DIFF
--- a/src/SMTP.php
+++ b/src/SMTP.php
@@ -280,6 +280,8 @@ class SMTP
         }
         //Is this a PSR-3 logger?
         if ($this->Debugoutput instanceof \Psr\Log\LoggerInterface) {
+            //Remove ending line breaks potentialy added by calls to SMTP::client_send()
+            $str = preg_replace('/[' . static::LE .']+$/', '', $str);
             $this->Debugoutput->debug($str);
 
             return;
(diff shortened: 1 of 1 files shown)


src/SMTP.php
```diff
@@ -280,6 +280,8 @@ protected function edebug($str, $level = 0)
         }
         //Is this a PSR-3 logger?
         if ($this->Debugoutput instanceof \Psr\Log\LoggerInterface) {
+            //Remove ending line breaks potentialy added by calls to SMTP::client_send()
+            $str = preg_replace('/[' . static::LE .']+$/', '', $str);
```
Synchro, line 284: Instead of that `static::LE`, this should use an explicit `\r\n` as we are matching individual meta characters in a character class, not a literal string. `static::LE` contains a literal `"\r\n"` which represents actual CR and LF chars, not their meta-equivalents. Basically do what the other debug output options do.

Synchro: This would also fail the code style check, which requires spaces either side of string concat `.` chars, but that's academic as you won't need to do concat with the above change.

src/SMTP.php
```diff
@@ -280,6 +280,8 @@ protected function edebug($str, $level = 0)
         }
         //Is this a PSR-3 logger?
         if ($this->Debugoutput instanceof \Psr\Log\LoggerInterface) {
+            //Remove ending line breaks potentialy added by calls to SMTP::client_send()
+            $str = preg_replace('/[\r\n]+$/', '', $str);
```
jrfnl, line 284: Just wondering - could/should this be simplified to `rtrim($str, "\r\n")` ?

Synchro: Yep, that seems reasonable – and it could be used in the other cases as well. I expect it will be marginally faster, but I don't think optimising debugging output is really something to worry about!

PR #3090, opened 2024-09-11
Composer: prevent a lock file from being created

jrfnl:
Composer 1.10.0 introduced a `lock` config option, which, when set to `false` will prevent a `composer.lock` file from being created and will ignore it when one exists.

This is a useful option for packages like PHPMailer where the `lock` file has no meaning.

It also makes life more straight-forward for contributors as they don't have to remember that for this repo they should use `composer update` instead of `composer install`. Both will now work the same.

Refs:
https://getcomposer.org/doc/06-config.md#lock

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,8 @@
     "config": {
         "allow-plugins": {
             "dealerdirect/phpcodesniffer-composer-installer": true
-        }
+        },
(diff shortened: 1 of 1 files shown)


Synchro:
Good idea, thank you!

PR #3091, opened 2024-09-11
DKIMTest: two fixes for tests for when OpenSSL is disabled/unavailable

jrfnl:
### DKIMTest: make exception expectation more specific

As things were, the `DKIMTest::testDKIMSignOpenSSLNotAvailableException()` test _could_ potentially pass even when another `Exception` than the `PHPMailer\PHPMailer\Exception` was being thrown, as _all_ exceptions extend the PHP native `Exception` class.
Now this risk is not that high, as there is also a check on the exception message, but still.

Making the exception expectation more specific (by changing the import `use` statement), should still make the test more stable.

### DKIMTest::testDKIMSignOpenSSLNotAvailable(): fix the test

As things were, the `DKIMTest::testDKIMSignOpenSSLNotAvailable()` could not pass as the `DKIMTest` class sets the `USE_EXCEPTIONS` class constant to `true`, which means the method would fail on an exception.

As this test is specifically about testing the behaviour when exceptions are _disabled_, the test needs to be in its own test class, which sets `USE_EXCEPTIONS` to `false`.

That should allow the test to run properly and to pass.

### DIFF
--- a/test/PHPMailer/DKIMTest.php
+++ b/test/PHPMailer/DKIMTest.php
@@ -241,21 +241,6 @@ final class DKIMTest extends SendTestCase
     }
 
     /**
-     * Verify behaviour of the DKIM_Sign method when Open SSL is not available.
-     *
-     * @covers \PHPMailer\PHPMailer\PHPMailer::DKIM_Sign
-     */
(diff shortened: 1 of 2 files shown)


Synchro:
Thanks

PR #3096, opened 2024-09-18
GH Actions: test with different sets of extensions

jrfnl:
PHP 8.4 removes the IMAP extension (moved to PECL).

With this in mind, I've reviewed how the tests are being run versus the extension requirements and recommendations.

As things are, the tests are currently run in the "ideal" environment, i.e. with all required and optional extensions available.

However, the codebase also contains fall-backs for when certain extensions are **_not_** available and for at least some of those fallbacks, there are dedicated tests available, but in an ideal environment those tests will not run and the fall-backs are not tested, which is the case with the current CI setup.

To improve this situation, I'm proposing to keep running the tests against all PHP versions with the "ideal" extension set, but to also have additional test runs with a far more limited set of PHP extensions.

To determine which extensions should be in each set, I've looked at the following:
* `@requires` tags found in the test suite and the conditions for calls to `markTestSkipped()`. This brought to light that the `openssl` extension was currently not listed in the "ideal" extension set. This has now been fixed.
* The required extensions of PHPUnit - `dom, json, libxml, mbstring, tokenizer, xml, xmlwriter`.
* The required extensions of PHPMailer itself - `ctype, filter, hash`.
* Not strictly required, but more for convenience/workflow speed: `curl` for Composer.
* And `xdebug` will still be enabled/disabled based on the `coverage` setting.

Note: while some tests would benefit from being run _without_ the `mbstring` extension, that's unfortunately not an option as `mbstring` is a requirement of PHPUnit :shrug:

Also note, the tests with the "minimal" extension setup needs to run `composer install` with an `--ignore-platform-req` flag to prevent running into the following issue:
```
Running update with --no-dev does not mean require-dev is ignored, it just means the packages will not be installed. If dev requirements are blocking the update you have to resolve those problems.
```

As this extension "requirement" is for a dependency which is not used in the test run, the extension requirement can be safely ignored.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -88,23 +88,39 @@ jobs:
     strategy:
       matrix:
         php: ['5.6', '7.0', '7.1', '7.2', '7.3', '7.4', '8.0', '8.1', '8.2']
+        extensions: ['optimal', 'minimal']
         coverage: [false]
         experimental: [false]
         include:
           # Run code coverage on high/low PHP.
           - php: '5.5'
+            extensions: 'optimal'
+            coverage: true
+            experimental: false
+          - php: '5.5'
+            extensions: 'minimal'
             coverage: true
             experimental: false
           - php: '8.3'
+            extensions: 'optimal'
+            coverage: true
+            experimental: false
+          - php: '8.3'
+            extensions: 'minimal'
             coverage: true
             experimental: false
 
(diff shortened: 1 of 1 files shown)


jrfnl:
@Synchro Got any questions about this PR ?

Synchro:
No, it's a good idea, thank you.

PR #3098, opened 2024-09-24
Modifying READE.md file

Nats-hope:
Making it more clear


Synchro:
This really doesn’t clarify anything.

PR #3104, opened 2024-10-15
Add SMTPUTF8 support

Synchro:
A new discussion based on changes in [this PR](https://github.com/PHPMailer/PHPMailer/pull/3000).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -151,7 +151,7 @@ jobs:
             echo 'COMPOSER_OPTIONS=' >> $GITHUB_OUTPUT
           else
             # Minimal.
-            echo 'EXT=none, curl, dom, json, libxml, tokenizer, xml, xmlwriter, ctype, filter, hash' >> $GITHUB_OUTPUT
+            echo 'EXT=none, curl, dom, json, libxml, mbstring, tokenizer, xml, xmlwriter, ctype, filter, hash' >> $GITHUB_OUTPUT
             echo 'COMPOSER_OPTIONS=--ignore-platform-req=ext-simplexml' >> $GITHUB_OUTPUT
           fi
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -151,7 +151,7 @@ jobs:
             echo 'COMPOSER_OPTIONS=' >> $GITHUB_OUTPUT
           else
             # Minimal.
-            echo 'EXT=none, curl, dom, json, libxml, mbstring, tokenizer, xml, xmlwriter, ctype, filter, hash' >> $GITHUB_OUTPUT
+            echo 'EXT=none, curl, dom, json, libxml, tokenizer, xml, xmlwriter, ctype, filter, hash' >> $GITHUB_OUTPUT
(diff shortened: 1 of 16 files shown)


Synchro:
@arnt I've moved discussion here to have a fresh start and get the tests passing cleanly.

Things I'm concerned about:

* If `PHPMailer::$UseSMTPUTF8` is a public property, users can set or unset it arbitrarily, likely causing problems. I think it should be protected and setting it could be either fully automatic (e.g. if you add an address that uses local-part UTF-8), or explicitly enabled with a `UseSMTPUTF8` method that could also set `CharSet` and the `eid` validation pattern. Calling the method without params could return its current state, which will be useful if it has been enabled automatically. One problem with that approach is that it's not quite consistent with how `isSMTP` works.
* We could make validation fail if the `eid` pattern is selected but `CharSet` is not UTF-8, even if it would otherwise pass in an effort to stop people "using it wrong".
* The `testUnsupportedSmtpUTF8` test passes for the wrong reason, not because it's using ISO-8859-1 (which is the default anyway), but because the UTF-8 validator has not been set. It still passes when `CharSet` is UTF-8.

Quite a few problems are caused by `CharSet` being a public property that can be changed *after* setting values, e.g. set an ISO-8859-1 subject, and then set `CharSet` to UTF-8, resulting in corrupt content. This is why there is all the complication about enqueuing addresses for puny encoding as we can't be sure what `CharSet` will be until send time. That's a separate problem though really.

Further thoughts / ideas?

Synchro:
One thing about DX: as it stands, the only way to say that you explicitly want to use SMTPUTF8 is to set `PHPMailer::$validator = 'eai'`, which isn't very intuitive. This is partly why I suggesting having a method like `useSMTPUTF8` or `allowUnicodeAddresses` as it's much more obvious what it means.

arnt:
I pushed changes that sets UseSMTPUTF8 automatically and sends unencoded UTF8 text.

AIUI you like unencoded UTF8 text. I'm very happy to send that, it just looks better, both in the subject and in addresses.

I'm less than entirely happy with the UseSMTPUTF8 logic. Settting it automatically like this has major benefits, but it looks a little magic. I actually thought about dropping the 'eai' validator and changing the 'php' validator to work differently if the Mailer says it supports SMTPUTF8 (qmail and sendmail were updated years ago, SMTP is observable).

Synchro:
Thanks for those changes, they look sensible, and unencoded UTF-8 is definitely _the way_.

Any ideas for how to minimise magic? I wonder if @jrfnl has an anti-wizard hat?

Something occurred to me about punycoded addresses. The only place they matter is in headers, and the headers are generally not subject to whatever `CharSet` is set to because that only really applies to the body. If we are not doing SMTPUTF8, the headers are all ASCII, and anything outside that gets RFC2047-encoded as Q or B strings. However, it would be relatively harmless to *always* encode those headers as UTF-8, even if that's not what `CharSet` is set to, and that would mean that we could skip the whole enqueuing business for puny code addresses, which would make things much easier to validate before sending.

I suspect this is somewhat outside the scope of this PR though...

PR #3115, opened 2024-11-22
GH Actions: PHP 8.4 has been released

jrfnl:
* Builds against PHP 8.4 are no longer allowed to fail.
* Add _allowed to fail_ builds against PHP 8.5.

Ref: https://www.php.net/releases/8.4/en.php

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -46,10 +46,10 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        php: ['5.5', '7.2', '8.0', '8.1', '8.2', '8.3']
+        php: ['5.5', '7.2', '8.0', '8.4']
         experimental: [false]
         include:
-          - php: '8.4'
+          - php: '8.5'
             experimental: true
 
     name: "Lint: PHP ${{ matrix.php }}"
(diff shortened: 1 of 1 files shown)


jrfnl:
Note: I kind of suspect the "minimal" builds will fail on PHP 8.4. If that's the case, this will be related to this issue: https://github.com/shivammathur/setup-php/issues/882

.github/workflows/tests.yml
```diff
@@ -46,10 +46,10 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        php: ['5.5', '7.2', '8.0', '8.1', '8.2', '8.3']
+        php: ['5.5', '7.2', '8.0', '8.4']
```
jrfnl, line 49: Hmm.. linting seemed to have grown out of bounds a little over time. It was meant to only lint against high/low of each supported PHP major. No need for additional runs as the tests are run against all supported PHP versions anyhow.

This bring the lint build back to the original intention.

Synchro:
Thanks. I'd already made a start on this [in here](https://github.com/PHPMailer/PHPMailer/tree/php-8.4), and the changes are largely the same as what you've done here, but I didn't know what the next version was to be called, so I'd not dealt with that bit. I'll use your changes as it's more complete than mine.

jrfnl:
@Synchro Great minds think alike ? ;-) (sorry, didn't see you were already working on this)

Issue #2994, opened 2023-12-12
sigterm when in DATA state causes QUIT to fail


Stevehans:
# Problem description

If my code is killed (e.g. with a SIGTERM), then the destructor calls smtpClose() which then calls smtp->quit().

But if data is being sent, then the quit comment is ignored.

I think we need a status to indicate that we are in the DATA mode, and send the DATA END before the QUIT?

Synchro:
Hm. Tricky! I don't think sending a DATA END is appropriate there either, because the message sending is incomplete, so I think it would be expected to simply drop the connection instead of cleaning up nicely - the server will deal with it. Unless you are proposing to wait until the send is finished?

Stevehans:
No, I think you're right, at present I've added some code to my copy that sets inData to either true or false in the data() method, and if true it skips sending the QUIT and just closes the connection.

I think that's the right action, as part delivering a partial email is definitely not in spec.

Synchro:
I'd suspect it would apply to any SMTP command, not just DATA, it's just that DATA is the one most likely to be in progress if a send takes more than a few milliseconds. That said, DATA (maybe AUTH?) is the only one that involves more than one call to `fwrite`, which is about all the granularity we can do.

PR #3121, opened 2024-12-13
Update SMTP.php

Stevehans:
Resolves issue 2994

### DIFF
--- a/src/SMTP.php
+++ b/src/SMTP.php
@@ -265,6 +265,12 @@ class SMTP
     protected $last_reply = '';
 
     /**
+     * Whether we are in the DATA phase or not
+     * @var bool
+     */
+    private $inData = false;
+
+    /**
      * Output debugging info via a user-selected method.
      *
      * @param string $str   Debug string to output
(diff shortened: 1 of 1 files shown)


anandfresh (approved):

Synchro:
I think I've got a better idea for how to solve this. As I said in #2994, I suspect that this could happen during any command, not just `DATA`, so rather than having a flag purely for that command, I think a better approach would be (since SMTP is stateful) to have a `state` property that records what the last command was. We can then use that the same way as you've done with the `inData` property to check the current state during shutdown. Thoughts?

Sridhass:
@Synchro Have you added PHPMailerAutoload in this?  Hope this might be helpful if added to this repo.  
Thanks!

Stevehans:
I think you're right, I was just going for the quick fix, but tracking state makes more sense.

On 20 Dec 2024 20:42, Marcus Bointon ***@***.***> wrote:

I think I've got a better idea for how to solve this. As I said in #2994<https://github.com/PHPMailer/PHPMailer/issues/2994>, I suspect that this could happen during any command, not just DATA, so rather than having a flag purely for that command, I think a better approach would be (since SMTP is stateful) to have a state property that records what the last command was. We can then use that the same way as you've done with the inData property to check the current state during shutdown. Thoughts?

Synchro:
> @Synchro Have you added PHPMailerAutoload in this?  Hope this might be helpful if added to this repo.  
> 
> Thanks!

Why would I do that? PHPMailerAutoload is long dead; use composer.

kmk142789 (approved):

PR #3132, opened 2025-02-05
Replace Swiftmailer with Symfony Mailer

no-one:
As Swiftmailer is end of life (https://symfony.com/blog/the-end-of-swiftmailer), it is better to refer to its successor.


Synchro:
Thanks

PR #3149, opened 2025-03-24
Allow to override the addresses that the mail is actually sent to when using SMTP.

PhilLehmann:
In SMTP, we have to possibility to define 'to', 'cc' and 'bcc' addresses - but to send the message to completely different recipients when invoking the RCPT TO command in the SMTP conversation.

Typical use case might be to implement a outlook-like distribution list scenario:
- mail is sent from alice@example.org to list@example.org
- PHP script handles the mail and sends it to the members of the distribution list, e.g. bob@example.org and chris@example.org - keeping the list@example.org in the 'to', but not sending it to list@example.org again.
- bob@example.org and chris@example.org receive the mail with list@example.org as the sole 'to' address

Thanks for considering this PR

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -335,6 +335,14 @@ class PHPMailer
     public $SMTPOptions = [];
 
     /**
+     * An array of addresses that should be used for the SMTP RCPT TO command.
+     * By default (if null or empty), the messages are sent to all 'to', 'cc' and 'bcc' addresses.
+     *
+     * @var array
+     */
+    protected $SMTPRecipientTo = null;
+
+    /**
      * SMTP username.
      *
      * @var string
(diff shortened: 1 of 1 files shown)


Synchro:
What you describe sounds just like standard list sending. I'm not sure what you gain by your approach other than adding complication in having to go out of your way to exempt the list address, and you're trying to wrap two message sends and a receive (which PHPMailer doesn't handle at all) into one, which isn't SMTP's job. Exploding the recipient list belongs in your app, not PHPMailer, and it can all be achieved with current code because what that receiver does is exactly the same as a regular BCC-only list send.

PhilLehmann:
I concur, with one exception. How can I put list@example.org in the 'to' without the mail being sent there again?

Of course, I could filter it out on the receiving end of my app. But this feels like waste.

Synchro:
There are these stages to consider:

1. Initiator sends a message to list@example.com
2. List system receives the message, gets the recipient list for sending it on to
3. List system creates a new message sent to nobody, all recipients in BCC

Note that this is two sends, not one, and PHPMailer can do both without changing anything, though it's likely that the first message would not be sent by PHPMailer anyway. The problem is in step 2, which is a higher-level concern outside PHPMailer's scope. You don't need to put the list address in `to` at all; if anything I'd expect to to appear in `from`, since you can't put the original sender's email in there without forging it (the right way is to put them in reply-to).

Issue #3150, opened 2025-03-24
Wrong line endings on PHP8.3 (Linux)


cpiber:
# Problem description

The output we're seeing is similar to #2870, except we're using standard `mail`.
The client receives emails with added newlines in the content part when sending emails with attachments (i.e. multipart).
Forcing `PHP_EOL` fixes the issue.

The [commit referenced in the blame](https://github.com/php/php-src/commit/6983ae751cd301886c966b84367fc7aaa1273b2d) changed from "\n" to "\r\n". Current code is this:
```
	char *line_sep = PG(mail_mixed_lf_and_crlf) ? "\n" : "\r\n";
```
So depending on the setting, either will be used. The change was introduced in https://github.com/php/php-src/commit/cc931af35ddc0a07c0a669eb4da9dc66e1f73af4, which seems to have landed in 8.2.4. That said, according to the settings below, I should not be running into this case, so I'm unsure of what I'm seeing.

php info:
```
PHP Version 8.3.14
System 	Linux cloud9-vm104 6.1.90+2-ph #ph SMP PREEMPT_DYNAMIC Wed May 8 11:21:28 UTC 2024 x86_64
Build Date 	Dec 5 2024 16:11:01
Build System 	Linux
Server API 	CGI/FastCGI
Virtual Directory Support 	disabled
Configuration File (php.ini) Path 	/usr/local/php8.3/lib
Loaded Configuration File 	
Scan this dir for additional .ini files 	/home/aamdxwfu
Additional .ini files parsed 	/home/aamdxwfu/php8.3.ini
PHP API 	20230831
PHP Extension 	20230831
Zend Extension 	420230831
Zend Extension Build 	API420230831,NTS
PHP Extension Build 	API20230831,NTS
Debug Build 	no
Thread Safety 	disabled
Zend Signal Handling 	enabled
Zend Memory Manager 	enabled
Zend Multibyte Support 	provided by mbstring
Zend Max Execution Timers 	disabled
IPv6 Support 	enabled
DTrace Support 	disabled
Registered PHP Streams	https, ftps, compress.zlib, compress.bzip2, php, file, glob, data, http, ftp, phar, zip
Registered Stream Socket Transports	tcp, udp, unix, udg, ssl, tls, tlsv1.0, tlsv1.1, tlsv1.2, tlsv1.3
[...]

Issue #2870, opened 2023-02-06
Wrong line endings with PHP 8.1 and nullmailer


AdrienPoupa:
We recently updated PHPMailer from an ancient 5.2 version to the latest, 6.7.1.

We have a bit of an exotic setup, with PHPMailer using the `mail` driver, then `nullmailer` handles sending the email to a remote SMTP server. This is done for performance purposes, so we don't have to wait synchronously for PHPMailer to connect to the SMTP server and block the user while it is doing it.

When sending multi `multipart/mixed` emails, the raw email looks as follows:

```
...
X-Mailer: PHPMailer 6.7.1 (https://github.com/PHPMailer/PHPMailer)
MIME-Version: 1.0
Content-Type: multipart/mixed; boundary="b1=_123"
Content-Transfer-Encoding: 8bit

--b1=_123

Content-Type: text/html; charset=UTF-8

Content-Transfer-Encoding: 8bit
```

Notice the extra new lines after the boundary between the two `Content` entries. It should be:

```
X-Mailer: PHPMailer 6.7.1 (https://github.com/PHPMailer/PHPMailer)
MIME-Version: 1.0
Content-Type: multipart/mixed; boundary="b1=_123"
Content-Transfer-Encoding: 8bit

--b1=_123
Content-Type: text/html; charset=UTF-8
Content-Transfer-Encoding: 8bit
```

We narrowed it down to this block in `preSend`:
```php
    public function preSend()
    {
        if (
            'smtp' === $this->Mailer
            || ('mail' === $this->Mailer && (\PHP_VERSION_ID >= 80000 || stripos(PHP_OS, 'WIN') === 0))
        ) {
            //SMTP mandates RFC-compliant line endings
            //and it's also used with mail() on Windows
            static::setLE(self::CRLF);
        } else {
            //Maintain backward compatibility with legacy Linux command line mailers
            static::setLE(PHP_EOL);
        }
```

Because we are running the `mailer` driver and PHP 8.1, `static::setLE(self::CRLF);` is triggered. But since we are using a legacy Linux CLI mailer, what we need is `static::setLE(PHP_EOL);`.

Now, I understand our setup is not ideal, convoluted and probably not common.

However, I would have appreciated more flexibility for setting the line endings. I had to overwrite `setLE` in a child class to prevent the CRLF call from modifying the line endings.

Maybe this block could be extracted to a smaller, overwritable function? Or this could be an editable parameter somewhere? I had something like this in mind:

```php
[...]

PR #3151, opened 2025-03-25
Wrong line endings on PHP8.3 (Linux) #3150 - Issue attempt

sudojoao:
#3150 

> The output we're seeing is similar to https://github.com/PHPMailer/PHPMailer/issues/2870, except we're using standard mail.
>The client receives emails with added newlines in the content part when sending emails with attachments (i.e. multipart).
>Forcing PHP_EOL fixes the issue.
>
>The [commit referenced in the blame](https://github.com/php/php-src/commit/6983ae751cd301886c966b84367fc7aaa1273b2d) >changed from "\n" to "\r\n". Current code is this:
>
>	char *line_sep = PG(mail_mixed_lf_and_crlf) ? "\n" : "\r\n";
>
>So depending on the setting, either will be used. The change was introduced in https://github.com/php/php-src/commit/>cc931af35ddc0a07c0a669eb4da9dc66e1f73af4, which seems to have landed in 8.2.4.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -1548,7 +1548,7 @@ class PHPMailer
             static::setLE(self::CRLF);
         } else {
             //Maintain backward compatibility with legacy Linux command line mailers
-            //Force PHP_EOL for all mail() calls on Linux
+            //Force PHP_EOL for all mail() calls on Linux - regardless of PHP versions -
             static::setLE(PHP_EOL);
         }
         //Check for buggy PHP versions that add a header with an incorrect line break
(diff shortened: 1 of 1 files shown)


Synchro:
I'm not sure what's going on here. Two of these commits are just comments, and the other removes a version check that has nothing to do with the platform, and the php-src link you provided is broken.

The source snippet you included mentions `mail_mixed_lf_and_crlf`, which doesn't suggest a platform-specific problem, but more that you're providing content with mixed line breaks, which is a different problem than you describe, and quite reasonably will result in extra breaks.

PR #3184, opened 2025-07-25
Change action_function to callable phpdoc type

brianstoop:
I made this change because we use a a object method array for the action_function which should work with the PHPMailer code.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -563,7 +563,7 @@ class PHPMailer
      *   string  $extra         extra information of possible use
      *                          "smtp_transaction_id' => last smtp transaction id
      *
-     * @var string
+     * @var callable
      */
     public $action_function = '';
 


Synchro:
I'm not entirely sure what the correct type designation should be given that it can just be a `string` containing a global function name. Reading phpstan docs, I see that both `callable-string` and `callable` exist, so I'm wondering if it should be better declared as `callable|callable-string`, and whether that is compatible with it containng an empty string (it's default value).

jrfnl:
For future reference if the property would ever get typed - `callable` is not a valid property type 🤷🏻‍♀️

Synchro:
@jrfnl Do you have a better idea how it should be typed, given that an array-format callable is clearly not a string? Would `string|array` or `string|callable|array` be better?

jrfnl:
> Do you have a better idea how it should be typed, given that an array-format callable is clearly not a string? Would `string|array` or `string|callable|array` be better?

@Synchro No, I haven't got any advise on how to type it. I'd say: use whatever the tooling you use for static analysis supports 😉 
My remark was more to raise awareness of the type not being one natively supported by PHP for properties.

pprkut:
Just as additional piece of information, `$this->action_function` is primarily used as first argument for `call_user_func()`, which is documented with the `callable` type: https://www.php.net/manual/en/function.call-user-func.php.
You could be more specific, but I think at its core that is the intention that should be communicated here.

brianstoop:
This commit has already changed the type to `callable|callable-string` https://github.com/PHPMailer/PHPMailer/commit/f202f351cad2ce8c17f59dceca3b7229280a9e80. So I'll close this PR

PR #3187, opened 2025-08-04
GH Actions: don't run cron jobs on forks

jrfnl:
While workflows are disabled by default in forks, it is quite common for contributors to enable them to verify CI will pass before submitting a pull request.

When enabling workflow runs in forks, it's "all or nothing". This means that:
* All workflows which are only intended to be run on the canonical repo will also be enabled. These workflows will also often need access to repo-specific secrets and will typically fail when run from a fork.
* Workflows which contain cron jobs will also be enabled. Depending on the type of account the contributor has, this can burn through their "CI minutes".

This commit is based on a review of workflows containing cron jobs and disables running the jobs when a cron job is triggered in a fork.

### DIFF
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -12,6 +12,9 @@ permissions: read-all
 
 jobs:
   analysis:
+    # Don't run the cron job on forks.
+    if: ${{ github.event_name != 'schedule' || github.event.repository.fork == false }}
+
     name: Scorecards analysis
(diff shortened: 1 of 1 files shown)


Synchro:
Thanks – I didn't even know this was a problem!

jrfnl:
> Thanks – I didn't even know this was a problem!

It's only a small issue. Disabling cron jobs on forks just makes for a tiny improvement in the contributing experience ;-)

PR #3191, opened 2025-08-10
PHP 8.5 | Tests: prevent deprecation notice for Reflection*::setAccessible()

jrfnl:
Since PHP 8.1, calling the `Reflection*::setAccessible()` methods is no longer necessary as reflected properties/methods/etc will always be accessible. However, the method calls are still needed for PHP < 8.1.

As of PHP 8.5, calling the `Reflection*::setAccessible()` methods is now formally deprecated and will yield a deprecation notice, which will fail test runs. As of PHP 9.0, the `setAccessible()` method(s) will be removed.

With the latter in mind, this commit prevents the deprecation notice by making the calls to `setAccessible()` conditional.

Silencing the deprecation would mean, this would need to be "fixed" again come PHP 9.0, while the current solution should be stable, including for PHP 9.0.

Ref: https://wiki.php.net/rfc/deprecations_php_8_5#extreflection_deprecations

### DIFF
--- a/test/OAuth/OAuthTest.php
+++ b/test/OAuth/OAuthTest.php
@@ -36,7 +36,7 @@ final class OAuthTest extends TestCase
         $PHPMailer = new PHPMailer(true);
         $reflection = new \ReflectionClass($PHPMailer);
         $property = $reflection->getProperty('oauth');
-        $property->setAccessible(true);
+        (\PHP_VERSION_ID < 80100) && $property->setAccessible(true);
         $property->setValue($PHPMailer, true);
         self::assertTrue($PHPMailer->getOAuth(), 'Initial value of oauth property is not true');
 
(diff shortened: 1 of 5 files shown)


jrfnl:
FYI: the deprecation notices coming from the PHPUnit Polyfills should be fixed now if the build would be retriggered. I've released new versions fixing those.

test/OAuth/OAuthTest.php
```diff
@@ -36,7 +36,7 @@ public function testOAuth()
         $PHPMailer = new PHPMailer(true);
         $reflection = new \ReflectionClass($PHPMailer);
         $property = $reflection->getProperty('oauth');
-        $property->setAccessible(true);
+        (\PHP_VERSION_ID < 80100) && $property->setAccessible(true);
```
Synchro, line 39: I've never seen this syntax before — you can just have expressions lying around like this?

jrfnl: Of course, the verbose way would be to:
```php
if (\PHP_VERSION_ID < 80100) {
    $property->setAccessible(true);
}
```

... but this felt (to me) like a simple and elegant enough version of the same. Let me know if you want it differently.

jrfnl:
@Synchro Could you please retrigger the build ? The PHP 8.5 issues with Composer should be fixed now Composer 2.8.11 has been released.

Issue #3183, opened 2025-07-25
stream_select retry not working for other language (locale) than English


christianseel:
# Problem description
This is highly related to https://github.com/PHPMailer/PHPMailer/pull/2074

PHPMailer version 6.10.0 with PHP 8.4

Our own logging revealed several different errors when attempting to send emails with PHPMailer and a remote SMTP server (port 465, tls):
- SMTP Error: The following recipients failed: *******:  SMTP server error: QUIT command failed Detail: Accepted
- SMTP Error: data not accepted
- The following From address failed: ***** : MAIL FROM command failed,,0, 

Using the PHPMailer debug log, we traced the error back to the same initial error for all cases:

```
SMTP Debug: Connection failed. Error #2: stream_select(): Unable to select [4]: Unterbrechung während des Betriebssystemaufrufs (max_fd=17) [/.../vendor/phpmailer/phpmailer/src/SMTP.php line 1299]
SMTP Debug: SMTP -> get_lines(): select failed (stream_select(): Unable to select [4]: Unterbrechung während des Betriebssystemaufrufs (max_fd=17))
````

`stream_select` fails with an `Interrupted system call` (translated from German `Unterbrechung während des Betriebssystemaufrufs`) warning from PHP.

We found PR https://github.com/PHPMailer/PHPMailer/pull/2074 that is supposed to handle this (legitimate?) signal. However our PHP application has `setlocale('de_DE')` set, causing PHP warnings to be written in German (`Unterbrechung während des Betriebssystemaufrufs`).

[The code that is supposed to handle this case](https://github.com/PHPMailer/PHPMailer/blob/master/src/SMTP.php#L1343) only handles English warning/error messages. In our case some emails failed because of this error.

**I want to make people aware of the issue and ask if you see any way to adjust PHPMailer itself to handle different languages.** Our workaround might be a potential solution?

# Workaround
We implemented a workaround by temporarily setting the locale to English and reverting to the original locale after the email was sent.

```
// with "0", the locale setting is not affected, only the current setting is returned.
$previousLocale = setlocale(LC_MESSAGES, '0');
// Enfore english messages from PHP
setlocale(LC_MESSAGES, 'en_US.UTF-8');
try {
    // Send email
    $mailer->send();
} catch (\Exception $exception) {
    // e.g. log exception
} finally {
    // Reset previous locale
    setlocale(LC_MESSAGES, $previousLocale);
}
```

# Debug output
```
...
[...]

PR #3193, opened 2025-08-11
Enhance 'interrupted system call' check to handle non-english locale

christianseel:
This is a fix for https://github.com/PHPMailer/PHPMailer/issues/3183 – stream_select retry not working for other language (locale) than English

I received a great hint by @teefax – who pointed out that the `stream_select(): Unable to select` part of the php warning message is not translated for other locales. Only the `interrupted system call` part is translated.

That `Unable to select` is followed by an error number in square brackets. However that number can be different based on the operating system of the server. But PHP has a constant [SOCKET_EINTR](https://www.php.net/manual/en/sockets.constants.php#constant.socket-eintr) for that.

So the recommendation is to check for the SOCKET_EINTR constant which is defined under Windows and UNIX-like platforms (if available on the platform) and use that with the other english warning text to catch those interrupted system calls for non-english locale applications.

I left the existing check to avoid any potential unknown breaking change and extended the if-condition.

I tested this with our application and it worked as expected (without the setlocale workaround shared in https://github.com/PHPMailer/PHPMailer/issues/3183).

### DIFF
--- a/src/SMTP.php
+++ b/src/SMTP.php
@@ -1340,7 +1340,16 @@ class SMTP
 
                 //stream_select returns false when the `select` system call is interrupted
                 //by an incoming signal, try the select again
-                if (stripos($message, 'interrupted system call') !== false) {
+                if (
+                    stripos($message, 'interrupted system call') !== false ||
+                    (
+                        // on applications with a different locale than english, the message above is not found because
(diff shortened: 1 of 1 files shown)


Synchro:
Thanks — Nice detective work there!

Issue #3194, opened 2025-08-11
Tests: Missing Translation Strings


SirLouen:
What is the idea of having a unit test that checks for missing translation string?

Should this be just informative instead of a breaking test?

<details>

<summary>Full log</summary>

```bash
1) PHPMailer\Test\Language\TranslationCompletenessTest::testTranslations
Missing translations in id: buggy_php, invalid_header, smtp_code, smtp_code_ex, smtp_detail, no_smtputf8
Missing translations in et: buggy_php, invalid_header, invalid_hostentry, invalid_host, smtp_code, smtp_code_ex, smtp_detail, no_smtputf8
Missing translations in tl: buggy_php, invalid_header, invalid_hostentry, invalid_host, smtp_code, smtp_code_ex, smtp_detail, no_smtputf8
Missing translations in ca: buggy_php, extension_missing, invalid_header, invalid_hostentry, invalid_host, smtp_code, smtp_code_ex, smtp_detail, no_smtputf8
Missing translations in uk: buggy_php, invalid_header, invalid_hostentry, invalid_host, smtp_code, smtp_code_ex, smtp_detail, no_smtputf8
Missing translations in pl: no_smtputf8
Missing translations in hu: buggy_php, invalid_header, invalid_hostentry, invalid_host, smtp_code, smtp_code_ex, smtp_detail, no_smtputf8
Missing translations in sr: buggy_php, invalid_header, invalid_hostentry, invalid_host, smtp_code, smtp_code_ex, smtp_detail, no_smtputf8
Missing translations in lt: buggy_php, extension_missing, invalid_header, invalid_hostentry, invalid_host, smtp_code, smtp_code_ex, smtp_detail, no_smtputf8
Missing translations in da: no_smtputf8
Missing translations in ms: buggy_php, invalid_header, invalid_hostentry, invalid_host, smtp_code, smtp_code_ex, smtp_detail, no_smtputf8
Missing translations in nb: no_smtputf8
Missing translations in hy: buggy_php, invalid_header, invalid_hostentry, invalid_host, smtp_code, smtp_code_ex, smtp_detail, no_smtputf8
Missing translations in pt: no_smtputf8
Missing translations in hi: no_smtputf8
Missing translations in fr: no_smtputf8
Missing translations in sl: no_smtputf8
[...]

PR #3195, opened 2025-08-11
Making Translation Strings Test Informative

SirLouen:
Regarding this issue
https://github.com/PHPMailer/PHPMailer/issues/3194

With this patch, we make this test informative instead of mandatory to pass. There are a ton of translation strings missing, and given the rare languages available like Latin, I'm uncertain if they will be ever fully sorted out (unless using some automatic translator)

### DIFF
--- a/test/Language/TranslationCompletenessTest.php
+++ b/test/Language/TranslationCompletenessTest.php
@@ -91,13 +91,13 @@ final class TranslationCompletenessTest extends TestCase
             }
         }
 
-        // Make this test informational only: skip with details instead of failing.
+        // Make this test informational only.
         if ($err !== '') {
             $this->markTestSkipped("Translation completeness (informational):\n" . $err);
             return;
         }
 
-        // No differences found: count a dummy assertion so the test is not marked as risky.
+        // Avoid marking test as risky.
         $this->addToAssertionCount(1);
     }
 }
--- a/test/Language/TranslationCompletenessTest.php
+++ b/test/Language/TranslationCompletenessTest.php
@@ -91,7 +91,13 @@ final class TranslationCompletenessTest extends TestCase
             }
         }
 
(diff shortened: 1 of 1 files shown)


test/Language/TranslationCompletenessTest.php
```diff
@@ -91,7 +91,13 @@ public function testTranslations()
...
-        // If we have no extra and no missing translations, $err will be empty.
-        self::assertEmpty($err, $err);
+        // Make this test informational only.
+        if ($err !== '') {
+            $this->markTestSkipped("Translation completeness (informational):\n" . $err);
+            return;
+        }
+
+        // Avoid marking test as risky.
+        $this->addToAssertionCount(1);
```
jrfnl, line 101: Looks like you are using a PHPUnit internal method here.

Why not just mark the test with the `@doesNotPerformAssertions` annotation ?

test/Language/TranslationCompletenessTest.php
```diff
@@ -91,7 +91,13 @@ public function testTranslations()
             }
         }
 
-        // If we have no extra and no missing translations, $err will be empty.
-        self::assertEmpty($err, $err);
+        // Make this test informational only.
+        if ($err !== '') {
+            $this->markTestSkipped("Translation completeness (informational):\n" . $err);
```
jrfnl, line 96: Is `markTestSkipped()` the right method to use ? Would `markTestIncomplete()` maybe be appropriate ?

jrfnl (commented):
To be honest, I don't see the point of this change.

As per [my comment in #3194:](https://github.com/PHPMailer/PHPMailer/issues/3194#issuecomment-3176955998)

> The test is excluded by default, so I don't see what the problem is ?
> 
> It's a helper for maintainers, not a test which is run as part of the test suite.

SirLouen:
Solved no need for changes

Issue #3196, opened 2025-08-11
Tests: Targeting No-Mbstring Requirement Tests


SirLouen:
There are a couple of tests:

`testAddressSplittingNativeNoMbstring` 
`testAddressSplittingImapNoMbstring`
`testAddReplyToFailsOn8BitCharInDomainWithoutOptionalExtensions`
`testSetFromFailsOn8BitCharInDomainWithoutOptionalExtensions`

I was thinking if we can specifically target them without having to list them manually. Something like adding all them to a group. This can be useful for things like excluding them without having to targeting them manually. 

Also commented here
https://github.com/PHPMailer/PHPMailer/discussions/3192

PR #3197, opened 2025-08-11
Grouping No MbString Tests

SirLouen:
Following this issue
https://github.com/PHPMailer/PHPMailer/issues/3196

The idea here is to group these 4 tests for purposes as explained in the issue.

It could be very useful to target them all at once.

### DIFF
--- a/test/PHPMailer/PHPMailerTest.php
+++ b/test/PHPMailer/PHPMailerTest.php
@@ -1346,6 +1346,8 @@ EOT;
     /**
      * @requires extension mbstring
      * @requires function idn_to_ascii
+     * 
+     * @group mbstringExtRequired
      */
     public function testGivenIdnAddress_addAddress_returns_true()
     {
--- a/test/PHPMailer/PHPMailerTest.php
+++ b/test/PHPMailer/PHPMailerTest.php
@@ -278,6 +278,9 @@ EOT;
 
     /**
      * Send a message containing ISO-8859-1 text.
+     * 
+     * @requires extension mbstring
+     * @group mbstringExtRequired
      */
     public function testHtmlIso8859()
     {
--- a/test/PHPMailer/PHPMailerTest.php
+++ b/test/PHPMailer/PHPMailerTest.php
@@ -1117,6 +1117,8 @@ EOT;
 
     /**
      * Tests CharSet and Unicode -> ASCII conversions for addresses with IDN.
+     * 
+     * @group mbstringExtRequired
      */
     public function testConvertEncoding()
     {
@@ -1162,6 +1164,8 @@ EOT;
 
     /**
      * Tests removal of duplicate recipients and reply-tos.
+     * 
+     * @group mbstringExtRequired
      */
     public function testDuplicateIDNRemoved()
     {
@@ -1230,6 +1234,8 @@ EOT;
 
     /**
      * Test SMTPUTF8 usage, including when it is not to be used.
+     * 
+     * @group mbstringExtRequired
      */
     public function testSmtpUTF8()
     {
--- a/test/PHPMailer/ParseAddressesTest.php
+++ b/test/PHPMailer/ParseAddressesTest.php
@@ -33,6 +33,7 @@ final class ParseAddressesTest extends TestCase
      * with the Mbstring extension available.
      *
      * @requires extension mbstring
+     * @group mbstringExtRequired
      *
      * @dataProvider dataAddressSplitting
      *
@@ -64,6 +65,7 @@ final class ParseAddressesTest extends TestCase
      *
      * @requires extension imap
      * @requires extension mbstring
+     * @group mbstringExtRequired
      *
      * @dataProvider dataAddressSplitting
      *
@@ -93,7 +95,7 @@ final class ParseAddressesTest extends TestCase
      * Test RFC822 address splitting using the PHPMailer native implementation
      * without the Mbstring extension.
      *
-     * @group nombstring
+     * @group mbstringExtDisabled
      * @dataProvider dataAddressSplitting
      *
      * @param string $addrstr  The address list string.
@@ -128,7 +130,7 @@ final class ParseAddressesTest extends TestCase
      *
      * @requires extension imap
      *
-     * @group nombstring
+     * @group mbstringExtDisabled
      * @dataProvider dataAddressSplitting
      *
      * @param string $addrstr  The address list string.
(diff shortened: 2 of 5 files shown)


jrfnl (commented):
Looks like a valid improvement. I do wonder if we could come up with a more readable group name. Maybe something like `mbstringExtensionUnavailable`, though that may be a bit long ?

Also the tests in the `ReplyToGetSetClearTest` and the `SetFromTest` classes require either the Mbstring extension *or* the Intl extension, so maybe those should get their own group name ?

PHPUnit allows for running multiple groups in one go, so that should still make running these tests more straight forward.

SirLouen:
> Looks like a valid improvement. I do wonder if we could come up with a more readable group name. Maybe something like mbstringExtensionUnavailable, though that may be a bit long ?

Maybe I can switch this to `mbstringExtDisabled` to keep it short? 

I'm building a fully containerized testing environment like `wp-env` and the current pain point I found was actually targeting those tests that don't need mbstring (and probably in the future, they will be removed once PHPUnit 8 and 9 get EOL and phpmailer needs to upgrade). 

> Also the tests in the ReplyToGetSetClearTest and the SetFromTest classes require either the Mbstring extension or the Intl extension, so maybe those should get their own group name ?

The thing is that mbstring has become mostly a de facto now. It took me a while to actually find an image without mbstring by default (and that handled PHP 7.4 packages and ended with a somewhat older alpine container). 

Maybe we could also be grouping those who actually need `mbstring` in case we want to exclude them for the specific version of testing without mbstring (while still running the rest of the suite), but perhaps we should treat this independently instead of all in one go, useful to control support for older PHP versions.

SirLouen:
@jrfnl I've been thinking on your proposal and to make a fully operative no-mbstring suite, we need to tag all the tests that are requiring mbstring to exclude them selectively

I've added two groups `mbstringExtRequired` and  `mbstringExtDisabled`.

SirLouen:
Let me add 3 extra ones and a little thing I've noted before proceeding

SirLouen:
I'm going to add one extra test  `testHtmlIso8859` and also take advantage of this PR to add, apart from the group, the `@requires extension mbstring` as it is using both `mb_convert_encoding` and `mb_detect_encoding`

jrfnl:
> ... to make a fully operative no-mbstring suite, we need to tag all the tests that are requiring mbstring to exclude them selectively
> 
> I've added two groups `mbstringExtRequired` and `mbstringExtDisabled`.

I'm not sure I understand what you are trying to do (at all). The `@requires` annotation already takes care of handling when the mbstring extension is required. It just can't be used for when the test requirement is for the extension to **_not_** be available.

So adding `@group` tags for the tests which require mbstring to be disabled makes some sort of sense, but anything else doesn't.

SirLouen:
@jrfnl the idea is to group them to be able to exclude them in certain testing protocols to provide a cleaner testing report that focus on tests that are actually important for that specific protocol (in this case, the protocol that excludes mbstring)
Like in: https://github.com/SirLouen/phpmailer-dev/blob/949b55e5bf9b8c2ed12d8f10b4f8450146f683d3/.lando.yml#L92

For now I've identified 4 protocols:
1. The regular one
2. The one without mbstring
3. The one without OpenSSL
4. The one with Qmail as the entrypoint MTA.

jrfnl:
> @jrfnl the idea is to group them to be able to exclude them in certain testing protocols to provide a cleaner testing report that focus on tests that are actually important for that specific protocol (in this case, the protocol that excludes mbstring) Like in: https://github.com/SirLouen/phpmailer-dev/blob/949b55e5bf9b8c2ed12d8f10b4f8450146f683d3/.lando.yml#L92

Well, I still don't really understand. The tests for this package need to be set up in a way to make it straight forward to contribute to this package.

If you want to run the tests in a different project environment, that's not something this project should need to be concerned with and is something you should largely be able to configure with a custom PHPUnit configuration file to run those tests you want to run.

I don't understand why you want to make changes to the test suite for **this** package just to accommodate some test set up in a completely unrelated package.

SirLouen:
> I don't understand why you want to make changes to the test suite for **this** package just to accommodate some test set up in a completely unrelated package.

@jrfnl long story short: there are a good amount of `no-mbstring` tests not passing for a reason. 
image
(And there are 2 more that are will pop as soon as some of these are sorted, so I count, minimum, 7 in total)

And one of the reasons is that testing PHPMailer is not being done for all environments possible (the ones I mentioned above). Checking the GHA, this is only focusing on the main testing group. For the rest, as-is now, it's a serious pain because there are no build tools fully backing this. I already [commented on this](https://github.com/PHPMailer/PHPMailer/discussions/3175) here because I was looking for more feedback, and after read the proposal is when [I got to this conclusion](https://github.com/PHPMailer/PHPMailer/discussions/3198).

I'm trying to lend a hand on this. And this proposal is just part of the path to get there. This "package" as you call it, is being made ad-hoc for this purpose and for this project. Maybe when I polish it, we could consider fully integrating this with the project. It will simplify the life of testing massively for any external contributor (as I told you, it's inspired by WP `env` but taking `lando` because it simplifies all the tooling and scripting that is being done manually in JS in `wp-env`). 

I've been researching through PHPUnit documentation, and I could not find a way to individually exclude certain tests. In this case, when I'm running an environment with no `mbstring` extension loaded, for a clean output, without skipped tests, it's needed to target specifically these tests, and the only way to do this is by grouping. I don't really see how this can negatively impact this project, but I think it could positively improve it once the tooling is cleanly working. 

If you have some other idea to do a workaround for for this, I would like to hear it.

Synchro:
I have a hankering for a PR to PHPUnit that adds support for `@notrequires` or `@omits` tags...

Synchro:
My idea for adding `notrequires` tags to PHPUnit was shot down, but there was [a constructive suggestion](https://github.com/sebastianbergmann/phpunit/issues/6318#issuecomment-3188388338) to use `Assert::markTestSkipped` to implement it in the test instead.

kmk142789 (approved):

Youngfarmer95 (approved):

Synchro:
Is this still relevant?

SirLouen:
I'm still holding a local branch for this. For full env testing it's necessary, or you are just missing on scenarios by skipping them unknowingly

Unfortunatelly I'm a bit less active because I dropped a lot of my activity in WP teams lately. I think a bit of grouping is needed on this suite, although I leave it for your call.

thibaut-tbc-bujiriri:
can i work on this issues ?

Synchro:
Be my guest!

Issue #3199, opened 2025-08-12
Tests: Message is not precise for mbstring/intl test requirements


SirLouen:
The function `idnSupported` requires PHP extension `intl` and `mbstring` to return `true`

```php
function_exists('idn_to_ascii') && function_exists('mb_convert_encoding');
```

3 tests are returning the following message:
```
intl and/or mbstring extensions are not available`
```

Which is a little misleading because both extensions are required, not one or the other.

I'm sending a little PR to improve this copy.

Synchro:
Fixed by #3200

PR #3200, opened 2025-08-12
Improving Copy for mbstring intl test requirements

SirLouen:
Adding this PR regarding #3199 
All the info is there.


SirLouen:
@Synchro I think you can merge this easily. It was a mistake in the copy. When I tried to run these tests with one or another of the extension it failed. They need both extensions at the same time.

Issue #3207, opened 2025-08-17
Warning for the doCallback in SingleTo


SirLouen:
`Warning: Undefined array key "address" in /app/PHPMailer/src/PHPMailer.php on line 2001 Warning: Undefined array key "name" in /app/PHPMailer/src/PHPMailer.php on line 2001 Warning: Undefined array key "address" in /app/PHPMailer/src/PHPMailer.php on line 2001 Warning: Undefined array key "name" in /app/PHPMailer/src/PHPMailer.php on line 2001`

How to Trigger this?
Following this instructions: https://github.com/PHPMailer/PHPMailer/issues/3204#issuecomment-3194305707

The problem here is that `parseAddresses` return an array of associative arrays, and the two doCallback are getting the values as if `$addrinfo` was a simple associative array

Issue #3204, opened 2025-08-16
ISO-8859-1 decoding is not working as expected


SirLouen:
I'm reviewing the tests for `ParseAddressesTest.php` and I have a hunch that they have been made ad-hoc picking up the results (assuming that the results are correct, not the other way around). Maybe I'm wrong but I need some extra review because after a good amount of time testing a couple of combinations, I can't happen to receive consistent results.

Here is an example:

https://github.com/PHPMailer/PHPMailer/blob/800919c6a41d3bc4e87406725531ebe06b5546b4/test/PHPMailer/ParseAddressesTest.php#L210-L218

Lets put the string in just one line
`=?iso-8859-1?B?QWJjZGVmZ2ggSWprbG3DsSDmnIPorbDlrqTpoJDntITns7vntbE=?=`

```php
$encoded = '=?iso-8859-1?B?QWJjZGVmZ2ggSWprbG3DsSDmnIPorbDlrqTpoJDntITns7vntbE=?=';
$decoded = mb_decode_mimeheader($encoded);
echo $decoded;
```

Which results in this: 
```
Abcdefgh IjklmÃ± æè­°å®¤é ´
                           ç³»çµ
```

But if we change the string to:
`=?utf-8?B?QWJjZGVmZ2ggSWprbG3DsSDmnIPorbDlrqTpoJDntITns7vntbE=?=`

The result is: `Abcdefgh Ijklmñ 會議室預約系統` as in the test. 

Basically the code is forcing UTF-8 conversion, which looks fine, because in this example, this is wrongly labeled as UTF-8 encoding, and the system is taking care of this wrong label. Plus, this is not a valid RFC2047 string... which should be handled but definitely not parsed.

But this is introducing a bigger problem. What happens if we actually use a correctly labeled encoding for the right string like:

`=?ISO-8859-1?Q?Willkommen_in_unserem_Caf=E9!?=`

Using the same code above, the result is `Willkommen in unserem Café!`

However, the current code is resulting in `Willkommen in unserem Caf�`

Because, like in the previous example, it thinks it was wrongly labeled and tries to use UTF-8, but in this case it's not. 

Generally, no one sends this encoding for `latin-1` but the text directly in the header without any encoding at all, because latin-1 is MIME default, but this is effectively an edge case that I believe should be considered to be RFC 2047 compliant.

The reality is that, in the absence of an explicit encoding like in the first example, it should not be handled correctly.

It appears that it happens the same for:

[...]

PR #3208, opened 2025-08-17
Sorting the Array problem in doCallback

SirLouen:
Follow-up to this issue https://github.com/PHPMailer/PHPMailer/issues/3207 I found while digging into https://github.com/PHPMailer/PHPMailer/issues/3204 (they are unrelated, but it popped out of the blue)

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -1851,7 +1851,7 @@ class PHPMailer
                         $this->Subject,
                         $body,
                         $this->From,
-                        []  
+                        []
                     );
                 }
                 $this->edebug("Result: " . ($result === 0 ? 'true' : 'false'));
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -1842,16 +1842,18 @@ class PHPMailer
                 fwrite($mail, $body);
                 $result = pclose($mail);
                 $addrinfo = static::parseAddresses($toAddr, true, $this->CharSet);
-                $this->doCallback(
-                    ($result === 0),
-                    [[$addrinfo['address'], $addrinfo['name']]],
-                    $this->cc,
-                    $this->bcc,
-                    $this->Subject,
-                    $body,
-                    $this->From,
-                    []
-                );
+                foreach ($addrinfo as $addr) {
+                    $this->doCallback(
+                        ($result === 0),
+                        [[$addr['address'], $addr['name']]],
+                        $this->cc,
+                        $this->bcc,
+                        $this->Subject,
+                        $body,
+                        $this->From,
+                        []  
+                    );
+                }
                 $this->edebug("Result: " . ($result === 0 ? 'true' : 'false'));
                 if (0 !== $result) {
                     throw new Exception($this->lang('execute') . $this->Sendmail, self::STOP_CRITICAL);
@@ -2017,16 +2019,18 @@ class PHPMailer
             foreach ($toArr as $toAddr) {
                 $result = $this->mailPassthru($toAddr, $this->Subject, $body, $header, $params);
                 $addrinfo = static::parseAddresses($toAddr, true, $this->CharSet);
-                $this->doCallback(
-                    $result,
-                    [[$addrinfo['address'], $addrinfo['name']]],
-                    $this->cc,
-                    $this->bcc,
-                    $this->Subject,
-                    $body,
-                    $this->From,
-                    []
-                );
+                foreach ($addrinfo as $addr) {
+                    $this->doCallback(
+                        $result,
+                        [[$addr['address'], $addr['name']]],
+                        $this->cc,
+                        $this->bcc,
+                        $this->Subject,
+                        $body,
+                        $this->From,
+                        []
+                    );
+                }
             }
         } else {
             $result = $this->mailPassthru($to, $this->Subject, $body, $header, $params);


jrfnl:
This PR is missing a test proving the bug and safeguarding the fix.

SirLouen:
@jrfnl I have not added tests for this for two reasons:
- There were no tests previously for this
- This is a deprecated section (based on a deprecated conditional). 

I only used this for a demo to trigger another errors and happened to find this wrong structure by chance. This is just a micro fix to sort it, safeguarding some deprecated, that theoretically should not be modified further, is just going to increase the testing processing output.

Synchro:
While it would be nice to have tests, for this, it is indeed a section of code that is effectively deprecated, and clearly it's not being used as nobody has reported this straightforward bug!

jrfnl:
The other side of that coin is that with tests, this bug might have been found way earlier and that deprecated code should still be functional (and thus covered by tests) until it is actually removed.... 🤷🏻‍♀️

SirLouen:
> The other side of that coin is that with tests, this bug might have been found way earlier and that deprecated code should still be functional (and thus covered by tests) until it is actually removed.... 🤷🏻‍♀️

I will go further and say that this is not a code testing issue but a very simple code quality problem that could be solved by implementing something like Psalm or PHPStan. Yes, in this case, since it's a structure issue with anything (namely, a unit test) that simply called this section of code, it would have been fixed.

But on the other side, I've already found the same, but the other way around: [unit tests actually obfuscating errors in code](https://github.com/PHPMailer/PHPMailer/issues/3210). The problem with adding unit tests for existing code just for the sake of having tests all over the place is that, generally, we simply adapt the UT output to the results, assuming that the code is correct. But many times, it's not. So this is not always true, and for this case, given that it's deprecated and code should not be further modified already been manually tested (by me), it should be simply forgotten and maybe removed one day. Adding unit tests only increases the testing load (not terrible but not great either). 

But even more importantly, it is too much of an effort to analyze now the test cases for this. Unless you are genuinely interested in this particular deprecated thing, for some reason, we are going to repeat the wrong practice, as in my previous example: adding tests that perfectly fit the current output instead of of adding tests that actually make sense according to an expected output that has been previously broken down (a time waster doing this for a deprecated section, ngl, if I had no choice, I would rather remove it).

jrfnl:
@SirLouen I take offense by your remarks.

As the person who wrote a significant number of the tests, I know for a fact they were not written to "document the current output". I found a significant number of bugs while writing tests for PHPMailer a few years back and fixed all of those too.

SirLouen:
> @SirLouen I take offense by your remarks.
> 
> As the person who wrote a significant number of the tests, I know for a fact they were not written to "document the current output". I found a significant number of bugs while writing tests for PHPMailer a few years back and fixed all of those too.

@jrfnl no offense intended. You are one of my main WP references, and I admire your epic job on testing and linting.

But what I'm trying to explain here, is that sometimes, unit tests backfire. Mainly for those that are not created by the same person who actually wrote the code, because it's super difficult to imagine all the cases in many scenarios, and we end sometimes adjusting end tests to the output, which can be dangerous, because they will bring a false sense of certainty. So, in some cases, were code was not too well documented, its better to leave tests undone until someone happens to take the opportunity to re-review that portion of code and do the tests consequently. This is the justification of why I preferred to leave these tests undone.

jrfnl:
> But what I'm trying to explain here, is that sometimes, unit tests backfire. Mainly for those that are not created by the same person who actually wrote the code, because it's super difficult to imagine all the cases in many scenarios, and we end sometimes adjusting end tests to the output, which can be dangerous, because they will bring a false sense of certainty. So, in some cases, were code was not too well documented, its better to leave tests undone until someone happens to take the opportunity to re-review that portion of code and do the tests consequently. This is the justification of why I preferred to leave these tests undone.

I understand your point, but also think you are wrong. Let me explain.

> Mainly for those that are not created by the same person who actually wrote the code, because it's super difficult to imagine all the cases in many scenarios.

It doesn't matter whether someone is the original author of the code writing the tests alongside the code or whether the tests are written afterwards, it's **ALWAYS** super difficult to imagine all possible scenarios.

Quite often, the original author will have started out writing the code to solve a certain problem and will have blind spots for other use cases they didn't think of, for which the code will also be used.

Quite often, the original author will have worked on the code for a significant block of time and will have stared at the code so much they just don't see what's wrong with it anymore.

Quite often, the original author will use test driven development and the resulting code may then be fully covered by tests, but still not actually function correctly (there's no code related to the tests they _didn't write_, so test coverage will show 100%, even though the code is buggy).

While the same applies if someone writes tests after the fact, if the original code was written a while ago, the "test author" has the benefit of hindsight by seeing which bugs were reported and fixed since the original code entered the code base (helped along by reproduction scenarios included in the bug reports), so at least has some clues for tests which should definitely be added.

That does, of course, presume such a "test author" uses git history and git blame to try and understand why the code is as it is today...

> ... and we end sometimes adjusting end tests to the output, which can be dangerous, because they will bring a false sense of certainty

While that is certainly true in some cases, though less often if tests are written by people who have plenty of experience, it isn't always a bad thing and may even be done deliberately.

Documenting the current behaviour - even if not fully correct - can be helpful to demonstrate why a bug fix - or a breaking change - is needed later. And especially in the case of breaking changes, having the test expectations change for those is part of documenting the breaking change.

Generally speaking, I'd say that in that case, it is a good idea to also open an issue to discuss whether the observed behaviour is correct and how it should potentially be changed in the future (alongside the PR to add the tests documenting the current behaviour).

> I will go further and say that this is not a code testing issue but a very simple code quality problem that could be solved by implementing something like Psalm or PHPStan.

In my experience, tools like Psalm and PHPStan have definite value, but should not be overestimated and are most definitely no replacement for having good tests.

More than anything, I too often see people blindly adding type casts throughout a code base "because PHPStan flagged something", while in > 90% of the cases, that's not the correct way to fix whatever PHPStan flagged. More often, the real fix lies in fixing an incorrect type annotation elsewhere in the codebase, or even in a dependency, or in adding more defensive coding.

The problem with that is, that these types casts - just like type declarations in a non-strict codebase - **_hide_** bugs, instead of  solving them.

And that's aside from the fact that I still see way too many false positives coming from PHPStan/Psalm and that less experienced devs take these as truth instead of realizing those are false positives.

> Adding unit tests only increases the testing load (not terrible but not great either).

That should **NEVER** be a consideration against adding tests. With the CI landscape as it currently is, running tests costs nothing but time for open source projects.
And I'd rather wait 1 minute extra for CI to finish, than release something with a preventable bug.

> So this is not always true, and for this case, given that it's deprecated and code should not be further modified already been manually tested (by me), it should be simply forgotten and maybe removed one day.

When I suggested to add tests, I didn't say "add tests covering every possible scenario". I asked for a test **_specifically to cover this bug_** to safeguard that it won't be accidentally re-introduced. That test can even be annotated with a `@coversNothing` annotation and an `{@internal Only testing bug xyz, not the full method}` note to highlight that the rest of the code involved is not tested.

The current test suite is incomplete and not adding tests for a specific situation which is clearly testable (which it is as you manually tested it), is not helping in making it more comprehensive.

All in all, this discussion also highlights why it is better for a codebase to have tests for the different levels of testing - unit, integration, end-to-end - and why the focus should not just be on unit tests (which it too often is).

Synchro:
> Quite often, the original author will have started out writing the code to solve a certain problem and will have blind spots for other use cases they didn't think of, for which the code will also be used.

That's exactly what happened with `parseAddresses`. I had a few people that asked for it, though I didn't really want to add it, hence its implementation as a static, standalone function, but one that then ended up being used by SingleTo, which I've never been keen on either, net result was the awkward situation we find ourselves in. I would love to be able to drop a bunch of ugly features in PHPMailer, but BC is a killer.

SirLouen:
I think we are approaching this like the two sides of the same coin. On my side, I'm stating an example where unit tests don't do the cut, but curiously a static analysis tool like the one commented does. And on your example it happens exactly the opposite. This exactly is what makes them great: that they should coexist.

In this particular case, I was not really "testing" the `doCallback` function, but I was just willing to trigger the function within (to test `parseAddresses` btw), just to prove myself something, and it happened to flash an unexpected warning. A warning that a static analysis would have caught in a glimpse, because it is pretty obvious for a tool of this type. 

Surely, if the creator have added a unit test, this would have been avoided, so I also advocate that sending unit tests by default should be like a requirement (more in a full PHP library, where every single line can be easily unit tested). 

But things happen, and as Synchro commented, maybe he wrote this while he was not very happy on supporting it (just for the sake of satisfying the crowd on a side-quest) and maybe it was released fast and furiously (I was reviewing the blames and they are a little confusing on the event line). 

The same applies now with this PR: I don't really want to take much time on this deprecated useless thing, but I just noticed this warning that was kind of obnoxious, made quick fix, no unit tests, no intention to support this. Probably if Mr. Stan was running in the background back in the day, Synchro would have fixed this warning on the fly. 

Now that I'm thinking, maybe it's a good opportunity to give a second thought to introducing Stan or Psalm.

Synchro:
I've mostly given up on psalm; phpstan is where it's at, and it runs in my IDE anyway even if it's not configured in the library itself. rector is quite cool too, but it can't be trusted to make non-breaking changes.

PR #3211, opened 2025-08-22
Fix linting issue on PHP 8.5

jrfnl:
As per the options described in https://github.com/PHPMailer/PHPMailer/pull/3202#issuecomment-3212478928.

Note: the linting ignore comment triggers some PHPCS errors (_sigh_), so I'm selectively excluding those.
Alternatively, it could be considered to exclude test fixture files completely from the PHPCS scan.


jrfnl:
Note: the test failures on PHP 8.5 should be fixed via PR #3191

Synchro:
Thank you!

Issue #3215, opened 2025-08-23
Staticfying the Language Pack


SirLouen:
When going through #3210 one of the proposed solutions is to add a Warning to inform about a fact.

The problem here is that we are dealing with a problem inside the realm of static functions, which doesn't have direct access to the language pack, which belongs to the PHPMailer class.

I believe that, ideally, the internationalization functions should be extrinsic to the PHPMailer class (and the PHPMailer as a whole). But given how much work this is, I would recommend doing this in two steps:

In a first step and to get out of this scenario for now, and be able to achieve the desired functionality, I would say that "staticfying" the language methods, so now we could have access to the `lang` strings both from dynamic and static methods.

In a second step in the future, maybe extracting the whole lang pack into a separate class would be ideal, but this could be extra work to be done, and we need to get to a resolution with #3210 and #3202 (that are chaining each other), so this could be a good first step.

PR #3216, opened 2025-08-23
Staticfying the Language Pack

SirLouen:
Following here the ticket #3215


Synchro:
Thanks, this is a good idea. Language functions don't need access to $this.

PR #3218, opened 2025-08-25
Fix: Encoding fallback when mbstring extension is missing

prab18hat:
This PR adds an iconv fallback to hasMultiBytes and base64EncodeWrapMB when the mbstring extension is not available.
Includes a new test NoMbstringTest.php to verify correct behavior.
Ensures PHPMailer works gracefully in environments without mbstring

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -3693,9 +3693,14 @@ class PHPMailer
         if (function_exists('mb_strlen')) {
             return strlen($str) > mb_strlen($str, $this->CharSet);
         }
+        
+        // Fallback to iconv if available
+        if (function_exists('iconv_strlen')) {
+            return strlen($str) > iconv_strlen($str, $this->CharSet);
+        }
 
-        //Assume no multibytes (we can't handle without mbstring functions anyway)
-        return false;
+        // If neither mbstring nor iconv is available, make a basic check for non-ASCII characters
+        return preg_match('/[^\x00-\x7F]/', $str) === 1;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Synchro:
I'm interested to know what @SirLouen thinks of this. I feel that this is heading in the direction of having too many fallbacks. I would prefer not to support so many combinations of enabled/disabled extensions – I foresee a future version of PHPMailer that *requires* `mbstring`, rather than heading in the direction of ever more convoluted fallbacks. It's just too messy, as we are seeing in other recent PRs.

SirLouen:
Yes, @Synchro, this PR is massively overlapping with #3202 and like we commented on the RFC822 address splitting topic, going for RFC2047 grammars is the same pain. I truly believe now that if anyone wants these “advanced” features, they should be forced to install the corresponding extensions. I don't see any value in reinventing the wheel a million times. 

We could further discuss things in #3202, like the possibility of leaving just mbstring or not, etc... to avoid splitting context so I would close this and keep on the details on the other PR (but first we need to fix #3217 before moving forward, because it's dependent).

Synchro:
Sorry @prab18hat, maybe next time!

prab18hat:
> Sorry @prab18hat, maybe next time!

Thanks @Synchro and @SirLouen for the review 🙏
I understand your concerns about multiple fallbacks and overlapping with #3202.
This was my first contribution attempt, and I really appreciate the feedback.
I’ll follow discussions in #3202 and #3217 to learn more about the project direction and possibly contribute in a more useful area next time.

PR #3219, opened 2025-08-26
Add language translations

bbannon:
Translating some things for another project.
Used google translate.
Then realized they should go here, too.


Synchro:
That's a lot of translations, thanks, but I don't think it's worth bulk adding them like this in the absence of any specific demand. The vast majority of these messages are only ever seen by developers, so unless someone actually wants to see errors in their own language while developing (historically, how all the others have been added), these are not likely to help anyone, while adding unnecessary size to millions of downloads.

Any app using PHPMailer is much more likely to provide its own translations.

bbannon:
The lead on the other project doesn't want to use your translations. So I
redid them.
I've never understood why he wants them all done again by.other
translators. Seems a waste of time to me.

But since I already had them, I figured to offer them to you.

Also, there has to be a way for people to only download the languages they
use.
Maybe you can figure it out. *I bet many other projects could benefit from
that.*
*He doesn't think it's possible. And I don't want to argue any more.*

SirLouen:
Anyway, I would not add automatic translations. 
They could be totally wrong, and generally this is poorly perceived. 
For example, WP has a ton of untranslated languages and strings and there are no problems with this. But it is very poorly perceived that any editor adds automatic translated strings. 
Anyone who can be interested on seeing their strings translated could: Submit a list with the translations for the language if they are not too tech-savvy to do the PR, or submit a PR directly. Any of the two will work and they will get translated.

bbannon:
if you don't want them, that's fine.

in the other project, I put comment at the top of file saying it's from
google translate to be used as a starting point and asked any native
speakers to please submit corrections.
So far, no corrections have come in.

PR #3226, opened 2025-09-11
Fix encoding header for SMIME signed messages with long lines

schorschii:
When creating SMIME signed messages, the `Content-Transfer-Encoding` header is currently broken if the message contains long lines.

This means `Content-Transfer-Encoding` is set to `8bit` while the the body is encoded `quoted-printable`.

This is because `$this->getMailMIME()` is called before the `hasLineLongerThanMax()` check. This change moves the `$this->getMailMIME()` call accordingly.


Synchro:
Good catch – Could you write a test for this please, and sign your commits?

Synchro:
Thank you!

PR #3228, opened 2025-09-20
GH Actions/scorecard: update permissions

jrfnl:
... to match the current recommendations.

I've removed the "read" permissions as those should only be needed for "private" repos.

Ref: https://github.com/ossf/scorecard-action#additional-permissions-for-private-repositories

### DIFF
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -7,8 +7,7 @@ on:
   push:
     branches: [ "master" ]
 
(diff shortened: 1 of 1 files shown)


Synchro:
Thank you Juliette!

jrfnl:
@Synchro You're welcome. I've got a few more of these type of things coming up - doing them in separate commits/PRs, so it is easier to monitor if something would break.

PR #3233, opened 2025-09-22
Fix unnecessary memory spike in PHPMailer::attachAll

K0bin:
`PHPMailer::attachAll` writes the contents of all attachments to an array first.
It only ever appends at the end of the array and never does anything else with the array besides turning it into a string at the end with `implode`.

This `implode` call creates a copy of all attachment contents.
The original array will only get freed after the function returns.
So for a very brief time, there are two copies of all attachments in memory causing a spike in memory usage.

As far as I know, the limit of typical mail servers for attachments is somewhere between 10 and 20MB.
Considering the default memory limit of 128MB this can be significant.

Thankfully it's easy to fix by just appending a string right from the start.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -3430,7 +3430,7 @@ class PHPMailer
     protected function attachAll($disposition_type, $boundary)
     {
         //Return text of body
-        $mime = [];
+        $mime = '';
         $cidUniq = [];
         $incl = [];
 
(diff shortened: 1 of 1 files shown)


K0bin:
Actually, there's a copy anyway when appending the string produced by `attachAll` to `$body`, so this PR unfortunately doesn't solve the problem.

That could be solved using a reference but that would change error handling behavior.
Right now, `attachAll` returns an empty string if encoding any of the attachments fails.
So either all attachments get attached or none at all.
Avoiding the copy would require doing the error checking up front.

Synchro:
Thanks for trying! I would love to be able to send attachments larger than memory, but doing streaming compression, base64/qp encoding, S/MIME, and DKIM signing all at once is really tricky.

Issue #3234, opened 2025-09-24
Why Reply-To is an associative array?


SirLouen:
I was looking in the past (only 16 years ago) and got to this
https://github.com/PHPMailer/PHPMailer/commit/8d4fc6d507a5b640059bac757dc3f6966e46e3ca

Historically ReplyTo was a regular indexed array
```php
public function AddReplyTo($address, $name = '') {
      $this->ReplyTo[] = array(trim($address), $name);
```

But then it becomes an associative array. From all 4 list address types, Reply-To is the associative array one. The rest are indexed arrays. It doesn't make much sense to me because I don't see that it's particularly use for anything

https://github.com/PHPMailer/PHPMailer/blob/6a6e66abf4106fe52db3bbd7a5315bc5d29df995/src/PHPMailer.php#L2795

But this generates a strange inconsistency among all 4.

Synchro:
This is true, however, it's also handled entirely separately from the `to`, `cc`, and `bcc` addressing properties, so it doesn't really have any impact. We could change it to use the same format as the others, but it would gain only consistency, no other benefits, and it's all hidden behind `addReplyTo` anyway so on balance 🤷‍♂️.

SirLouen:
The only benefit I found is that I've been doing some unit testing for this, passing same parameters for all, and I was a little bit confused because it was not passing for reply-to. After checking further, I discovered this pattern.

Basically I was looping through all 4 with the exact same parameters and expecting same results on the PHPMailer side. I don't think it wil have any advantage user-wise, apart from the consistency of being able to treat all them 4 like identical data structures

Synchro:
I had thought that `ReplyTo` was public, so changing it would be a BC break, but it's actually protected so we should be able to change this without a problem, though it would constitute a minor internal API change that might affect subclasses, so the next release would need to be a `x.n+1.0` not a `x.n.1`.

SirLouen:
Ok, so since the idea is moving into 6.11.0, it's a good opportunity to add a PR for this.

PR #3237, opened 2025-09-26
Reply-To to Indexed Array Conversion

SirLouen:
Following the conversation in #3234 this is the new version with Reply-To converted to  Indexed array like the rest. 
The idea of having an associative array, like the all recipients array was ok to be able to check with `array_key_exists` to avoid duplicates, but we can easily check if the address is already there with a simple algo

I've also sorted out all the tests to point to the first element of the indexed array.

With this patch now all 4 addresses list are identical (To, Cc, Bcc and Reply-To). And From is not because we have commented that for now we are not going to go nuts [with multiple Froms](https://github.com/PHPMailer/PHPMailer/discussions/3213)

With this and #3202, I think we are ready for 6.11.0?


Synchro:
That's tidy, thanks

PR #3240, opened 2025-09-30
Regressing deprecatedArg

SirLouen:
In continuation to #3202 
https://github.com/PHPMailer/PHPMailer/pull/3202#pullrequestreview-3282052464


SirLouen:
@Synchro I was thinking... can't we use parameters in the translated strings?

For example in this case:

`deprecated_argument' => 'Argument $useimap is deprecated',`
It would be better if instead of hardcoding `$useimap` we use have the parameter like `%s` and use `deprecated_argument` for this and any in the future deprecations with the corresponding argument.

Synchro:
One thing with language strings is that while they look like PHP, and are stored as .php files, they are actually handled as raw text. This is due to old vulnerabilities that exploited issues in dynamic language strings, and I’m keen to avoid that reoccurring!

SirLouen:
@Synchro then I was thinking that instead of using `deprecated_argument` text, we need to specify which argument is, like `deprecated_argument_useimap`. So it can be extensible to future arguments like `deprecated_argument_foo` and `deprecated_argument_bar`

Also, I wanted to add [the last suggestion](https://github.com/PHPMailer/PHPMailer/pull/3202#discussion_r2391144659) by @jrfnl. You merged too quickly. 😛

Let me add another PR with these changes.

PR #3241, opened 2025-09-30
Improving docs and translations for deprecated useimap

SirLouen:
I assume that you wanted to rush the 6.11.1 to avoid any troubles. 

I've added this docs changes that @jrfnl commented in #3202 

Also, as I commented in #3240 if we cannot pass dynamic values to the translation strings, then it's going to be better to name each array key for current and future `deprecated_argument` values.


Synchro:
Historically we have avoided having embedded values in translations by simply appending relevant values from the context in which they are called, so for example instead of phrasing it as `"missing imap extension"`, say `"missing extension: "`, and then append the name of the missing ext, which does not need translating. That way we get to reuse the translation elsewhere without making a proliferation problem like this.

SirLouen:
Ok, added this idea. Yes, I can confirm it looks better.

Issue #3260, opened 2025-10-29
SMTP + XOAUTH2 authentication error handling


greew:
I have helped a few people searching for errors when using XOAUTH2 auth and Azure. 

One of the common problems here are that the call to [`$OAuth->getOauth64()`](https://github.com/PHPMailer/PHPMailer/blob/master/src/SMTP.php#L636) throws an exception within this code, like
- `\League\OAuth2\Client\Provider\Exception\IdentityProviderException`
- `\UnexpectedValueException`
- `\GuzzleHttp\Exception\GuzzleException`

Since neither of these are a `\PHPMailer\PHPMailer\Exception` they are not caught in the normal catch handlers and are very hard to debug.

I would like to suggest, that we use something like 

```php
try {
    $oauth = $OAuth->getOauth64();
} catch (\Exception $e) {
    throw new \PHPMailer\PHPMailer\Exception("SMTP authentication error", 0, $e);
}
```

That way, we use PHPMailer exceptions and still have access to the previous exception which can tell us more about what actually went wrong, but we don't expose any sensitive information to the user.

Besides this, I'd recommend adding a guide of how to use this `$previous` exception and get data out of it.

I'll happily create a PR for this, if wanted.

Synchro:
That sounds reasonable, so please go ahead with a PR

greew:
And already done! 😃

PR #3261, opened 2025-10-29
Add XOAUTH2 token exception handling

greew:
If generating an OAuth2 token fails in the provider and an exception is thrown, we never catch that exception, but just let exception propagate up to whatever.

I suggest this change, such that we can handle errors when generating oauth tokens. It is based on the fact that `League\OAuth2\Client\Provider\AbstractProvider` can throw exceptions during getting the oauth token, but we are nowhere near to knowing why.

I have decided on catching all exception (`\Exception`) because if someone uses another `OAuthTokenProvider` (it's just an interface, so it's possible), we won't know specifically which exceptions can be thrown here, so we need to catch them all.

By throwing a PHPMailer exception and adding the existing exception as the `$previous` variable, we can show a generic error message to regular users and advice how developers can get actually meaningful error messages from the previous exception.

As discussed in #3260

### DIFF
--- a/src/SMTP.php
+++ b/src/SMTP.php
@@ -633,7 +633,13 @@ class SMTP
                 if (null === $OAuth) {
                     return false;
                 }
-                $oauth = $OAuth->getOauth64();
+                try {
+                    $oauth = $OAuth->getOauth64();
+                } catch (\Exception $e) {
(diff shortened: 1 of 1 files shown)


greew:
I have already added https://github.com/PHPMailer/PHPMailer/wiki/Troubleshooting#xoauth2-specific-errors to the wiki.

PR #3272, opened 2025-11-18
GH Actions: update for the release of PHP 8.5

jrfnl:
... which is expected to be released this Thursday.

* Builds against PHP 8.5 are no longer allowed to fail.
* Update PHP version on which code coverage is run (high should now be 8.5).
* Add _allowed to fail_ build against PHP 8.6.
* Update the README.

Note: for some jobs I use "nightly" for the "next" PHP version, for some `8.6`. While it may appear there is no difference and this is true for the better part of the year, there is a difference for about two months.

To illustrate, consider PHP 8.5:
* PHP "nightly" refers to the PHP `master` branch, so was PHP 8.5 until the PHP 8.5 was branched off when the first RC was cut in September.
* As of that moment, "nightly" basically became PHP 8.6, so to test against PHP 8.5, one would need to explicitly request `8.5`.
* As of the release of PHP 8.5, it is expected for "nightly" to be PHP 8.6, so the difference is moot again.

For that reason, the unit test workflow uses the explicit `8.6` version for PHP "next".

P.S.: this time I checked and didn't see a prep branch from @Synchro ;-)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -50,10 +50,10 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        php: ['5.5', '7.2', '8.0', '8.4']
+        php: ['5.5', '7.2', '8.0', '8.5']
         experimental: [false]
         include:
-          - php: '8.5'
+          - php: 'nightly'
             experimental: true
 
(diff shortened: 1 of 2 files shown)


Synchro:
Thorough as always! I'll merge this after the 8.5 release.

Issue #3242, opened 2025-09-30
PHPMailer 6.11.1 and PHP imap


campbell-m:
I must be missing something!  PHPMailer now complains "Using simplified address parser is not recommended. Install the PHP IMAP extension for full RFC822 parsing.", but [the imap extension was deprecated as of PHP 8.4.0](https://www.php.net/manual/en/intro.imap.php).

duckydan:
I see the same thing.

Synchro:
Annoying but true – but it has always been the case that the parser doesn't work properly without the imap extension, it's just that now we are saying so.
Given that message and address parsing is not a core feature of PHPMailer, I'd recommend looking at more complex parsers that do not rely on the imap extension.

/cc @SirLouen

SirLouen:
@campbell-m @duckydan I assume that you are explicitly calling `parseAddresses` in your setup.

For very basic parsing, `imap` is not required and everything will still be working. Note that this is just a **User Notice**, not a warning, not an error. You can ignore this if your protocols are only requiring elementary parsing. This error is triggered when you have the second argument of `parseAddresses` set to anything other than `null`: it doesnt mean that you have to use the `imap` extension. Once you set this to null, this message will disappear, and if you don't have `imap` extension installed, it will failover into a basic algorithm to parse addresses (mostly addresses like `Foo Bar <foo@exameple.org>`  or just addresses, nothing else). This will be exactly the same as before. 

But if you need some serious parsing, as many  people has been asking, the `imap` extension along with mbstring and some extra code is what is doing the trickery and what we agreed to maintain here. For now, we cannot build from scratch all the logic to maintain the RFC822, and there is no other simpler library to keep this up. 

The imap library has been [moved to PECL](https://github.com/php/pecl-mail-imap) and “unbundled” from PHP itself. 

Maybe if we can find a better substitute library in the following months, we can consider switching. Also note that the `imap` library was also being used before 6.11. With this user notice we are informing to those spare users of `parseAddresses` function that imap is better than nothing.

@Synchro maybe we should add a wiki page commenting about this?

campbell-m:
> This error is triggered when you have the second argument of parseAddresses set to anything other than `null`: 

[...]

PR #3276, opened 2025-11-23
Use PHPCompatibility 10.0.0(-alpha1)

jrfnl:
_Opening this PR in draft as there are a few issues for which I'd like the input of the maintainer to determine the solution direction (see comments)._

---

### CS: start using PHPCompatibility 10 

Long anticipated, finally here: PHPCompatibility 10.0.0-alpha1 🎉

PHPCompatibility 10.0.0 brings huge improvements in both what is being detected (> 50 new sniffs), as well as the detection accuracy for pre-existing sniffs.

Even though still "unstable", it is stable enough for our purposes and the advantages of using it outweigh the disadvantage of it being an unstable version. By setting the `minimum-stability` and `prefer-stable` settings in the `composer.json`, we can ensure that we don't get the `dev-develop` branch, but rather get a `10.0.0` tag, unstable or not.

Includes updating the PHPCS version constraints to match.
Includes updating the exclusions in the ruleset for changes in the error codes due to further changes in PHP having been made (and now being detected).

Ref:
* https://github.com/PHPCompatibility/PHPCompatibility/wiki/Upgrading-to-PHPCompatibility-10.0
* https://github.com/PHPCompatibility/PHPCompatibility/releases/tag/10.0.0-alpha1

### PHPCS/PHPCompatibility: favour inline ignores over blanket rule excludes 

Best practice tweak regarding the use of PHPCompatibility.

It is strongly recommended to use inline ignore annotations when a reported issue is not a problem (because it is accompanied by a `function_exists()`, `defined()` or other check), instead of excluding a rule completely via the ruleset.

Blanket code-base wide ignores mean that:
* ... if a PR introduces new code (or changes existing code) which uses a non-cross-version compatible PHP feature...
* ... and the code doesn't have the right safeguards in place for cross-version compatibility...
* ... PHPCompatibility would not flag it because of the codebase wide ignore...
* ... which could cause problems for the end-users.

Selective, inline ignores ensure that only the annotated error is ignored and only for that specific bit of code, preventing the above described problem.

It also means that when something changes in PHP - like a deprecated method being removed -, you will be notified about the issue again so you can review if the current cross-version compatibility tweak is still the most optimal one.

### PHPCS: add a few more selective ignores 

... for things already handled correctly in the code.

### Examples/mailing_list: remove use of E_STRICT 

While this is only example code, it is still code which should be functional for all supported PHP versions.

As `E_STRICT` has been deprecated since PHP 8.4 and wasn't really all that useful since PHP 7.0 anyway, it's probably better not to mention it in the example code.

Ref: https://wiki.php.net/rfc/deprecations_php_8_4#remove_e_strict_error_level_and_deprecate_e_strict_constant

### Escape special characters the same way in all PHP versions

The `htmlspecialchars()` function is used to escape arbitrary text strings for display.

Original the default for the `$flags` parameter of that function in PHP was `ENT_COMPAT`, which translates to "convert double quotes to `&quot;` and leave single quotes alone".

As of PHP 8.1, the default value for the `$flags` parameter has been made more robust and was changed to `ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401`, which translates to "convert both double and single quotes, replace invalid code unit sequences with a Unicode Replacement Character and treat code as HTML 4.01".

For code to provide the same/predictable output cross-version PHP, the `$flags` parameter should be explicitly set and what with the new default value being the more robust one, this commit adds that value for `$flags` in all instances of function calls to `htmlspecialchars()`.

Once the application minimum PHP version is PHP 8.1 or higher, the parameter can be removed again (as the value will then be the same as the default parameter value).

Ref: https://www.php.net/manual/en/function.htmlspecialchars.php

### DIFF
--- a/examples/gmail.phps
+++ b/examples/gmail.phps
@@ -83,15 +83,14 @@ if (!$mail->send()) {
     echo 'Message sent!';
     //Section 2: IMAP
     //Uncomment these to save your message in the 'Sent Mail' folder.
-    #if (save_mail($mail->getSentMIMEMessage())) {
-    #    echo "Message saved!";
-    #}
+    #save_mail($mail->getSentMIMEMessage());
 }
 
 //Section 2: IMAP
 //This example uses the directorytree/imapengine IMAP library: https://imapengine.com
 //Earlier versions of this code used the deprecated PHP imap_* functions.
-function save_mail($message) {
+function save_mail($message)
+{
     $mailbox = new \DirectoryTree\ImapEngine\Mailbox([
         'host' => 'imap.gmail.com',
         'port' => 993,
@@ -104,5 +103,4 @@ function save_mail($message) {
     $folder = $mailbox->folders()->find('Sent Mail');
 
     $folder->messages()->append($message);
-    return true;
 }


jrfnl:
Issue 1: the example code in the `gmail.phps` example uses the IMAP extension, while that extension has been removed from PHP in PHP 8.4.

Possible solutions:
* Add stern warning about the example code in the file.
* Update the example code with conditions to only use the imap functions if available and throw an exception if they are not available.
* Use an alternative extension ?

Opinions ?

```
FILE: examples/gmail.phps
------------------------------------------------------------------------------------------------------------------------
FOUND 3 ERRORS AFFECTING 3 LINES
------------------------------------------------------------------------------------------------------------------------
 102 | ERROR | Function imap_open() is removed since PHP 8.4
     |       | (PHPCompatibility.FunctionUse.RemovedFunctions.imap_openRemoved)
 104 | ERROR | Function imap_append() is removed since PHP 8.4
     |       | (PHPCompatibility.FunctionUse.RemovedFunctions.imap_appendRemoved)
 105 | ERROR | Function imap_close() is removed since PHP 8.4
     |       | (PHPCompatibility.FunctionUse.RemovedFunctions.imap_closeRemoved)
------------------------------------------------------------------------------------------------------------------------
```

jrfnl:
Issue 2: in multiple places in the codebase, though mostly in example code, the `htmlspecialchars()` function is used in a way which isn't necessarily cross-version compatible, but for the most part, this won't be problematic.

Possible solutions:
1. Add the `$flags` parameter and lead by setting a good example 😉 
    The question then becomes _set it to which value ?_
2. Ignore the issue completely (via the ruleset).
3. Selectively ignore it in the files in which it is currently being flagged (via the ruleset).
4. Selectively ignore it for the lines currently flagged (via inline ignore annotations).

Opinions ? 

```
FILE: get_oauth_token.php
------------------------------------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
------------------------------------------------------------------------------------------------------------------------
 181 | ERROR | The default value of the $flags parameter for htmlspecialchars() was changed from ENT_COMPAT to
     |       | ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401 in PHP 8.1. For cross-version compatibility, the $flags
     |       | parameter should be explicitly set.
     |       | (PHPCompatibility.ParameterValues.NewHTMLEntitiesFlagsDefault.NotSet)
------------------------------------------------------------------------------------------------------------------------


FILE: examples/mailing_list.phps
------------------------------------------------------------------------------------------------------------------------
FOUND 4 ERRORS AFFECTING 4 LINES
------------------------------------------------------------------------------------------------------------------------
 54 | ERROR | The default value of the $flags parameter for htmlspecialchars() was changed from ENT_COMPAT to
    |       | ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401 in PHP 8.1. For cross-version compatibility, the $flags
    |       | parameter should be explicitly set. (PHPCompatibility.ParameterValues.NewHTMLEntitiesFlagsDefault.NotSet)
 69 | ERROR | The default value of the $flags parameter for htmlspecialchars() was changed from ENT_COMPAT to
    |       | ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401 in PHP 8.1. For cross-version compatibility, the $flags
    |       | parameter should be explicitly set. (PHPCompatibility.ParameterValues.NewHTMLEntitiesFlagsDefault.NotSet)
 70 | ERROR | The default value of the $flags parameter for htmlspecialchars() was changed from ENT_COMPAT to
    |       | ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401 in PHP 8.1. For cross-version compatibility, the $flags
    |       | parameter should be explicitly set. (PHPCompatibility.ParameterValues.NewHTMLEntitiesFlagsDefault.NotSet)
 78 | ERROR | The default value of the $flags parameter for htmlspecialchars() was changed from ENT_COMPAT to
    |       | ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401 in PHP 8.1. For cross-version compatibility, the $flags
    |       | parameter should be explicitly set. (PHPCompatibility.ParameterValues.NewHTMLEntitiesFlagsDefault.NotSet)
------------------------------------------------------------------------------------------------------------------------


FILE: examples/send_file_upload.phps
------------------------------------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
------------------------------------------------------------------------------------------------------------------------
 57 | ERROR | The default value of the $flags parameter for htmlspecialchars() was changed from ENT_COMPAT to
    |       | ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401 in PHP 8.1. For cross-version compatibility, the $flags
    |       | parameter should be explicitly set. (PHPCompatibility.ParameterValues.NewHTMLEntitiesFlagsDefault.NotSet)
------------------------------------------------------------------------------------------------------------------------


FILE: examples/send_multiple_file_upload.phps
------------------------------------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
------------------------------------------------------------------------------------------------------------------------
 57 | ERROR | The default value of the $flags parameter for htmlspecialchars() was changed from ENT_COMPAT to
    |       | ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401 in PHP 8.1. For cross-version compatibility, the $flags
    |       | parameter should be explicitly set. (PHPCompatibility.ParameterValues.NewHTMLEntitiesFlagsDefault.NotSet)
------------------------------------------------------------------------------------------------------------------------
```

Synchro:
Regarding IMAP, this is indeed an annoyingly tricky issue, see discussion in #3242. The choices are:

1. Restore the broken code, ignore the problem; after all, it's been that way for years
2. Leave the IMAP dependency in place (making silently broken code noisy)
3. Drop the functionality that requires IMAP altogether
4. Introduce a dependency (which would be the only one) to provide the missing functionality currently provided by the IMAP ext.

jrfnl:
> Regarding IMAP, this is indeed an annoyingly tricky issue, see discussion in #3242. The choices are:
> 
> 1. Restore the broken code, ignore the problem; after all, it's been that way for years
> 2. Leave the IMAP dependency in place (making silently broken code noisy)
> 3. Drop the functionality that requires IMAP altogether
> 4. Introduce a dependency (which would be the only one) to provide the missing functionality currently provided by the IMAP ext.

Yeah, I saw & read that discussion before I opened this PR.

The IMAP usages in the actual package are all safeguarded sufficiently though, so those are not my concern for this PR.

The only ones left which aren't safeguarded are those in that example file, and as it is an example file, it doesn't justify introducing a dependency or even investing a lot of time.

So, I think that brings us back to my original suggestions:

> * Add stern warning about the example code in the file.
> * Update the example code with conditions to only use the imap functions if available and throw an exception if they are not available.

Synchro:
That's fair; the examples are often forgotten about as they are not core code.

In the gmail example, it might be better to use the new IMAP library mentioned in the IMAP thread, even though it's an external dep.

On `htmlspecialchars`, I think going with the 8.1 default value explicitly is good. It's both compatible and safe (AFAIK?).

jrfnl:
> That's fair; the examples are often forgotten about as they are not core code.
> 
> In the gmail example, it might be better to use the new IMAP library mentioned in the IMAP thread, even though it's an external dep.
> 
> On `htmlspecialchars`, I think going with the 8.1 default value explicitly is good. It's both compatible and safe (AFAIK?).

Sounds like a plan. Want to do the honours for the gmail example ? I'll add a commit for the htmlspecialchars instances.

Synchro:
I've dealt with the gmail IMAP thing; I changed the example to use imapengine.

jrfnl:
Thanks @Synchro ! In that case, I believe this PR is ready for review/merge.

Issue #3189, opened 2025-08-08
isQmail and no Sender blows


SirLouen:
In the scenario where we are using a qmail instance, with `isQmail` but we have not explicitely specified a `Sender` address, by the time it gets to [this line](https://github.com/SirLouen/PHPMailer/blob/4644d9958719f7a2eda827fbcd93a10e3aa26480/src/PHPMailer.php#L1822), system will blow up, because `qmail-inject` does not recognize the `-t` parameter.

Synchro:
I see that and can fix it, however, the workaround is to set a `Sender`, or to use SMTP to localhost, which (as I've said to you before) is faster and safer than `sendmail` or its equivalents. I use a qmail-based server myself, and have no problem submitting over 600,000 messages per hour via SMTP.

SirLouen:
Don't worry about workarounds; I'm working in the Mail component in WordPress, and through my review I could be finding random issues in this library, so I'm reporting them.

Synchro:
Does this change look OK to you?

SirLouen:
> Does this change look OK to you?

I will be testing and ping you back.

SirLouen:
@Synchro 

I've been reviewing this weekend with both a Qmail server online and my local test suite. 
I'm using these images for the Qmail testing
https://github.com/SirLouen/phpmailer-dev/
https://github.com/SirLouen/phpmailer-dev/blob/main/Dockerfile.qmail

qmail by default always shipped a sendmail wrapper in `/var/qmail/bin/sendmail`. 
https://github.com/amery/qmail/blob/master/sendmail.c
Basically this wrapper flags the `-t` with `-H` and it has been shipped since Qmail's early days (earlier than PHPMailer first version at least).

What I cannot really understand [is this workaround](https://github.com/SirLouen/PHPMailer/blob/a018520fada5c4b8743cf7adf13d315fa518fef3/src/PHPMailer.php#L1013) forcing `qmail inject`.

By default, all Qmail since the first versions, as I said, shipped with a `sendmail` wrapper executable in `/var/qmail/bin/sendmail` for `qmail-inject`. I am not sure who would prefer to directly use the `qmail-inject` directly. It appears that [this PR that motivated this](https://github.com/PHPMailer/PHPMailer/pull/126). The problem is that there is no description here, and I don't see any context of why this person was willing to use directly the `php-inject` instead of the `sendmail` wrapper. 

[...]

PR #3283, opened 2025-12-06
Fixing troubles with mailSend and Sender enabled

### LINKED ISSUE #3282: Sender troubles in mailSend
Following this conversation https://github.com/PHPMailer/PHPMailer/discussions/3281
I've noted that the Sender logic here

https://github.com/PHPMailer/PHPMailer/blob/360ae911ce62e25e11249f6140fa58939f556ebe/src/PHPMailer.php#L2059-L2070

This is problematic, for two reasons:
The use of `sendmail_from` is not innocuous for non-Windows systems. Non-Windows systems cannot make use of `sendmail_from` but despite this, PHPMailer is bringing a functionality they don't have unnecessarily (enabling Sender).

I think we need to give this a second thought, but there is a second, bigger problem: enabling Sender has caused a major outage in several WordPress installations after 6.9. The idea of returning sender by default was mine because this is the best way to achieve an easier pass on DMARC for most setups (obviously some setups will struggle in the process, but this was expected). 

But there was an issue I described here https://github.com/PHPMailer/PHPMailer/discussions/3281, that for some reason, users without Sender could send email, and adding the `Sender` they could not and this made 0 sense to me.

I'm sending a PR to explain the solution.

SirLouen:
Fixes #3282

Here is the problem: 
If the users already had defined the `sendmail_path` we were adding again to params the `-f` when the Sender was specified. 

This provoked a double sender (one from params and one from the sender path)

This patch simply check if the `-f` has been already set to avoid setting it again. 

About the Testing, it has been a little more complex than I expected
The problem here is that `sendmail_path` is `PHP_INI_SYSTEM` so it's impossible to modifiy this in runtime, so the only way to modify it is through the `php.ini` or passing the parameter to the PHPUnit binary run as I explain below.

I was checking the previous tests like `testMailSend` assuming that they were right as they were setting 
```php
ini_set('sendmail_path', '/usr/sbin/sendmail -t -i ');
```

But this is basically not doing anything as this cannot be set. 

The only way to execute the test with a different `sendmail_path` is through running something like this:

```
php -d sendmail_path="/usr/sbin/sendmail -t -i -frpath@example.org" ./vendor/bin/phpunit --filter  testMailSendWithSendmailParams
```

On the other side `sendmail_from` can be set as its not a `PHP_INI_SYSTEM` parameter. 

Note that the logic is very interesting in:

https://github.com/PHPMailer/PHPMailer/blob/360ae911ce62e25e11249f6140fa58939f556ebe/src/PHPMailer.php#L2059-L2070

Because adding a `sendmail_from` , enables the Sender in L2062 (regardless if the `setFrom` is `true` or not and then the L2064 validates the condition. This was the reason here https://github.com/PHPMailer/PHPMailer/discussions/3281 enabling it, despite being a non-Windows parameter, was changing the course of action (and this is why I thought it was a Poltergeist 🤣 because, how was it possible, that `sendmail_from` was doing something if the PHP docs explicitely said that this was a Windows-only parameter?). I reviewed the whole mail code in PHP and Sendmail server to only find that the problem was here...

This fix is somewhat time-sensitive to be rolled out in a new PHPMailer version because I need to send the patch downstream for next WP 6.9.1 release with the next version of PHPMailer to fix everyone's problems.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -221,6 +221,13 @@ class PHPMailer
     public $Sendmail = '/usr/sbin/sendmail';
 
     /**
+     * Parsed sendmail parameters extracted from sendmail_path.
+     *
+     * @var array<string, string>
+     */
+    private $SendmailParams = [];
+
+    /**
      * Whether mail() uses a fully sendmail-compatible MTA.
      * One which supports sendmail's "-oi -f" options.
      *
@@ -989,6 +996,57 @@ class PHPMailer
     }
 
     /**
+     * Extract supported sendmail parameters from a path string.
+     *
+     * @param string $sendmailPath
+     *
+     * @return string The sendmail path without the parsed parameters
+     */
+    private function parseSendmailPath($sendmailPath)
+    {
+        $sendmailPath = trim((string)$sendmailPath);
+        if ($sendmailPath === '') {
+            return $sendmailPath;
+        }
+
+        $parts = preg_split('/\s+/', $sendmailPath);
+        if (empty($parts)) {
+            return $sendmailPath;
+        }
+
+        $command = array_shift($parts);
+        $remainder = [];
+
+        // Parse only -t, -i, and -f parameters.
+        for ($i = 0; $i < count($parts); ++$i) {
+            $part = $parts[$i];
+            if (preg_match('/^-(i|t)$/', $part, $matches)) {
+                $this->SendmailParams[$matches[0]] = '';
+                continue;
+            }
+            if (preg_match('/^-f(.*)$/', $part, $matches)) {
+                $address = $matches[1];
+                if ($address === '' && isset($parts[$i + 1]) && strpos($parts[$i + 1], '-') !== 0) {
+                    $address = $parts[++$i];
+                }
+                if (static::validateAddress($address)) {
+                    $this->SendmailParams['-f'] = $address;
+                    continue;
+                }
+            }
+
+            $remainder[] = $part;
+        }
+
+        // The params that are not parsed are added back to the command.
+        if (!empty($remainder)) {
+            $command .= ' ' . implode(' ', $remainder);
+        }
+
+        return $command;
+    }
+
+    /**
      * Send messages using $Sendmail.
      */
     public function isSendmail()
@@ -996,10 +1054,9 @@ class PHPMailer
         $ini_sendmail_path = ini_get('sendmail_path');
 
         if (false === stripos($ini_sendmail_path, 'sendmail')) {
-            $this->Sendmail = '/usr/sbin/sendmail';
-        } else {
-            $this->Sendmail = $ini_sendmail_path;
+            $ini_sendmail_path = '/usr/sbin/sendmail';
         }
+        $this->Sendmail = $this->parseSendmailPath($ini_sendmail_path);
         $this->Mailer = 'sendmail';
     }
 
@@ -1011,10 +1068,9 @@ class PHPMailer
         $ini_sendmail_path = ini_get('sendmail_path');
 
         if (false === stripos($ini_sendmail_path, 'qmail')) {
-            $this->Sendmail = '/var/qmail/bin/qmail-inject';
-        } else {
-            $this->Sendmail = $ini_sendmail_path;
+            $ini_sendmail_path = '/var/qmail/bin/qmail-inject';
         }
+        $this->Sendmail = $this->parseSendmailPath($ini_sendmail_path);
         $this->Mailer = 'qmail';
     }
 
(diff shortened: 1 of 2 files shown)


SirLouen:
A WordPress member has also noticed that this is happening in the `sendmailSend` with `$this->Sendmail`, and I completely overlooked it. 

So I've also added a fix for this. I'm not super-happy with the fact that many other params could be potentially replicated.

Another proposal was to avoid users using the `sendmail_path` for the `isSendmail` method but it misses the point if there is actually another real path like `/opt/bin/sendmail` or whatever. 

The only alternative that I can think is to strip the params in `isSendmail` before setting the `$this->Sendmail` property.

I would like to hear your opinion @Synchro. I would not like to delay this to much because I would like to have it ready for 6.9.1 as it has been a massive breaking change in WP.

SirLouen:
I've been rethinking this, and I feel that for `isSendMail` there are two possible options
1. I'm thinking in the head of someone living in a webhost with no agency over `php.ini`. In this scenario for someone willing to use `isSendmail` maybe stripping the parameters and setting them back accordingly is the best option
2. The alternative is to just ignore and ask them to fix the `sendmail_path`. Although this is a little unfair, because many shared hosts just share the same type of config across their servers and they are not willing to do this for their clients. So I'm willing to look further into the first option. I'm going to send a patch with this idea.

I've rolled back the idea because I've noticed that we don't have much agency about which parameters can the MTA introduce, and the casuistry is too extense for us to control. For example:

```
/usr/sbin/sendmail -frpath@example.com
/usr/sbin/sendmail -oi
```

How many characters shall we take from there parameters? 
Because if we take only 1, in the second option, we are missing the `i` and taking it as a value, but if we take 2, we are taking the `r` from the `-f` and the r doesnt mean anything. So we cannot simply just parse any parameter, as some of them are beyond of our control and could be of use to the given use case.

The only option that comes to my mind, is that given we already add a controlled set of parameters:

1. `-f`
2. `-oi` (that is the same as `-i` and `-i` is what [is standard nowadays in the PHP docs](https://www.php.net/manual/en/mail.configuration.php)) 
3. `-t`

We parse the `sendmail_path` for any of these 3, and then set them back accordingly.

SirLouen:
Here is the final implementation

I've also taken advantage of this little revamp and fix what I was talking about in #3189, to improve a bit the logic regarding `qmail`. Now `sendmailSend` and `mailSend` are more robust and it will not be double adding parameters to the final command.

I've also added the unit tests with some examples. Don't hesitate to add more examples.

SirLouen:
@Synchro 

I'm going to do a briefing for the final code conclusions, so it can be easier to review:

1. `parseSendmailPath` new helper function: The idea is that we parse the `sendmail_path` so we leave it cleaner (what goes into `$this->Sendmail`). We should remove any potential duplications like `-t` and `-i` and the most important `-f` which has catastrophic results on duplication. I've added unit test for this function, with several use cases on how we receive the path, and how we may return the path. This function also sets the `Sender` address if the `sendmail_path` provides a `-f` as we are removing it here.

2. For both `isSendmail` and `isQmail` we take advantage of this helper function, to set the path correctly. 

3. Improved the logic of building the `sendmailSend` method `sendmail` command. With an unified `Sender` we don't have to do so many ifs and else that are a little confusing. Now its pretty simple: If we have a `Sender` from whatever method we take it (either from the `sendmail_from` or  the `sendmail_path` with the `-f` parameter, or the `fromMail` with the `true` value for the sender, since its unified, it will simply set the `-f` accordingly to the final `sendmail` command (for both qmail and any sendmail-based MTA)

4. Improved `MailTransportTest` unit tests. 
  - Set up and Tear down functions to manage the `sendmail_from` manipulation in this test class.
  - Two new tests, to check for the troubles explained in this PR, basically testing the clash of multiple sender addresses coming from different sources for both  the `sendmailSend` and `mailSend` methods.
  - And the last unit tests and commented in the first part, to test `parseSendmailPath` with a data provider, that can add more examples.

And that's all. Ping me up if you need any clarification.

SirLouen:
Hey @Synchro just passing to inform you that 6.9.1 for WP is slated for next year, no clear date yet, but there is plenty of time to land this, probably 3 or 4 weeks. Anyway, I did a fix on the WP side, so even if we can't land this on time, it won't affect any more the people.  I'm pretty much handling the essence of this ticket but before we set the `Sender`.

SirLouen:
hey @Synchro have you got any chance to give a check to this?

Synchro:
Hi, apologies for the delay, I've been busy with [something](https://badsmtp.com). I've been reading back through your changes and it all looks sensible, so I'll go ahead and merge this.

SirLouen:
@Synchro nice, now what will be good is to get a minor new release for this.

What is  that project about? Like a Mailpit?

Synchro:
Done!

It's a bit like smtp-sink, can do the same kind of thing as mailhog/mailpit, but the main purpose is to generate error responses on demand without having to configure anything on the server side.

SirLouen:
I will take a look. Sounds great

gmariani-cv:
There is an error with this patch. I have hundreds of sites hosted on Nexcess (now LiquidWeb) and Nexcess by defaults hosts them on unique random domains (i.e. [random].nxcli.io), but aliases the live domain to your site. Because of this, the server configure sendmail path looks like this:

`/sbin/sendmail -t -i -f postmaster@[random].nxcli.io`

In order to have mail sent from the website validate, we have a .user.ini file on every site with something like this:

`sendmail_from=wordpress@example.com`

This method worked up until WordPress 7.0 which includes PHPMailer 7.0.2 and this patch. The one line causing this issue is this:

`if (self::isShellSafe($this->Sender) && strpos($phpmailer_path, ' -f') === false) {`

Because of the way it creates the command it creates a command with two -f this this:

`/sbin/sendmail -t -i -f postmaster@[random].nxcli.io -oi -fwordpress@example.com -t`

Which is then ignored. But we need both in order to override the original address. Because of this change, none of our sites can send mail anymore and it's causing huge problems. My solution for now is to override the logic with this patch:

```
add_action('phpmailer_init', function ($phpmailer) {
    $phpmailer->isSendmail();
    $ini_path = ABSPATH . '.user.ini';
    if (!file_exists($ini_path)) return;
    $contents = parse_ini_file($ini_path);
    if (isset($contents['sendmail_from'])) $phpmailer->Sender = $contents['sendmail_from'];
}, 999);
```

isSendmail() reads in the .ini settings but crucially switches it from 'mail' to 'sendmail' which then uses sendmailSend() which doesn't have the guard in place. I am then able to re-set the From address from the .user.ini and send mail successfully again. Maybe a better patch would be to replace the existing -f flag instead of concatenating it? Something like this maybe?

```
if (self::isShellSafe($this->Sender)) {
    $phpmailer_path = preg_replace('/\s-f\S+/', '', $phpmailer_path);
    $params = sprintf('-f%s', $this->Sender);
}
```

That way it's the best of both worlds. For confirmation I did some testing as well, the _gf suffix is Gravity Forms, the _ini suffix is from .user.ini:

In v6.9.4
```
From: wordpress_gf@example.com
Received-SPF: Pass (mailfrom) identity=mailfrom; client-ip=###.###.###.###; helo=cloudhost-#######.us-midwest-1.nxcli.net; envelope-from=wordpress_ini@example.com; receiver=<UNKNOWN>
```

In v7.0
```
From: wordpress_gf@example.com
Received-SPF: None (mailfrom) identity=mailfrom; client-ip=###.###.###.###; helo=cloudhost--#######..us-midwest-1.nxcli.net; envelope-from=postmaster@[random].nxcli.io; receiver=<UNKNOWN>
```

In v7.0 when I revert the one line to what it was in v6.9.4
```
From: wordpress_gf@example.com
Received-SPF: Pass (mailfrom) identity=mailfrom; client-ip=###.###.###.###; helo=cloudhost--#######.us-midwest-1.nxcli.net; envelope-from=wordpress_ini@example.com; receiver=<UNKNOWN>
```

So I can say definitively that one line is causing the problem. Hopefully this can be patched and then integrated upstream into WordPress 7.0.1 or something. Thanks for your attention to this!

desrosj:
@gmariani-cv Could you please create a new issue here on this repository and an [issue upstream in WordPress Core's Trac instance](https://core.trac.wordpress.org/) for tracking purposes?

PR #3285, opened 2025-12-08
Dependabot/gh-actions: move to bi-weekly schedule

jrfnl:
:point_right: Important: this is for **version** updates only, not for security updates, which are handled separately and don't depend on this configuration.

---

PR #3229 updated the GitHub Actions workflows used in this repo to use "pinned" versions for external action runners to improve workflow security.

The current "frequency" is weekly. As these updates are rarely time-sensitive, it should be fine to receive them less frequently.

This commit tries to make it so by changing the Dependabot schedule for GitHub Actions to once every two weeks and late in the day when the queue should be mostly empty (as long as it's not a Monday), so the update PR will come in on a more predictable schedule.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,7 +8,8 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "cron"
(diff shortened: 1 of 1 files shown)


Synchro:
Thanks Juliette!

PR #3288, opened 2025-12-22
Improve memory usage when sending large attachments

RobinvanderVliet:
When I try to send an attachment that is 20 MB in size, I get the following error:

> **Fatal error:** Allowed memory size of 134217728 bytes exhausted (tried to allocate 4096 bytes) in **/var/www/vendor/phpmailer/phpmailer/src/SMTP.php** on line **80**2

But I do succeed in sending my email with the following patch. My suggestion makes it iterate over the file, without creating complete duplicate of it in memory.

### DIFF
--- a/src/SMTP.php
+++ b/src/SMTP.php
@@ -770,6 +770,25 @@ class SMTP
         }
     }
 
+    private function iterateLines($s)
+    {
+        $start = 0;
+        $length = strlen($s);
+
+        for ($i = 0; $i < $length; $i++) {
+            $c = $s[$i];
+            if ($c === "\n" || $c === "\r") {
+                yield substr($s, $start, $i - $start);
+                if ($c === "\r" && $i + 1 < $length && $s[$i + 1] === "\n") {
+                    $i++;
+                }
+                $start = $i + 1;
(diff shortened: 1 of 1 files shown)


Synchro:
Thanks, I've known that memory consumption was quite bad. I've been avoiding it as I thought it would be likely to cause problems with DKIM, S/MIME etc, but this approach seems to work well.

RobinvanderVliet:
@Synchro Thank you for the quick reply and merge!

I've noticed that I encounter other memory exhaustion errors when I try to send even bigger files. I'll try to submit some more patches in the future to make PHPMailer more memory efficient.

Synchro:
It doesn't surprise me – I've always wanted to be able to do things like send files that are larger than available memory, however, the need to have all the content available to do things like set content length, calculate hashes, DKIM or S/MIME sign, encode bodies, etc, make that quite tricky as they often stack up on each other. It would take a lot of juggling to keep them all straight. I don't see any way that it could be done without making multiple passes through all body content and attachments, so it may be that we need to sacrifice a little speed to reduce memory usage.

PR #3323, opened 2026-04-15
Fix fallback address parsing for display names containing commas

erhan1209:
The fallback RFC822 address parser in `PHPMailer::parseSimplerAddresses()` was splitting recipient lists with a plain `explode(',')`. That caused valid address lists to break in environments without the IMAP extension when display names or comments contained commas, for example `"Doe, Jane" <jane@example.com>`.

This change replaces the naive split with a parser-aware address list splitter that preserves commas inside quoted names, comments, and angle-bracketed address forms. It also adds regression tests covering quoted display names with commas and comments containing commas.

Testing:
Unable to run PHPUnit in this environment because `php`/`phpunit` are not installed locally.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -1354,7 +1354,7 @@ class PHPMailer
         trigger_error(self::lang('imap_recommended'), E_USER_NOTICE);
 
         $addresses = [];
-        $list = explode(',', $addrstr);
+        $list = static::splitAddressList($addrstr);
         foreach ($list as $address) {
             $address = trim($address);
             //Is there a separate name part?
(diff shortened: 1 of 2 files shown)


erhan1209:
Sorry, just figured how verifying works.

Synchro:
Note the name of the function. if you look back at the commit history for this code, you'll see this was a deliberate choice, and it was created essentially as a placeholder due to the deprecation of the IMAP extension. PHPMailer doesn't want to get into the monstrously complicated realm of parsing messages and addresses, as even the approach you have taken is not sufficient. Put another way, if the provided simple splitter is not good enough, use an external library that implements proper parsing and override the method yourself; it's outside PHPMailer's goals. Sorry to disappoint.

PR #3328, opened 2026-05-10
Restore deprecated useimap boolean compatibility

masskrdjn:
## Summary

This restores backward compatibility for the deprecated `$useimap` argument in `PHPMailer::parseAddresses()`.

Previously, passing `true` or `false` was valid. The argument is now deprecated, but recent behavior triggered an `E_USER_DEPRECATED` warning for boolean values and ignored `false`, which could break applications that support multiple PHPMailer versions or convert PHP warnings into exceptions.

## Changes

- Keep accepting historical boolean values for `$useimap` without emitting a runtime deprecation warning.
- Restore the previous behavior where `$useimap === false` forces the simpler non-IMAP parser.
- Add a regression test covering both `true` and `false`.

## Tests

```powershell
.\vendor\bin\phpunit --no-coverage test\PHPMailer\ParseAddressesTest.php

Result : OK (26 tests, 40 assertions)
```

---

Note: this is my first contribution here. I noticed the branch protection requires verified signatures; I can re-sign the commit if needed.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -1284,26 +1284,27 @@ class PHPMailer
     /**
      * Parse and validate a string containing one or more RFC822-style comma-separated email addresses
      * of the form "display name <address>" into an array of name/address pairs.
-     * Uses the imap_rfc822_parse_adrlist function if the IMAP extension is available.
+     * Uses the imap_rfc822_parse_adrlist function if the IMAP extension is available and
+     * the deprecated $useimap argument is truthy.
      * Note that quotes in the name part are removed.
      *
      * @see https://www.andrew.cmu.edu/user/agreen1/testing/mrbs/web/Mail/RFC822.php A more careful implementation
      *
      * @param string $addrstr The address list string
      * @param bool|null $useimap Deprecated in PHPMailer 6.11.0.
-     *                           Previously this argument determined whether to use
-     *                           the IMAP extension to parse the list and accepted a boolean value.
+     *                           Truthy values request the deprecated IMAP parser
+     *                           and trigger a deprecation warning.
      * @param string $charset The charset to use when decoding the address list string.
      *
      * @return array
      */
     public static function parseAddresses($addrstr, $useimap = null, $charset = self::CHARSET_ISO88591)
     {
-        if ($useimap !== null && !is_bool($useimap)) {
+        if ($useimap == true) {
             trigger_error(self::lang('deprecated_argument') . '$useimap', E_USER_DEPRECATED);
         }
         $addresses = [];
-        if ($useimap !== false && function_exists('imap_rfc822_parse_adrlist')) {
+        if ($useimap == true && function_exists('imap_rfc822_parse_adrlist')) {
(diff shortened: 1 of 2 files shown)


Synchro:
Yes, please sign your commits.

I see what you're aiming for here. We don't need to worry about using deprecated `imap_*` functions if the caller has explicitly passed `false`, but I think you've gone too far the other way, in that the deprecation warning *should* be triggered if the user passes `true`, like this:

```php
        if ($useimap === true) {
            trigger_error(self::lang('deprecated_argument') . '$useimap', E_USER_DEPRECATED);
        }
        $addresses = [];
        if ($useimap === true && function_exists('imap_rfc822_parse_adrlist')) {
```

This alters the behaviour slightly so that `null` and `false` work the same way, but `true` will trigger the warning. Previously `null` did nothing, but both `false` and `true` triggered the warning, even though `false` meant that it would not try to use the underlying deprecated functions that were the reason for the deprecation in PHPMailer.

Thoughts?

Synchro:
It might be more effective if the two comparisons used `==` instead of `===` because then it would trap non-boolean truthy values too, much as your `!is_bool()` check did.

jrfnl:
> It might be more effective if the two comparisons used `==` instead of `===` because then it would trap non-boolean truthy values too, much as your `!is_bool()` check did.

With ☝🏻 in mind, it might also be worth it to make the test one with a data provider which passes a range of values to run the test with ?

masskrdjn:
Thanks for your feedbacks, that makes sense. 

@Synchro I agree that passing `false` should not trigger the deprecation warning, as it explicitly avoids the deprecated IMAP path.

Using `== true` sounds reasonable if the intended behavior is to treat truthy values as an explicit request for the IMAP path. I can update the patch accordingly and add a data provider covering `null`, `false`, `true`, and a few truthy/falsy non-boolean values so the intended behavior is documented.

@jrfnl I also see the point about `true`: if the caller explicitly asks for IMAP parsing, triggering the deprecation warning is appropriate.

What do you think about the following?

```php
public static function parseAddresses($addrstr, $useimap = null, $charset = self::CHARSET_ISO88591)
{
    if ($useimap == true) {
        trigger_error(self::lang('deprecated_argument') . '$useimap', E_USER_DEPRECATED);
    }
    $addresses = [];
    if ($useimap == true && function_exists('imap_rfc822_parse_adrlist')) {
        //Use this built-in parser if it's available
        // phpcs:ignore PHPCompatibility.FunctionUse.RemovedFunctions.imap_rfc822_parse_adrlistRemoved -- wrapped in function_exists()
        $list = imap_rfc822_parse_adrlist($addrstr, '');
```

Synchro:
Yes, that's exactly what I was suggesting.

Annoyingly the CI test action is failing at the moment due to an upstream problem with fetching packages inside the test env. It does this occasionally.

Synchro:
Thanks, I will tag this for release soon, along with some other changes, after CI has started working again.

masskrdjn:
Nice, thanks

PR #3330, opened 2026-05-14
README patch.

Maikuolan:
Adds mention of 7.x to the legacy paragraph.


Synchro:
Thanks for the PR. This change is too specific, so I have made references to later versions more generic in the readme, and also updated the UPGRADING file, which contained similar references.

Maikuolan:
Sounds good. :+1:

PR #3340, opened 2026-06-03
Clarify Gmail app password usage in Gmail SMTP example

Safaa-Raboudi:
This PR updates the Gmail SMTP example to clarify that Gmail users should use an app password when 2-Step Verification is enabled.

It also points users to the XOAUTH2 example for OAuth-based authentication.

### DIFF
--- a/examples/gmail.phps
+++ b/examples/gmail.phps
@@ -48,6 +48,8 @@ $mail->SMTPAuth = true;
 $mail->Username = 'username@gmail.com';
 
 //Password to use for SMTP authentication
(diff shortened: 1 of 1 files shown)


Synchro:
This code example already contains a comment pointing at the xoauth2 example; I don't see any need to repeat that. A pointer to app passwords is reasonable, as it's not immediately obvious that regular passwords don 't work with gmail any more, even though they look identical in use, so I recommend adding a link to [Google's docs on that](https://support.google.com/mail/answer/185833?hl=en).

Also, please sign your commits.

Synchro:
Thanks

PR #3347, opened 2026-06-18
Fix mailSend() Sender override when sendmail_path already contains -f

### LINKED ISSUE #3342: .user.ini sendmail_from override broken in 7.0.2
There is an error with this patch. I have hundreds of sites hosted on Nexcess (now LiquidWeb) and Nexcess by defaults hosts them on unique random domains (i.e. [random].nxcli.io), but aliases the live domain to your site. Because of this, the server configure sendmail path looks like this:

`/sbin/sendmail -t -i -f postmaster@[random].nxcli.io`

In order to have mail sent from the website validate, we have a .user.ini file set sendmail_from on every site with something like this:

`sendmail_from=wordpress@example.com`

This method worked up until WordPress 7.0 which includes PHPMailer 7.0.2 and pull request #3283 . The one line causing this issue is this:

`if (self::isShellSafe($this->Sender) && strpos($phpmailer_path, ' -f') === false) {`

Because of the way it creates the command it creates a command with two -f like this:

`/sbin/sendmail -t -i -f postmaster@[random].nxcli.io -oi -fwordpress@example.com -t`

Which is then ignored by the new guard. But we need both in order to override the original address. Because of this change, none of our sites can send mail anymore and it's causing huge problems. My solution for now is to override the logic with this patch:

```
add_action('phpmailer_init', function ($phpmailer) {
    $phpmailer->isSendmail();
    $sendmail_from_value = ini_get('sendmail_from');
    if (empty($sendmail_from_value)) return;
    $phpmailer->Sender = $sendmail_from_value;
}, 999);
```

isSendmail() reads in the .ini settings but crucially switches it from 'mail' to 'sendmail' which then uses sendmailSend() which doesn't have the guard in place. I am then able to re-set the From address from the .user.ini and send mail successfully again. Maybe a better patch would be to replace the existing -f flag instead of concatenating it? Something like this maybe?

```
if (self::isShellSafe($this->Sender)) {
    $phpmailer_path = preg_replace('/\s-f\S+/', '', $phpmailer_path);
    $params = sprintf('-f%s', $this->Sender);
}
```

That way it's the best of both worlds. For confirmation I did some testing as well, the _gf suffix is Gravity Forms, the _ini suffix is from .user.ini:

In v6.9.4
```
From: wordpress_gf@example.com
Received-SPF: Pass (mailfrom) identity=mailfrom; client-ip=###.###.###.###; helo=cloudhost-#######.us-midwest-1.nxcli.net; envelope-from=wordpress_ini@example.com; receiver=<UNKNOWN>
```

In v7.0
```
From: wordpress_gf@example.com
Received-SPF: None (mailfrom) identity=mailfrom; client-ip=###.###.###.###; helo=cloudhost--#######..us-midwest-1.nxcli.net; envelope-from=postmaster@[random].nxcli.io; receiver=<UNKNOWN>
```

In v7.0 when I revert the one line to what it was in v6.9.4
```
From: wordpress_gf@example.com
Received-SPF: Pass (mailfrom) identity=mailfrom; client-ip=###.###.###.###; helo=cloudhost--#######.us-midwest-1.nxcli.net; envelope-from=wordpress_ini@example.com; receiver=<UNKNOWN>
```

So I can say definitively that one line is causing the problem. Hopefully this can be patched and then integrated upstream into WordPress 7.0.1 or something. Thanks for your attention to this!

IamTIqbal:
## Problem

When `sendmail_path` is configured by a hosting provider with a `-f` flag (e.g. `/sbin/sendmail -t -i -f postmaster@random.nxcli.io`), the `mailSend()` method was skipping the user-specified `Sender` entirely due to this guard added in #3283:

```php
if (self::isShellSafe($this->Sender) && strpos($phpmailer_path, ' -f') === false) {
    $params = sprintf('-f%s', $this->Sender);
}
```

This meant the server-configured address appeared as the envelope sender instead of the intended `Sender`, breaking SPF validation and causing mail delivery issues.

Reported in #3342 — affects users on hosts like Nexcess/LiquidWeb where `sendmail_path` is pre-configured with a random subdomain `-f` address.

## Fix

Strip any pre-existing `-f` flag from `sendmail_path` before adding the user's `Sender` parameter, then restore the original `sendmail_path` after sending:

```php
if (self::isShellSafe($this->Sender)) {
    $clean_path = preg_replace('/ -f\s*\S+/', '', $phpmailer_path);
    if ($clean_path !== $phpmailer_path) {
        ini_set('sendmail_path', $clean_path);
    }
    $params = sprintf('-f%s', $this->Sender);
}
```

This ensures:
- The user-specified `Sender` always takes precedence over server defaults
- No duplicate `-f` flags are passed to sendmail
- The original `sendmail_path` is restored after each send
- Security guards (`isShellSafe`, `validateAddress`) remain in place

Fixes #3342


Synchro:
Overall, something as precarious as a method that sets command line params needs *very* thorough tests. Here you are allowing an unvalidated string be passed into the command line through `$phpmailer_path`. Also this PR contains random, unneeded, unrelated changes that look like AI slop. Do better.

PR #3358, opened 2026-09-10
Normalise line breaks in S/MIME signed messages

### LINKED ISSUE #3067: 550 5.6.11 SMTPSEND.BareLinefeedsAreIllegal; message contains bare linefeeds, which cannot be sent via DATA and receiving system does not support BDAT
PHPMailer 6.9.1

When signing a email with certificate and sending to some Mail servers whose are sensitive to new lines \r\n  (example email.seznam.cz) error message from server appears:

550 5.6.11 SMTPSEND.BareLinefeedsAreIllegal; message contains bare linefeeds, which cannot be sent via DATA and receiving system does not support BDAT  

The fix is add to file PHPMailer.php in function createBody():
`$body = preg_replace("/(?<!\r)\n/", "\r\n", $body);`

so new code looks like:
`                $file = tempnam(sys_get_temp_dir(), 'srcsign');
                $signed = tempnam(sys_get_temp_dir(), 'mailsign');
                $body = preg_replace("/(?<!\r)\n/", "\r\n", $body);
                file_put_contents($file, $body);

                //Workaround for PHP bug https://bugs.php.net/bug.php?id=69197
                if (empty($this->sign_extracerts_file)) {
                    $sign = @openssl_pkcs7_sign(
                        $file,
                        $signed,
                        'file://' . realpath($this->sign_cert_file),
                        ['file://' . realpath($this->sign_key_file), $this->sign_key_pass],
                        []
                    );
                } else {
                    $sign = @openssl_pkcs7_sign(
                        $file,
                        $signed,
                        'file://' . realpath($this->sign_cert_file),
                        ['file://' . realpath($this->sign_key_file), $this->sign_key_pass],
                        [],
                        PKCS7_DETACHED,
                        $this->sign_extracerts_file
                    );
                }

                @unlink($file);
                if ($sign) {
                    $body = file_get_contents($signed);
                    @unlink($signed);
                    //The message returned by openssl contains both headers and body, so need to split them up
                    $parts = explode("\n\n", $body, 2);
                    $this->MIMEHeader .= $parts[0] . static::$LE . static::$LE;
                    $body = $parts[1];
                    //$this->MIMEHeader = preg_replace("/(?<!\r)\n/", "\r\n", $this->MIMEHeader);
                    $body = preg_replace("/(?<!\r)\n/", "\r\n", $body);
                } else {
                    @unlink($signed);
                    throw new Exception($this->lang('signing') . openssl_error_string());
                }
            } catch (Exception $exc) {
                $body = '';
                if ($this->exceptions) {
                    throw $exc;
                }
            }
        }

        return $body;
`

Hope it helps someone.

dualfroz:
Fixes #3067.

`openssl_pkcs7_sign()` returns its headers and body with bare LF line breaks.
`createBody()` copied both into the message verbatim, so every S/MIME signed
message went out with LF-only line endings, and servers that enforce RFC 5321
line endings reject them with `550 5.6.11 SMTPSEND.BareLinefeedsAreIllegal`.

This runs openssl's output through the existing `normalizeBreaks()` helper,
matching what `DKIM_Sign()` already does with `$signHeader`.

Counted on a message built with `preSend()` and a self-signed certificate:

|                | bare LF | CRLF |
| -------------- | ------- | ---- |
| unsigned       | 0       | 11   |
| signed, before | 46      | 14   |
| signed, after  | 0       | 60   |

Unsigned messages are unaffected — every bare LF came from the signing step.

Normalising after signing does not invalidate the signature: S/MIME
canonicalises to CRLF before digesting, and `openssl smime -verify` reports
`Verification successful` both before and after the change.

`testSigning()` gained an assertion that the sent message contains no bare
linefeeds. It fails with 50 of them on `master` and passes with this change.

One note on the report: it also suggested normalising *before* signing, but the
body already has no bare LF at that point, so that half is not needed. The
header does need it — `MIME-Version:` lands on the fifth line.

### DIFF
--- a/src/PHPMailer.php
+++ b/src/PHPMailer.php
@@ -3320,8 +3320,9 @@ class PHPMailer
                     @unlink($signed);
                     //The message returned by openssl contains both headers and body, so need to split them up
                     $parts = explode("\n\n", $body, 2);
-                    $this->MIMEHeader .= $parts[0] . static::$LE . static::$LE;
-                    $body = $parts[1];
+                    //openssl emits bare LF breaks, which SMTP servers may reject
+                    $this->MIMEHeader .= static::normalizeBreaks($parts[0]) . static::$LE . static::$LE;
+                    $body = static::normalizeBreaks($parts[1]);
                 } else {
                     @unlink($signed);
                     throw new Exception(self::lang('signing') . openssl_error_string());
--- a/test/PHPMailer/PHPMailerTest.php
+++ b/test/PHPMailer/PHPMailerTest.php
@@ -971,6 +971,11 @@ EOT;
 
         $msg = $this->Mail->getSentMIMEMessage();
         self::assertStringNotContainsString("\r\n\r\nMIME-Version:", $msg, 'Incorrect MIME headers');
+        self::assertSame(
+            0,
+            preg_match_all("/(?<!\r)\n/", $msg),
+            'Signed message contains bare line feeds'
+        );
         unlink($certfile);
         unlink($keyfile);
     }


Synchro:
A nice clean fix, thanks. Please sign your commits though.

